PR #42, opened 2014-11-12
RNG Local Seeds

gammell:
This is mostly a synced version of [a pull request](https://bitbucket.org/ompl/ompl/pull-request/32/rng-local-seeds) from BitBucket. Discussion in said pull request had favoured creating a "quasi-random" constructor for creating samplers of known seeds; however, I am unclear if that is the correct answer, as it:

1. Provides no method to set deterministic behaviour for RNGs themselves, only samplers (i.e., RRT's goal sampling would be unaffected)
2. Doesn't scale well to different types of state samplers ([a future pull request](https://github.com/utiasASRL/batch-informed-trees/tree/informed_sampling))


gammell:
Hey guys,
So I take it this is still a "no"? I'm a little busy with some other things at the moment, but let me know and I can start working away on making the changes to get this merged.

mamoll:
Yes, still a "no". I should have been clearer about that. Ioan and I have discussed this extensively and have given it serious thought, but in the end it does not seem like the right solution for the reasons that Ioan has stated above. We are still eagerly anticipating your pull request for BIT* :-).

gammell:
No problem, these are you decisions to make after all. :)
I'll work at making a pull request with the changes Ioan approved then we can move on to the more exciting things (i.e., algorithms). It'll be a little slow though, I have some other things on the go this time of year.

PR #48, opened 2015-03-05
More miscellanea

gammell:
Hi Guys,
So I found a few more miscellaneous features when getting the Informed Sampling pull request ready. I figured it was better to do them separately first and keep things a little more neat and tidy.

So here you go!
Jon

### DIFF
--- a/src/ompl/base/OptimizationObjective.h
+++ b/src/ompl/base/OptimizationObjective.h
@@ -89,32 +89,17 @@ namespace ompl
             /** \brief Set the cost threshold for objective satisfaction. When a path is found with a cost better than the cost threshold, the objective is considered satisfied. */
             void setCostThreshold(Cost c);
 
-            /** \brief Get the cost that corresponds to an entire path. This implementation assumes \e Path is of type \e PathGeometric.*/
-            virtual Cost getCost(const Path &path) const;
-
             /** \brief Check whether the the cost \e c1 is considered better than the cost \e c2. By default, this returns true only if c1 is less by at least some threshold amount, for numerical robustness. */
             virtual bool isCostBetterThan(Cost c1, Cost c2) const;
 
-            /** \brief Compare whether cost \e c1 is worse than cost \e c2. By default defined as isCostBetterThan(c2, c1), as if c2 is better than c1, then c1 is worse than c2. */
-            virtual bool isCostWorseThan(Cost c1, Cost c2) const;
-
             /** \brief Compare whether cost \e c1 and cost \e c2 are equivalent. By default defined as !isCostBetterThan(c1, c2) && !isCostBetterThan(c2, c1), as if c1 is not better than c2, and c2 is not better than c1, then they are equal. */
             virtual bool isCostEquivalentTo(Cost c1, Cost c2) const;
-            
-            /** \brief Compare whether cost \e c1 and cost \e c2 are not equivalent. By default defined as isCostBetterThan(c1, c2) || isCostBetterThan(c2, c1), as if c1 is better than c2, or c2 is better than c1, then they are not equal. */
-            virtual bool isCostNotEquivalentTo(Cost c1, Cost c2) const;
-
-            /** \brief Compare whether cost \e c1 is better or equivalent to cost \e c2. By default defined by !isCostBetterThan(c2, c1), as if c2 is not better than c1, then c1 is better than, or equal to, c2. */
-            virtual bool isCostBetterThanOrEquivalentTo(Cost c1, Cost c2) const;
-
-            /** \brief Compare whether cost \e c1 is worse or equivalent to cost \e c2. By default defined by !this->isCostBetterThan(c1,c2) as if c1 is not better than c2, than c1 is worse than, or equal to, c2*/
-            virtual bool isCostWorseThanOrEquivalentTo(Cost a, Cost b) const;
 
             /** \brief Returns whether the cost is finite or not. By default calls std::isfinite on Cost::value(). */
             virtual bool isFinite(Cost cost) const;
 
             /** \brief Return the minimum cost given \e c1 and \e c2. Uses isCostBetterThan. */
-            virtual Cost minCost(Cost c1, Cost c2) const;
+            virtual Cost betterCost(Cost c1, Cost c2) const;
 
             /** \brief Evaluate a cost map defined on the state space at a state \e s. */
             virtual Cost stateCost(const State *s) const = 0;
@@ -124,12 +109,6 @@ namespace ompl
 
             /** \brief Get the cost that corresponds to combining the costs \e c1 and \e c2. Default implementation defines this combination as an addition. */
             virtual Cost combineCosts(Cost c1, Cost c2) const;
-
-            /** \brief Combine three costs as (c1 + c2) + c3. By default uses the two-argument version of combineCosts */
-            virtual Cost combineCosts(Cost c1, Cost c2, Cost c3) const;
-
-            /** \brief Combine four costs as ((c1 + c2) + c3) + c4. By default uses the two-argument version of combineCosts */
-            virtual Cost combineCosts(Cost c1, Cost c2, Cost c3, Cost c4) const;
 
             /** \brief Get the identity cost value. The identity cost value is the cost c_i such that, for all costs c, combineCosts(c, c_i) = combineCosts(c_i, c) = c. In other words, combining a cost with the identity cost does not change the original cost. By default, a cost with the value 0.0 is returned. It's very important to override this with the proper identity value for your optimization objectives, or else optimal planners may not work. */
             virtual Cost identityCost() const;
--- a/src/ompl/base/OptimizationObjective.h
+++ b/src/ompl/base/OptimizationObjective.h
@@ -95,6 +95,19 @@ namespace ompl
             /** \brief Check whether the the cost \e c1 is considered better than the cost \e c2. By default, this returns true only if c1 is less by at least some threshold amount, for numerical robustness. */
             virtual bool isCostBetterThan(Cost c1, Cost c2) const;
 
+            /** \brief Return the minimum cost given \e c1 and \e c2. Uses isCostBetterThan. */
+            virtual Cost minCost(Cost c1, Cost c2) const
+            {
+                if (isCostBetterThan(c1, c2))
+                {
+                    return c1;
+                }
+                else
+                {
+                    return c2;
+                }
+            }
+
             /** \brief Evaluate a cost map defined on the state space at a state \e s. */
             virtual Cost stateCost(const State *s) const = 0;
 
--- a/src/ompl/base/OptimizationObjective.h
+++ b/src/ompl/base/OptimizationObjective.h
@@ -126,10 +126,10 @@ namespace ompl
             virtual Cost combineCosts(Cost c1, Cost c2) const;
 
             /** \brief Combine three costs as (c1 + c2) + c3. By default uses the two-argument version of combineCosts */
-            virtual ompl::base::Cost combineCosts(Cost c1, Cost c2, Cost c3) const;
+            virtual Cost combineCosts(Cost c1, Cost c2, Cost c3) const;
 
             /** \brief Combine four costs as ((c1 + c2) + c3) + c4. By default uses the two-argument version of combineCosts */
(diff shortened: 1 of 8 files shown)


src/ompl/base/src/OptimizationObjective.cpp
```diff
@@ -65,16 +66,102 @@ void ompl::base::OptimizationObjective::setCostThreshold(Cost c)
...
+
+bool ompl::base::OptimizationObjective::isCostWorseThanOrEquivalentTo(Cost c1, Cost c2) const
+{
+    //If c1 is not better than c2, than c1 is worse than, or equal to, c2
+    return !isCostBetterThan(c1,c2);
+}
+
+bool ompl::base::OptimizationObjective::isFinite(Cost cost) const
+{
+    return std::isfinite(cost.value());
```
gammell: Alternatively, a better default may be:
    isCostBetterThan(cost, this->infiniteCost())

src/ompl/base/src/OptimizationObjective.cpp
```diff
@@ -65,16 +66,102 @@ void ompl::base::OptimizationObjective::setCostThreshold(Cost c)
...
     return c1.value() + magic::BETTER_PATH_COST_MARGIN < c2.value();
 }
 
+bool ompl::base::OptimizationObjective::isCostWorseThan(Cost c1, Cost c2) const
+{
+    //If c2 is better than c1, then c1 is worse than c2
+    return isCostBetterThan(c2, c1);
+}
+
+bool ompl::base::OptimizationObjective::areCostsEquivalent(Cost c1, Cost c2) const
```
gammell: I didn't need it, so I didn't define
```
    areCostsNotEquivalent
```
but I can add that if desired.

src/ompl/base/OptimizationObjective.h
```diff
@@ -85,9 +89,33 @@ namespace ompl
...
+            /** \brief Compare whether cost \e c1 is worse or equivalent to cost \e c2. By default defined by !this->isCostBetterThan(c1,c2) as if c1 is not better than c2, than c1 is worse than, or equal to, c2*/
+            virtual bool isCostWorseThanOrEquivalentTo(Cost a, Cost b) const;
+
+            /** \brief Returns whether the cost is finite or not. By default calls std::isfinite on Cost::value(). */
+            virtual bool isFinite(Cost cost) const;
+
+            /** \brief Return the minimum cost given \e c1 and \e c2. Uses isCostBetterThan. */
+            virtual Cost minCost(Cost c1, Cost c2) const;
+
             /** \brief Evaluate a cost map defined on the state space at a state \e s. */
```
mamoll: I am not a fan of adding functions that are just syntactic sugar like `isCostWorseThan`, `isCostNotEquivalentTo`, `isCostBetterThanOrEquivalent`, and `isCostWorseThanOrEquivalent`. Functions like `isCostEquivalent`, `isFinite`, and `minCost` are ok, since they actually save some space and clarify code if you use it a lot.

src/ompl/base/OptimizationObjective.h
```diff
@@ -97,6 +125,12 @@ namespace ompl
             /** \brief Get the cost that corresponds to combining the costs \e c1 and \e c2. Default implementation defines this combination as an addition. */
             virtual Cost combineCosts(Cost c1, Cost c2) const;
 
+            /** \brief Combine three costs as (c1 + c2) + c3. By default uses the two-argument version of combineCosts */
+            virtual Cost combineCosts(Cost c1, Cost c2, Cost c3) const;
+
+            /** \brief Combine four costs as ((c1 + c2) + c3) + c4. By default uses the two-argument version of combineCosts */
+            virtual Cost combineCosts(Cost c1, Cost c2, Cost c3, Cost c4) const;
+
```
mamoll: Are there use cases for non-default behavior for the three- and four-argument version of CombineCosts? If not, then I'd rather not add these.

src/ompl/base/PlannerTerminationCondition.h
```diff
@@ -120,6 +121,32 @@ namespace ompl
 
         /** \brief Return a termination condition that will become true \e duration seconds in the future (wall-time), but is checked in a separate thread, every \e interval seconds; \e interval must be less than \e duration */
         PlannerTerminationCondition timedPlannerTerminationCondition(double duration, double interval);
+
+        /** \brief Return a termination condition that will become true as soon as the problem definition has an exact solution */
+        PlannerTerminationCondition exactSolnPlannerTerminationCondition(ompl::base::ProblemDefinitionPtr pdef);
```
mamoll: Are there planners that do not terminate with an exact solution when given an infinite amount of time? If not, then why do we need this function?

src/ompl/base/PlannerTerminationCondition.h
```diff
@@ -120,6 +121,32 @@ namespace ompl
...
+
+        /** \brief A class to run a planner for a specific number of iterations. Casts to a PTC for use with Planner::solve */
+        class IterationTerminationCondition
```
mamoll: A couple of concerns:
- I am worried that people will compare different algorithms as a function of number of iterations. Since different algorithms can do wildly different things within an iteration (one alg. can take 1ms per iteration, while another takes 1s), this comparison is not meaningful.
- Why not derive from PlannerTerminationCondition? I guess you don't want it to run in a separate thread?

src/ompl/base/SpaceInformation.h
```diff
@@ -363,6 +363,12 @@ namespace ompl
             */
             unsigned int getMotionStates(const State *s1, const State *s2, std::vector<State*> &states, unsigned int count, bool endpoints, bool alloc) const;
 
+            /** \brief Get the total number of motion segments checked by the MotionValidator so far */
+            unsigned int getNumMotionsChecked() const
```
mamoll: Rename to getCheckedMotionCount for consistency?

src/ompl/base/OptimizationObjective.h
```diff
@@ -85,9 +89,33 @@ namespace ompl
             /** \brief Set the cost threshold for objective satisfaction. When a path is found with a cost better than the cost threshold, the objective is considered satisfied. */
             void setCostThreshold(Cost c);
 
+            /** \brief Get the cost that corresponds to an entire path. This implementation assumes \e Path is of type \e PathGeometric.*/
+            virtual Cost getCost(const Path &path) const;
```
mamoll: There is already is a Path::cost function.

src/ompl/base/src/OptimizationObjective.cpp
mamoll: Yes, I like this better.

src/ompl/base/src/OptimizationObjective.cpp
```diff
@@ -65,16 +66,108 @@ void ompl::base::OptimizationObjective::setCostThreshold(Cost c)
...
+    if (isCostBetterThan(c1, c2))
+    {
+        return c1;
+    }
+    else
+    {
+        return c2;
+    }
+}
+
```
mamoll: Maybe simplify to: `return isCostBetterThan(c1, c2) ? c1 : c2`

src/ompl/base/OptimizationObjective.h
```diff
@@ -85,9 +89,33 @@ namespace ompl
...
+            virtual bool isCostBetterThanOrEquivalentTo(Cost c1, Cost c2) const;
+
+            /** \brief Compare whether cost \e c1 is worse or equivalent to cost \e c2. By default defined by !this->isCostBetterThan(c1,c2) as if c1 is not better than c2, than c1 is worse than, or equal to, c2*/
+            virtual bool isCostWorseThanOrEquivalentTo(Cost a, Cost b) const;
+
+            /** \brief Returns whether the cost is finite or not. By default calls std::isfinite on Cost::value(). */
+            virtual bool isFinite(Cost cost) const;
+
+            /** \brief Return the minimum cost given \e c1 and \e c2. Uses isCostBetterThan. */
+            virtual Cost minCost(Cost c1, Cost c2) const;
```
mamoll: rename minCost to bestCost since lower cost isn't always better?

mamoll:
Thanks for the pull request! I have lots of small comments. In general, I am trying to be conservative in adding new API functions. I am fine with the things I didn't comment on.

src/ompl/base/PlannerTerminationCondition.h
gammell: So I put this in for anytime asymptotically optimal planners, which in the infinite-time case would never return. I was actually using it to investigate the quality of the first solution found by these planners, but it may also be useful in conjunction with and/or PTCs to make more complex halting conditions?

It's your call.

src/ompl/base/PlannerTerminationCondition.h
gammell: Maybe this needs a better name? It's not truly number of iterations, it's the number of calls to the PTC, which, depending on how a specific planner uses it, may not map 1:1 to iterations.

I used this for inspecting how a planner was behaving and to make videos of the search process. You're absolutely right that it would not be useful in comparing planners across iterations, but I do think it's useful to be able to look at a planner iteration by iteration...

I don't remember why I implemented it like this. To be honest, it may have been the only way I could figure out how to do it. :worried:

src/ompl/base/OptimizationObjective.h
gammell: Whoops, sorry, I have no idea how I missed that.

src/ompl/base/src/OptimizationObjective.cpp
gammell: We have different ideas of "simplify". :wink:

gammell:
To be honest, I am indifferent about most of the "syntactic sugar" changes, so I've undone them to your preference. I have them defined as helper functions in my class anyway and was planning to keep it that way (I have some classes that order on cost, and I'd rather have a "pure" less operator without the margin).

Other comments above.

src/ompl/base/PlannerTerminationCondition.h
giogadi: Just chiming in on this: I completely agree with Mark about the risk of people trying to compare iteration counts of algorithms, but in every OMPL project I've worked on I've needed to implement my own iteration-based PTC. The main reason is because an iteration-based, non-threaded PTC enables deterministic planner output given a particular random seed. This is extremely useful for testing.

src/ompl/base/PlannerTerminationCondition.h
mamoll: You can get cost-optimizing planners to terminate by setting the cost threshold in the optimization objective to infinite cost. You can also just the planner progress properties of the benchmark tools (written by @giogadi) to figure out the cost of the first solution as well as the convergence rate to the optimal solution (with a few sql queries to the benchmark database).

src/ompl/base/PlannerTerminationCondition.h
mamoll: How do you use this non-threaded PTC? The standard Planner::solve API seems to require something derived from PlannerTerminationCondition, which seems to always check the PTC periodically in a separate thread. Maybe there is a place for something like this, but not in ompl::base. Maybe there should some diagnostic/development tools in ompl::tools to help with getting deterministic behavior.

src/ompl/base/PlannerTerminationCondition.h
gammell: Oh yeah, I never thought of that.

There may be some use in conjunction with an and/or PTC (i.e., setting a *minimum* run time by ANDing with a timedPTC), but I can't vouch for whether there's any actual need for that.

As I said, I'm not overly attached to any of these, so it's whatever you want. :smile:

src/ompl/base/PlannerTerminationCondition.h
gammell: The () operator returns a PTC. So I used it like:

```C++
IterationTerminationCondition myPTC(1u)
ompl::base::Planner myPlanner;
... //Setup problem, planner etc.
while (...)
{
    myPlanner.solve(myPTC); //Run planner for 1 call to the PTC
    ... //Save planner status to file, etc.
    myPTC.reset(); //Reset the PTC to allow another iteration
}
```
I think I'm commenting concurrently, so I will stop for now, but I can move it anywhere you want (including /dev/null :wink:).

PR #50, opened 2015-05-21
BIT*

gammell:
Hopefully this PR is easier, as it's a stand-alone planner.
This is BIT* which will be presented at ICRA next week, where I will hopefully say it's available in OMPL. :wink:.

You can find a video explaining it on [YouTube] (https://www.youtube.com/watch?v=TQIoCC48gp4) and a preprint of the paper on [arXiv] (http://arxiv.org/abs/1405.5848).

Thanks,
Jon

### DIFF
--- a/src/ompl/geometric/planners/bitstar/BITstar.h
+++ b/src/ompl/geometric/planners/bitstar/BITstar.h
@@ -1,0 +1,614 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2014, University of Toronto
+*  All rights reserved.
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of the University of Toronto nor the names of its
+*     contributors may be used to endorse or promote products derived
+*     from this software without specific prior written permission.
+*
+*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+*  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+*  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+*  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+*  POSSIBILITY OF SUCH DAMAGE.
+*********************************************************************/
+
+/* Authors: Jonathan Gammell */
+
+#ifndef OMPL_GEOMETRIC_PLANNERS_BITSTAR_BITSTAR_
+#define OMPL_GEOMETRIC_PLANNERS_BITSTAR_BITSTAR_
+
+//STL:
+//std::string
+#include <string>
+//std::pair
+#include <utility>
+//std::vector
+#include <vector>
+
+//OMPL:
+//The base-class of planners:
+#include "ompl/base/Planner.h"
+//The nearest neighbours structure
+#include "ompl/datastructures/NearestNeighbors.h"
+//The informed sampler structure
+#include "ompl/base/samplers/InformedStateSampler.h"
+//Planner includes:
+//#include "ompl/geometric/planners/PlannerIncludes.h"
+
+//BIT*:
+//The helper data classes, Vertex.h and IntegratedQueue.h are included *after* the declaration of the BITstar class as they are member classes of BITstar.
+
+
+
+namespace ompl
+{
+    namespace geometric
+    {
+        /**
+            @anchor gBITstar
+            @par Short description
+            BIT* (Batch Informed Trees) is an anytime asymptotically optimal sampling-based
+            motion planning algorithm that extends Lifelong Planning A* (LPA*) techniques to continuous planning
+            problems. BIT* accomplishes this by processing batches of samples with a heuristic.
+            In doing so, it strikes a balance between algorithms like RRT* and FMT*.
+
+            @par J D. Gammell, S. S. Srinivasa, T. D. Barfoot, "Batch Informed Trees (BIT*): Sampling-based Optimal Planning via the Heuristically Guided Search of Implicit Random Geometric Graphs,"
+            To appear at ICRA 2015. <a href="http://arxiv.org/abs/1405.5848">arXiv:1405.5848 [cs.RO]</a>.
+            <a href="http://www.youtube.com/watch?v=MRzSfLpNBmA">Illustration video</a>.
+
+            @par TODO:
+            - Make k-nearest correct.
+            - Extend beyond single goal states to other samplable goals (i.e., goal sets).
+            - Generalize heuristics to make proper use of the optimization class.
+        */
+        /** \brief Batch Informed Trees (BIT*)*/
+        class BITstar : public ompl::base::Planner
+        {
+        public:
+            //Forward declarations so that the classes belong to BIT*:
+            /** \brief The vertex of implicit and explicit graphs. */
+            class Vertex;
+            /** \brief A generator of unique vertex IDs. */
+            class IdGenerator;
+            /** \brief The queue of edges to process as an integrated dual-stage queue (tracks both the expansion of vertices and the resulting edges) */
+            class IntegratedQueue;
+            //Helpful typedefs:
+            /** \brief A vertex shared pointer. */
+            typedef boost::shared_ptr<Vertex> VertexPtr;
+            /** \brief A \e constant vertex shared pointer. */
+            typedef boost::shared_ptr<const Vertex> VertexConstPtr;
+            /** \brief A vertex weak pointer. */
+            typedef boost::weak_ptr<Vertex> VertexWeakPtr;
+            /** \brief An integrated queue shared pointer. */
+            typedef boost::shared_ptr<IntegratedQueue> IntegratedQueuePtr;
+            /** \brief The vertex id type */
+            typedef unsigned int vid_t;
+
+            /** \brief Construct! */
+            BITstar(const base::SpaceInformationPtr& si, const std::string& name = "BITstar");
+
+            /** \brief Destruct! */
+            virtual ~BITstar();
+
+            /** \brief Setup */
+            virtual void setup();
+
+            /** \brief Clear */
+            virtual void clear();
+
+            /** \brief Solve */
+            base::PlannerStatus solve(const base::PlannerTerminationCondition& ptc);
+
+            /** \brief Get results */
+            virtual void getPlannerData(base::PlannerData& data) const;
+
+            /** \brief Get the next edge to be processed. Causes vertices in the queue to be expanded (if necessary) and therefore effects the run timings of the algorithm, but helpful for some videos and debugging. */
+            std::pair<ompl::base::State const*, ompl::base::State const*> getNextEdgeInQueue();
+
+            /** \brief Get the value of the next edge to be processed. Causes vertices in the queue to be expanded (if necessary) and therefore effects the run timings of the algorithm, but helpful for some videos and debugging. */
+            ompl::base::Cost getNextEdgeValueInQueue();
+
+            /** \brief Get the whole messy set of edges in the queue. Expensive but helpful for some videos */
+            void getEdgeQueue(std::vector<std::pair<VertexConstPtr, VertexConstPtr> >* edgesInQueue);
+
+            /** \brief Get the whole set of vertices to be expanded. Expensive but helpful for some videos */
+            void getVertexQueue(std::vector<VertexConstPtr>* verticesInQueue);
+
+            /** \brief Set a different nearest neighbours datastructure */
+            template<template<typename T> class NN>
+            void setNearestNeighbors();
+            ///////////////////////////////////////
+            // Planner settings:
+            /** \brief Set the rewiring scale factor, s, such that r_rrg = s \times r_rrg* */
+            void setRewireFactor(double rewireFactor);
+
+            /** \brief Get the rewiring scale factor. */
+            double getRewireFactor() const;
+
+            /** \brief Set the number of samplers per batch. */
+            void setSamplesPerBatch(unsigned int n);
+
+            /** \brief Get the number of samplers per batch. */
+            unsigned int getSamplesPerBatch() const;
+
+            /** \brief Enable a k-nearest search for instead of an r-disc search. */
+            void setKNearest(bool useKNearest);
+
+            /** \brief Get whether a k-nearest search is being used.*/
+            bool getKNearest() const;
+
+            /** \brief Enable tracking of failed edges. This currently is too expensive to be useful.*/
+            void setUseFailureTracking(bool trackFailures);
+
+            /** \brief Get whether a failed edge list is in use.*/
+            bool getUseFailureTracking() const;
+
+            /** \brief Enable "strict sorting" of the edge queue.
+            Rewirings can change the position in the queue of an edge.
+            When strict sorting is enabled, the effected edges are resorted
+            immediately, while disabling strict sorting delays this
+            resorting until the end of the batch. */
+            void setStrictQueueOrdering(bool beStrict);
+
+            /** \brief Get whether strict queue ordering is in use*/
+            bool getStrictQueueOrdering() const;
+
+            /** \brief Enable pruning of vertices/samples that CANNOT improve the current solution.
+            When a vertex in the graph is pruned, it's descendents are also pruned
+            (if they also cannot improve the solution) or placed back in
+            the set of free samples (if they could improve the solution).
+            This assures that a uniform density is maintained.*/
+            void setPruning(bool prune);
+
+            /** \brief Get whether graph and sample pruning is in use.*/
+            bool getPruning() const;
+
+            /** \brief Set the fractional change in the solution cost necessary for pruning to occur. */
+            void setPruneThresholdFraction(double fractionalChange);
+
+            /** \brief Get the fractional change in the solution cost necessary for pruning to occur. */
+            double getPruneThresholdFraction() const;
+
+            /** \brief Delay considering rewiring edges until an initial solution is found. This improves
+            the time required to find an initial solution when doing so requires multiple batches and has
+            no effects on theoretical asymptotic optimality (as the rewiring edges are eventually considered). */
+            void setDelayRewiringUntilInitialSolution(bool delayRewiring);
+
+            /** \brief Get whether BIT* is delaying rewiring until a solution is found. */
+            bool getDelayRewiringUntilInitialSolution() const;
+
+            /** \brief Stop the planner each time a solution improvement is found. Useful
+            for examining the intermediate solutions found by BIT*. */
+            void setStopOnSolnImprovement(bool stopOnChange);
+
+            /** \brief Get whether BIT* stops each time a solution is found. */
+            bool getStopOnSolnImprovement() const;
+            ///////////////////////////////////////
+
+            ///////////////////////////////////////
+            // Planner progress property functions
+            /** \brief Retrieve the best exact-solution cost found
+            as the raw data. (bestCost_) */
+            ompl::base::Cost bestCost() const;
+            /** \brief Retrieve the best exact-solution cost found
+            as a planner-progress property. (bestCost_) */
+            std::string bestCostProgressProperty() const;
+
+            /** \brief Retrieve the length of the best exact-solution found
+            as a planner-progress property. (bestLength_) */
+            std::string bestLengthProgressProperty() const;
+
+            /** \brief Retrieve the current number of free samples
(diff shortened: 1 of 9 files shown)


src/ompl/geometric/planners/bitstar/BITstar.h
```diff
@@ -0,0 +1,614 @@
...
+            virtual void clear();
+
+            /** \brief Solve */
+            base::PlannerStatus solve(const base::PlannerTerminationCondition& ptc);
+
+            /** \brief Get results */
+            virtual void getPlannerData(base::PlannerData& data) const;
+
+            /** \brief Get the next edge to be processed. Causes vertices in the queue to be expanded (if necessary) and therefore effects the run timings of the algorithm, but helpful for some videos and debugging. */
+            std::pair<ompl::base::State const*, ompl::base::State const*> getNextEdgeInQueue();
```
isucan: Does this need to be public?

src/ompl/geometric/planners/bitstar/BITstar.h
```diff
@@ -0,0 +1,614 @@
...
+
+            /** \brief Get the value of the next edge to be processed. Causes vertices in the queue to be expanded (if necessary) and therefore effects the run timings of the algorithm, but helpful for some videos and debugging. */
+            ompl::base::Cost getNextEdgeValueInQueue();
```
isucan: can this be private?

isucan: Ah, I now see your comments about debugging

src/ompl/geometric/planners/bitstar/BITstar.h
isucan: Any reason not to use const as prefix?

src/ompl/geometric/planners/bitstar/BITstar.h
```diff
@@ -0,0 +1,614 @@
...
+            /** \brief Stop the planner each time a solution improvement is found. Useful
+            for examining the intermediate solutions found by BIT*. */
+            void setStopOnSolnImprovement(bool stopOnChange);
+
+            /** \brief Get whether BIT* stops each time a solution is found. */
+            bool getStopOnSolnImprovement() const;
+            ///////////////////////////////////////
+
+            ///////////////////////////////////////
+            // Planner progress property functions
```
isucan: Can we make the progress property functions private?

src/ompl/geometric/planners/bitstar/BITstar.h
```diff
@@ -0,0 +1,614 @@
...
+
+        protected:
+            //Everything is only protected so we can create modifications without duplicating code by deriving from the class:
+
+            //Typedefs:
+            /** \brief A pair of vertices, i.e., an edge. */
+            typedef std::pair<VertexPtr, VertexPtr> vertex_pair_t;
+
+            /** \brief The OMPL::NearestNeighbors structure. */
+            typedef boost::shared_ptr< NearestNeighbors<VertexPtr> > vertex_nn_ptr_t;
```
isucan: The type name used here is not consistent with other types used above. Please use the Ptr suffix and a CamelCase name

src/ompl/geometric/planners/bitstar/BITstar.h
```diff
@@ -0,0 +1,614 @@
...
+            as a planner-progress property. (numNearestNeighbours_) */
+            std::string nearestNeighbourProgressProperty() const;
+            ///////////////////////////////////////
+
+        protected:
+            //Everything is only protected so we can create modifications without duplicating code by deriving from the class:
+
+            //Typedefs:
+            /** \brief A pair of vertices, i.e., an edge. */
+            typedef std::pair<VertexPtr, VertexPtr> vertex_pair_t;
```
isucan: Rename type to VertexPair?

src/ompl/geometric/planners/bitstar/BITstar.h
```diff
@@ -0,0 +1,614 @@
...
+            ///////////////////////////////////////////////////////////////////
+            //Helper functions for sorting queues/nearest-neighbour structures and the related calculations.
+            /** \brief The distance function used for nearest neighbours. Calculates the distance directionally from the given state to all the other states (can be used on states either in our out of the graph).*/
+            double nnDistance(const VertexPtr& a, const VertexPtr& b) const;
+            ///////////////////////////////////////////////////////////////////
+
+            ///////////////////////////////////////////////////////////////////
+            //Helper functions for various heuristics.
+            /** \brief Calculates a heuristic estimate of the cost of a solution constrained to pass through a vertex, independent of the current cost-to-come. I.e., combines the heuristic estimates of the cost-to-come and cost-to-go. */
+            ompl::base::Cost lowerBoundHeuristicVertex(const VertexPtr& edgePair) const;
```
isucan: You could remove "ompl::" if you wanted. That's how we do it elsewhere, but I am not too particular about this one.

src/ompl/geometric/planners/bitstar/BITstar.h
```diff
@@ -0,0 +1,614 @@
...
+            ///////////////////////////////////////////////////////////////////
+            //Helper functions to calculate parameters:
+            /** \brief Initialize the nearest-neighbour terms */
+            void initializeNearestTerms();
+
+            /** \brief Update the appropriate nearest-neighbour terms, r_ and k_ */
+            virtual void updateNearestTerms();
+
+            /** \brief Calculate the r for r-disc nearest neighbours, a function of the current graph */
+            double r(unsigned int N) const;
```
isucan: Please use longer function names for this one and the one below. Even if it is something like rConstant

src/ompl/geometric/planners/bitstar/BITstar.h
```diff
@@ -0,0 +1,614 @@
...
+            bool                                                     stopOnSolnChange_;
+            ///////////////////////////////////////
+        }; //class: BITstar
+    } //geometric
+} //ompl
+
+
+//BIT* Includes:
+//The Vertex ID generator class
+#include "ompl/geometric/planners/bitstar/datastructures/IdGenerator.h"
```
isucan: These seem not to be neded for this header file, since they are included here. Mind if we just include them in the .cpp?

src/ompl/geometric/planners/bitstar/datastructures/IdGenerator.h
```diff
@@ -0,0 +1,79 @@
...
+        /** \brief An ID generator class for vertex IDs.*/
+        class BITstar::IdGenerator
+        {
+        public:
+            /** \brief Generator a new id and increment the global/static counter of IDs. */
+            static BITstar::vid_t getNewId()
+            {
+                //STATIC variables:
+                //The next ID to be returned:
+                static BITstar::vid_t nextId = 0u;
```
isucan: I would prefer this initialization in the cpp. If someone changes the typedef from int to something else this won't work for c++ < 11. Furthermore, please use the style of locking in StateSpace and RandomNumbers instead of non-POD statics.

src/ompl/geometric/planners/bitstar/datastructures/IntegratedQueue.h
```diff
@@ -0,0 +1,453 @@
...
+        */
+
+        /** \brief A queue of edges to be processed that integrates both the expansion of \ref gVertex "Vertices" and the ordering of the resulting edges. */
+        class BITstar::IntegratedQueue
+        {
+        public:
+            ////////////////////////////////
+            //Data typedefs:
+            /** \brief A typedef for a pair of vertices, i.e., an edge */
+            typedef std::pair<VertexPtr, VertexPtr> vertex_pair_t;
```
isucan: I would prefer CamelCase type names :)

src/ompl/geometric/planners/bitstar/datastructures/Vertex.h
```diff
@@ -0,0 +1,227 @@
...
+
+/* Authors: Jonathan Gammell */
+
+#ifndef OMPL_GEOMETRIC_PLANNERS_BITSTAR_DATASTRUCTURES_VERTEX_
+#define OMPL_GEOMETRIC_PLANNERS_BITSTAR_DATASTRUCTURES_VERTEX_
+
+//vector
+#include <vector>
+//std::set and std::multiset
+//#include <set>
```
isucan: remove comment?

src/ompl/geometric/planners/bitstar/datastructures/src/IntegratedQueue.cpp
```diff
@@ -0,0 +1,1693 @@
...
+*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+*  POSSIBILITY OF SUCH DAMAGE.
+*********************************************************************/
+
+/* Authors: Jonathan Gammell */
+
+//Myself:
```
isucan: I don't think this distinction is needed any longer: this code is becoming part of OMPL :)

src/ompl/geometric/planners/bitstar/datastructures/src/IntegratedQueue.cpp
```diff
@@ -0,0 +1,1693 @@
...
+                    expand = false;
+                }
+
+
+//                //Prune back any edges in the front that have previously failed:
+//                if (useFailureTracking_ == true)
+//                {
+//                    if (edgeQueue_.begin()->second.first->hasAlreadyFailed(edgeQueue_.begin()->second.second) == true)
+//                    {
+//                        //Remove the edge from the queue:
```
isucan: Please remove comments. There is also quite some white space between functions. We usually use just one empty line.

src/ompl/geometric/planners/bitstar/datastructures/src/IntegratedQueue.cpp
```diff
@@ -0,0 +1,1693 @@
...
+
+                //Remove from lookups map as requested
+                if (removeLookups == true)
+                {
+                    vertexIterLookup_.erase(lookupIter);
+                    this->removeEdgesFrom(oldVertex);
+                }
+
+                //Check if I have been given permission to change sets:
+                if (bool(vertexNN) == true && bool(freeStateNN) == true)
```
isucan: please call static_cast explicitly for casting, rather than C-style cast

src/ompl/geometric/planners/bitstar/datastructures/src/IntegratedQueue.cpp
```diff
@@ -0,0 +1,1693 @@
...
+                bool found;
+                //The iterator to the mmapIterToRm in my list:
+                edge_queue_iter_list_t::iterator iterToList;
+
+                //Start at the front:
+                iterToList = iterToVertexListPair->second.begin();
+
+                //Iterate through the list and find mmapIterToRm
+                found = false;
+                while(found == false && iterToList != iterToVertexListPair->second.end())
```
isucan: white space after while

src/ompl/geometric/planners/bitstar/datastructures/src/IntegratedQueue.cpp
```diff
@@ -0,0 +1,1693 @@
...
+
+
+
+
+
+
+
+
+
+
```
isucan: This is what I mean by lots of empty space :)

src/ompl/geometric/planners/bitstar/src/BITstar.cpp
```diff
@@ -0,0 +1,1913 @@
...
+
+            //Register my progress info:
+            addPlannerProgressProperty("best cost DOUBLE", boost::bind(&BITstar::bestCostProgressProperty, this));
+            addPlannerProgressProperty("number of segments in solution path INTEGER", boost::bind(&BITstar::bestLengthProgressProperty, this));
+            addPlannerProgressProperty("current free states INTEGER", boost::bind(&BITstar::currentFreeProgressProperty, this));
+            addPlannerProgressProperty("current vertices INTEGER", boost::bind(&BITstar::currentVertexProgressProperty, this));
+            //addPlannerProgressProperty("vertex queue size INTEGER", boost::bind(&BITstar::vertexQueueSizeProgressProperty, this));
+            //addPlannerProgressProperty("edge queue size INTEGER", boost::bind(&BITstar::edgeQueueSizeProgressProperty, this));
+            //addPlannerProgressProperty("iterations INTEGER", boost::bind(&BITstar::iterationProgressProperty, this));
+            //addPlannerProgressProperty("batches INTEGER", boost::bind(&BITstar::batchesProgressProperty, this));
```
isucan: Why comment these out? If they are a performance issue, please comment on this

src/ompl/geometric/planners/bitstar/src/BITstar.cpp
```diff
@@ -0,0 +1,1913 @@
...
+
+            //The list of vertices
+            if (bool(vertexNN_) == true)
+            {
+                vertexNN_->clear();
+                vertexNN_.reset();
+            }
+
+            //The queue:
+            if (bool(intQueue_) == true)
```
isucan: Please use static_cast, here and elsewhere

src/ompl/geometric/planners/bitstar/src/BITstar.cpp
```diff
@@ -0,0 +1,1913 @@
...
+            }
+            //No else.
+
+            //Did we find a solution?
+            if (hasSolution_ == true)
+            {
+                data.markGoalState(goalVertex_->stateConst());
+            }
+
+            data.properties["best_solution_cost DOUBLE"] = this->bestCostProgressProperty();
```
isucan: I think there's a function to automatically add all the registered properties

isucan: we should not duplicate this

src/ompl/geometric/planners/bitstar/src/BITstar.cpp
```diff
@@ -0,0 +1,1913 @@
...
+
+                //Sample the shell of the state space that many times, adding the valid ones to the list of free states.
+                for (unsigned int i = 0u; i < uintNumSamples; ++i)
+                {
+                    //Variable
+                    //The new state:
+                    VertexPtr newState = boost::make_shared<Vertex>(Planner::si_, opt_);
+
+                    //Sample:
+                    sampler_->sampleUniformIgnoreBounds(newState->state(), costSampled_, reqCost);
```
isucan: remove comments?

isucan:
Thank you for your contribution! This looks great!
Note that I only checked style (please see my comments). I did not check this for correctness. 
+1 for merging (provided comments are addressed)

src/ompl/geometric/planners/bitstar/BITstar.h
gammell: Yes, it needs to be public, it's a function for making visualizations (and debugging).

I personally find `type const*` clearer, as when read right-to-left, you get "pointer to const type", and `type* const` gives "const pointer to type".

Will change tho'.

src/ompl/geometric/planners/bitstar/BITstar.h
gammell: Yup, visualizations (and debugging).

src/ompl/geometric/planners/bitstar/BITstar.h
gammell: Does that work? I followed RRT* as an example for progress properties, and there they're public...

src/ompl/geometric/planners/bitstar/BITstar.h
gammell: I personally find that "absolute" namespaces make code from large projects easier to understand (especially at a glance)... and I don't mind doing the extra typing. :smile:

src/ompl/geometric/planners/bitstar/BITstar.h
gammell: I'm not sure if that would work? I had a hard time defining member classes in separate files.

I don't remember all the details, but I know the only way I found to do it was forward decalre the classes in `BITstar` before including the class definitions; otherwise, I would have had these at the top `BITstar.h`...

I guess they'd work in the .cpp, but then when you include BITstar.h somewhere else, you'd be left with incomplete types for types that `BITstar` returns via public functions...

src/ompl/geometric/planners/bitstar/datastructures/src/IntegratedQueue.cpp
gammell: This is in the "voice of the class" :wink:.
It's just my default comment for a `.cpp` including its `.h`. I try to hold myself to a 1:1 ratio of comments on includes to explain their purpose. I guess I have a fairly verbose programming style.

src/ompl/geometric/planners/bitstar/datastructures/src/IntegratedQueue.cpp
gammell: Right.
So I do try to be consistent with white space, so when there are large gaps like this, it isn't out of laziness/sloppiness. My logic is as follows:
 - 1 line of white space inside functions. I find this helps break up the code and make things more readable.
 - 3 lines of white space between functions. This is to help identify where functions start and stop as there are single white space lines inside functions. I find this especially helpful when scrolling around in a file.
 - A bunch of white space between "groups" of functions. So for example between public and private functions (what this is), or before a bunch of really boring sets/gets (which I always try to hide at the bottom of a file).

src/ompl/geometric/planners/bitstar/src/BITstar.cpp
gammell: No, it's not really a performance issue. I just was trying to clean up the interface before putting it into the mainline. This was the same reason I removed the Python calls to some of the parameters above. They're almost definitely not necessary for anyone...

All these properties are still calculate/tracked, so if someone was interested in the info uncommenting these lines would be sufficient. I'll add a comment explaining this.

src/ompl/geometric/planners/bitstar/src/BITstar.cpp
gammell: Oh, is this not necessary? I assumed I'd copied this from `RRT*` as an example, but that does not appear to be the case (at least anymore), so I'm not sure where I got this idea from... I did think it was a little weird that I had to do it twice. :wink:

src/ompl/geometric/planners/bitstar/datastructures/IdGenerator.h
gammell: I have some new work that defines a different type of vertex, so I wanted to have the ID generator in a single place so I didn't have to duplicate the code into two separate `.cpp` files. That's why it's in it's own `.h` file.

I'm not sure why changing from `int` to another class wouldn't work as long as the other class has an assignment operator that takes `0u` and the preincrement operator. But to be honest, it wasn't necessarily my goal to make it an ID generator for any arbitrary type, I just found it made clearer code to make variables of type `VertexId` instead of just `unsigned int`. So I'm not sure if it's inability to handle arbitrary types is a problem?

I've tried to copy the locking style in `RandomNumbers`, but I'm not sure I fully understand what I've done. :wink:

src/ompl/geometric/planners/bitstar/datastructures/IntegratedQueue.h
gammell: For some reason I thought this was the style recommended in the Google C++ Style Guide, but I was incorrect. Fixed.

gammell:
Ok, it took me a few tries to get the changed `IdGenerator` to link properly, but I think this addresses all of @isucan's comments, other than white space and a couple other comments, which hopefully I've sufficiently justified.

src/ompl/geometric/planners/bitstar/BITstar.h
isucan: Yes, it does work. We should really change that in RRT*, I realized that was what you were following. Anyway, no big deal :)

src/ompl/geometric/planners/bitstar/BITstar.h
isucan: Sounds good to me! (I agree with your argument)

src/ompl/geometric/planners/bitstar/BITstar.h
isucan: In that case, please keep the headers included at the top of the file?

src/ompl/geometric/planners/bitstar/datastructures/src/IntegratedQueue.cpp
isucan: I see. I got confused, Maybe say "This class"? I think that would be clearer.
+1 for comments

src/ompl/geometric/planners/bitstar/datastructures/src/IntegratedQueue.cpp
```diff
@@ -0,0 +1,1679 @@
...
+                vertexToExpand_( vertexQueue_.begin() ),
+                edgeQueue_( boost::bind(&BITstar::IntegratedQueue::edgeQueueComparison, this, _1, _2) ), //This tells the edgeQueue_ to use the edgeQueueComparison for sorting
+                vertexIterLookup_(),
+                outgoingEdges_(),
+                incomingEdges_(),
+                resortVertices_(),
+                costThreshold_( std::numeric_limits<double>::infinity() ), //Purposeful gibberish
+                hasSolution_(false)
+        {
+            //The cost threshold:
```
isucan: One nitpick: In general, if the comment does not add much to what I would get from the variable name, I prefer to not add it.

src/ompl/geometric/planners/bitstar/datastructures/src/IntegratedQueue.cpp
isucan: I see your point, but I do not fully subscribe to the rules your propose. 1 space for functions, if you have different parts, makes sense sometimes, but I would expect that together with a comment maybe?
I think 1 line between functions is ample, since you get an extra almost empty line with "}" above. That said, I do not mind at all adding more space, so I can be convinced your rule is better. However, for consistency with the rest of OMPL it might be easier to follow the same style. I am not asking for this just to be particular, but if we follow a consistent style it is easier for others to read & understand the code.
Lastly, a bunch of space between groups of functions is not sufficiently well defined. Maybe add one empty line with a comment with a an additional commented marker make of - or = or / ? Too much empty space will make think the file is done :)

src/ompl/geometric/planners/bitstar/BITstar.h
gammell: I wish I could... The issue is that these three classes are defined as *member* classes of the `BITstar` class. If they're included before it is defined, the keyword `BITstar` will be unknown... When I tried to forward declare `BITstar` in the member classes, all hell broke loose, and I really didn't want to define these classes inline in this file (like some other planners do with their `Motion` class)...

There is a comment at the top of the `BITstar.h` explaining why three more includes are at the bottom... I figured it sort of mirrored what's done with the implementation file for a templated class...

gammell:
Ok, I've cleaned up the whitespace a bit more by replacing it with comments that explain the function groupings. The only change I didn't make was moving the member class includes to the top of `BITstar.h`, for the reasons discussed above.

isucan:
Looks good to me, +1 for merging

PR #52, opened 2015-06-15
BITstar demo

gammell:
Here is an attempt at making a demo for optimal planning that uses BIT\*, FMT\*, RRT\* and PRM\*. I used the existing optimization objectives that were available in the demo.

I was also able to isolate the ifdef for Eigen to only the direct path-length sampler.

FYI, I don't really know Python,
Jon

### DIFF
--- a/src/ompl/CMakeLists.txt
+++ b/src/ompl/CMakeLists.txt
@@ -15,10 +15,7 @@ file(GLOB_RECURSE OMPL_SOURCE_CODE datastructures/*.cpp util/*.cpp base/*.cpp ge
 if(NOT OMPL_HAVE_EIGEN3)
     list(REMOVE_ITEM OMPL_SOURCE_CODE
         "${VERSION_DIR}/util/src/ProlateHyperspheroid.cpp"
-        "${VERSION_DIR}/base/samplers/informed/src/PathLengthDirectInfSampler.cpp"
-        "${VERSION_DIR}/geometric/planners/bitstar/src/BITstar.cpp"
-        "${VERSION_DIR}/geometric/planners/bitstar/datastructures/src/IntegratedQueue.cpp"
-        "${VERSION_DIR}/geometric/planners/bitstar/datastructures/src/Vertex.cpp")
+        "${VERSION_DIR}/base/samplers/informed/src/PathLengthDirectInfSampler.cpp")
 endif()
 
 #############################################
--- a/src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
+++ b/src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
@@ -38,6 +38,8 @@
 #include <boost/make_shared.hpp>
 #if OMPL_HAVE_EIGEN3
 #include "ompl/base/samplers/informed/PathLengthDirectInfSampler.h"
+#else
+#include "ompl/base/samplers/informed/RejectionInfSampler.h"
 #endif
 
 ompl::base::PathLengthOptimizationObjective::
--- a/src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
+++ b/src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
@@ -70,7 +70,7 @@ ompl::base::InformedSamplerPtr ompl::base::PathLengthOptimizationObjective::allo
 #if OMPL_HAVE_EIGEN3
     return boost::make_shared<PathLengthDirectInfSampler>(probDefn, maxNumberCalls);
 #else
-    OMPL_ERROR("%s: Direct sampling of the path-length objective requires Eigen, but OMPL was compiled without Eigen support. If possible, please install Eigen and recompile OMPL. If this is not possible, you can manually create an instantiation of RejectionInfSampler to approximate the behaviour of direct informed sampling. ", description_.c_str());
+    throw Exception("Direct sampling of the path-length objective requires Eigen, but this version of OMPL was compiled without Eigen support. If possible, please install Eigen and recompile OMPL. If this is not possible, you can manually create an instantiation of RejectionInfSampler to approximate the behaviour of direct informed sampling.");
     // Return a null pointer to avoid compiler warnings
     return ompl::base::InformedSamplerPtr();
 #endif
--- a/src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
+++ b/src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
@@ -70,7 +70,8 @@ ompl::base::InformedSamplerPtr ompl::base::PathLengthOptimizationObjective::allo
 #if OMPL_HAVE_EIGEN3
     return boost::make_shared<PathLengthDirectInfSampler>(probDefn, maxNumberCalls);
 #else
-    OMPL_INFORM("%s: Direct sampling of the path-length objective requires Eigen. As this version of OMPL was compiled without Eigen support, the informed sampling is defaulting to rejection sampling.", description_.c_str());
-    return boost::make_shared<RejectionInfSampler>(probDefn, maxNumberCalls);
+    OMPL_ERROR("%s: Direct sampling of the path-length objective requires Eigen, but OMPL was compiled without Eigen support. If possible, please install Eigen and recompile OMPL. If this is not possible, you can manually create an instantiation of RejectionInfSampler to approximate the behaviour of direct informed sampling. ", description_.c_str());
+    // Return a null pointer to avoid compiler warnings
+    return ompl::base::InformedSamplerPtr();
 #endif
 }
--- a/src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
+++ b/src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
@@ -35,10 +35,10 @@
 /* Author: Luis G. Torres, Jonathan Gammell */
 
 #include "ompl/base/objectives/PathLengthOptimizationObjective.h"
+#include <boost/make_shared.hpp>
 #if OMPL_HAVE_EIGEN3
 #include "ompl/base/samplers/informed/PathLengthDirectInfSampler.h"
 #endif
-#include <boost/make_shared.hpp>
 
 ompl::base::PathLengthOptimizationObjective::
 PathLengthOptimizationObjective(const SpaceInformationPtr &si) :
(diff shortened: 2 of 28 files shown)


gammell:
So looking at the build logs, everything compiled fine but some tests failed.
I've seen the same behaviour with these tests on my computer, I think from before I started making changes?

Debug build (Job 12.1):

    Start 13: test_2denvs_geometric
    13/24 Test #13: test_2denvs_geometric .............***Failed  386.88 sec

Release build (Job 12.2):

    Start 13: test_2denvs_geometric
    13/24 Test #13: test_2denvs_geometric .............***Failed   95.86 sec
    Start 16: test_2dcircles_opt_geometric
    16/24 Test #16: test_2dcircles_opt_geometric ......***Failed   19.83 sec

src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
```diff
@@ -62,10 +64,13 @@ ompl::base::Cost ompl::base::PathLengthOptimizationObjective::motionCostHeuristi
...
 ompl::base::InformedSamplerPtr ompl::base::PathLengthOptimizationObjective::allocInformedStateSampler(const ProblemDefinitionPtr probDefn, unsigned int maxNumberCalls) const
 {
-    // Make the direct path-length informed sampler and return
+    // Make the direct path-length informed sampler and return. If OMPL was compiled with Eigen, a direct version is available, if not a rejection-based technique can be used
+#if OMPL_HAVE_EIGEN3
     return boost::make_shared<PathLengthDirectInfSampler>(probDefn, maxNumberCalls);
-}
+#else
+    OMPL_INFORM("%s: Direct sampling of the path-length objective requires Eigen. As this version of OMPL was compiled without Eigen support, the informed sampling is defaulting to rejection sampling.", description_.c_str());
+    return boost::make_shared<RejectionInfSampler>(probDefn, maxNumberCalls);
```
isucan: I would prefer throwing an exception here, so the user is forced to really understand that what they want is missing. The exception message could recommend rejection sampling.

src/ompl/geometric/planners/bitstar/BITstar.h
```diff
@@ -382,6 +377,9 @@ namespace ompl
 
             /** \brief Calculate the lower-bounding k-nearest RGG term for asymptotic almost-sure convergence to the optimal path (i.e., k_rrg* in Karaman and Frazzoli IJRR 11). This is a function of the state dimension and is left as a double for later accuracy in calculate k */
             double minimumRggK() const;
+
+            /** \brief Calculate the measure (i.e., volume) of a unit n-ball */
+            double unitNBallMeasure(unsigned int N) const;
```
isucan: This looks like something that would be useful perhaps in other parts of the code. Could we move this function to its own file (say, util/sphere.h,cpp?)

src/ompl/CMakeLists.txt
```diff
@@ -15,10 +15,7 @@ file(GLOB_RECURSE OMPL_SOURCE_CODE datastructures/*.cpp util/*.cpp base/*.cpp ge
 if(NOT OMPL_HAVE_EIGEN3)
     list(REMOVE_ITEM OMPL_SOURCE_CODE
         "${VERSION_DIR}/util/src/ProlateHyperspheroid.cpp"
-        "${VERSION_DIR}/base/samplers/informed/src/PathLengthDirectInfSampler.cpp"
```
isucan: Very nice!

isucan:
+1 for merging. Only minor comments made.

src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
gammell: So my thought is that it is completely valid to use rejection sampling for path length if a user is in a situation where they **can't** have Eigen... I do agree; however, that if they **could** have Eigen and simply **don't**, it'd be better to forcefully suggest they install it. :wink:

So how about a preprocessor directive to acknowledge they understand and want to use rejection sampling? Then the else block for `OMPL_HAVE_EIGEN3` would either create a rejection sampler or throw an error describing the options (acknowledge this is what you want to do, or install Eigen).

I'm thinking something like they do in `boost::ublas`, e.g., requiring the user to do something like

    #define OMPL_ALLOW_REJECTION_SAMPLING_FOR_PATH_LENGTH_OBJECTIVE

in their file.

I believe this will require moving the implementation of this function to the header though, as it's compile-time. Is that a problem?

src/ompl/geometric/planners/bitstar/BITstar.h
gammell: Sure. I think there's also a couple similar simple geometric functions in the ProlateHyperspheroid class, so I'll make something like `util/geometric_equations.h`.

src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
isucan: I understand that is valid, but even that preprocessor directive, that will be set to some default value and the user does not know of it. This is a one time notification that forces the user to allocate a different sampler, or to install Eigen. This I think is much better than silent  approximate fixes which may not be what the user wants.

src/ompl/geometric/planners/bitstar/BITstar.h
isucan: Excellent!

src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
gammell: Hm, still thinking this through.
I guess if in the absence of Eigen `PathLengthOptimizationObjective::allocInformedStateSampler` throws, a user could still create a rejection sampler manually for path length.

I.e. instead of,

    InformedSamplerPtr infSampler = opt_->allocInformedSampler(...);

doing

    InformedSamplerPtr infSamplerboost::make_shared<RejectionInfSampler>(...);

I guess that leaves all the functionality available to a motivated user. The only problematic case I can think of would be in the InformedStateSampler wrapper class, which calls `opt_->allocInformedSampler(...)` by default, but an overloaded constructor could look after that?...

src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
```diff
@@ -62,10 +64,14 @@ ompl::base::Cost ompl::base::PathLengthOptimizationObjective::motionCostHeuristi
...
 {
-    // Make the direct path-length informed sampler and return
+    // Make the direct path-length informed sampler and return. If OMPL was compiled with Eigen, a direct version is available, if not a rejection-based technique can be used
+#if OMPL_HAVE_EIGEN3
     return boost::make_shared<PathLengthDirectInfSampler>(probDefn, maxNumberCalls);
-}
+#else
+    OMPL_ERROR("%s: Direct sampling of the path-length objective requires Eigen, but OMPL was compiled without Eigen support. If possible, please install Eigen and recompile OMPL. If this is not possible, you can manually create an instantiation of RejectionInfSampler to approximate the behaviour of direct informed sampling. ", description_.c_str());
+    // Return a null pointer to avoid compiler warnings
+    return ompl::base::InformedSamplerPtr();
```
gammell: Two questions:
- Is `OMPL_ERROR(...)` or `throw Exception(...)` preferred here?
- Instead of being a runtime erorr, is there a way to make this a compile-time error that only gets raised if (1) the function is used and (2) `!OMPL_HAVE_EIGEN3`? The only example I could find only was [here] (http://stackoverflow.com/questions/18813731/compile-time-check-if-a-function-is-used-unused-c), but it requires the use of a dummy template argument (which struck me as not worth the complexity?).

src/ompl/base/samplers/InformedStateSampler.h
```diff
@@ -111,9 +111,12 @@ namespace ompl
...
             typedef boost::function<Cost ()> GetCurrentCostFunc;
 
-            /** \brief Construct a sampler that only generates states with a heuristic solution estimate that is less than the cost of the current solution. Requires a function pointer to a method to query the cost of the current solution. */
+            /** \brief Construct a sampler that only generates states with a heuristic solution estimate that is less than the cost of the current solution using the default informed sampler for the current optimization objective. Requires a function pointer to a method to query the cost of the current solution. */
             InformedStateSampler(const ProblemDefinitionPtr &probDefn, unsigned int maxNumberCalls, const GetCurrentCostFunc &costFunc);
 
+            /** \brief Construct a sampler that only generates states with a heuristic solution estimate that is less than the cost of the current solution using the provided informed sampler. Requires a function pointer to a method to query the cost of the current solution. */
+            InformedStateSampler(const ProblemDefinitionPtr &probDefn, const GetCurrentCostFunc &costFunc, const InformedSamplerPtr &infSampler);
```
gammell: A second version of the wrapper that allows for the explicit definition of the InformedSampler to use.

src/ompl/base/objectives/src/PathLengthOptimizationObjective.cpp
mamoll: * I'd say throwing an exception is better, since nothing good will happen after the null pointer is returned.
 * A compile time error would be nicer, but a runtime error is fine.

src/ompl/geometric/planners/bitstar/src/BITstar.cpp
```diff
@@ -351,7 +353,7 @@ namespace ompl
         {
             Planner::checkValidity();
             OMPL_INFORM("%s: Searching for a solution to the given planning problem.", Planner::getName().c_str());
-            this->statusMessage(ompl::msg::LOG_DEBUG, "Start solve");
+            // this->statusMessage(ompl::msg::LOG_DEBUG, "Start solve");
```
mamoll: delete rather than comment out

src/ompl/CMakeLists.txt
mamoll: Make sure this doesn't break the python bindings. I can fix if necessary.

mamoll:
:+1: Looks good. I just made a few minor comments. Thanks!

src/ompl/CMakeLists.txt
gammell: Yup, I've run make `clean_bindings && make update_bindings` and it seems fine.
This diff is a little misleading, `PathLengthDirectInfSampler.cpp` is still in the list, leaving the only two files that are skipped in the absence of Eigen, `ProlateHyperspheroid.cpp`, which uses Eigen, and `PathLengthDirectInfSampler.cpp` which uses PHS.

gammell:
I modified README.md to add a token comment about Eigen.

gammell:
If it's all possible, I'd like to submit the RRT* modification to use Informed Sampling as well before the next release. Stupidly I already merged these changes into that branch though, so I can't make the next PR until this is merged (without being confusing). I'll make that new PR as soon as this is merged, and you can decide if there's enough time to merge it now or later.

mamoll:
This one is merged, so let's move on to the next one!

PR #56, opened 2015-07-01
Informed RRT*

gammell:
Ok,
So here is the changes to use informed sampling in RRT\* (i.e., Informed RRT\*). While I was at it, I cleaned a thing or two up in the RRT\* code, so I'll be going through to put some inline comments about those as well.

I'm working to "relax" informed sampling to handle multiple starts/goals (i.e., GOAL_STATE**S**), so I'll submit that after this PR, hopefully before the next release.

### DIFF
--- a/src/ompl/base/src/PlannerTerminationCondition.cpp
+++ b/src/ompl/base/src/PlannerTerminationCondition.cpp
@@ -222,7 +222,12 @@ ompl::base::PlannerTerminationCondition ompl::base::plannerAndTerminationConditi
 
 ompl::base::PlannerTerminationCondition ompl::base::timedPlannerTerminationCondition(double duration)
 {
-    return PlannerTerminationCondition(boost::bind(&timePassed, time::now() + time::seconds(duration)));
+    return timedPlannerTerminationCondition(time::seconds(duration));
+}
+
+ompl::base::PlannerTerminationCondition ompl::base::timedPlannerTerminationCondition(time::duration duration)
+{
+    return PlannerTerminationCondition(boost::bind(&timePassed, time::now() + duration));
 }
 
 ompl::base::PlannerTerminationCondition ompl::base::timedPlannerTerminationCondition(double duration, double interval)
--- a/src/ompl/geometric/planners/bitstar/BITstar.h
+++ b/src/ompl/geometric/planners/bitstar/BITstar.h
@@ -75,15 +75,16 @@ namespace ompl
             @par Associated publications:
 
             J.D. Gammell, S. S. Srinivasa, T.D. Barfoot, "BIT*: Batch Informed Trees."
-            In Proceedings of the Information-based Grasp and Manipulation Planning Workshop at Robotics: Science and Systems (RSS).
-            Berkeley, CA, USA, 13 July 2014.
+            In Proceedings of the Information-based Grasp and Manipulation Planning Workshop at
+            Robotics: Science and Systems (RSS). Berkeley, CA, USA, 13 July 2014.
             <a href="http://asrl.utias.utoronto.ca/~tdb/bib/gammell_rss14.pdf">Extended Abstract</a>.
             <a href="http://asrl.utias.utoronto.ca/~tdb/bib/gammell_rss14_poster.pdf">Poster</a>.
 
-            J D. Gammell, S. S. Srinivasa, T. D. Barfoot, "Batch Informed Trees (BIT*): Sampling-based Optimal Planning via the Heuristically Guided Search of Implicit Random Geometric Graphs,"
+            J D. Gammell, S. S. Srinivasa, T. D. Barfoot, "Batch Informed Trees (BIT*): Sampling-based
+            Optimal Planning via the Heuristically Guided Search of Implicit Random Geometric Graphs,"
             In Proceedings of the IEEE International Conference on Robotics and Automation (ICRA).
-            Seattle, Washington, USA, 26-30 May 2015.
-            <a href="http://arxiv.org/abs/1405.5848">arXiv:1405.5848 [cs.RO]</a>.
+            Seattle, WA, USA, 26-30 May 2015.
+            DOI: <a href="http://dx.doi.org/10.1109/ICRA.2015.7139620">10.1109/ICRA.2015.7139620</a>.
             <a href="http://www.youtube.com/watch?v=MRzSfLpNBmA">Illustration video</a>.
 
             \todo
@@ -132,6 +133,8 @@ namespace ompl
             /** \brief Get results */
             virtual void getPlannerData(base::PlannerData& data) const;
 
+            ///////////////////////////////////////
+            // Planner info for debugging, etc:
             /** \brief Get the next edge to be processed. Causes vertices in the queue to be expanded (if necessary) and therefore effects the run timings of the algorithm, but helpful for some videos and debugging. */
             std::pair<const ompl::base::State*, const ompl::base::State*> getNextEdgeInQueue();
 
@@ -144,11 +147,19 @@ namespace ompl
             /** \brief Get the whole set of vertices to be expanded. Expensive but helpful for some videos */
             void getVertexQueue(std::vector<VertexConstPtr>* verticesInQueue);
 
+            /** \brief Get the number of iterations completed */
+            unsigned int numIterations() const;
+
+            /** \brief Retrieve the best exact-solution cost found.*/
+            ompl::base::Cost bestCost() const;
+            ///////////////////////////////////////
+
+            ///////////////////////////////////////
+            // Planner settings:
             /** \brief Set a different nearest neighbours datastructure */
             template<template<typename T> class NN>
             void setNearestNeighbors();
-            ///////////////////////////////////////
-            // Planner settings:
+
             /** \brief Set the rewiring scale factor, s, such that r_rrg = s \times r_rrg* */
             void setRewireFactor(double rewireFactor);
 
@@ -384,9 +395,6 @@ namespace ompl
 
             ///////////////////////////////////////
             // Planner progress property functions
-            /** \brief Retrieve the best exact-solution cost found
-            as the raw data. (bestCost_) */
-            ompl::base::Cost bestCost() const;
             /** \brief Retrieve the best exact-solution cost found
             as a planner-progress property. (bestCost_) */
             std::string bestCostProgressProperty() const;
--- a/src/ompl/geometric/planners/cforest/src/CForest.cpp
+++ b/src/ompl/geometric/planners/cforest/src/CForest.cpp
@@ -45,10 +45,10 @@ ompl::geometric::CForest::CForest(const base::SpaceInformationPtr &si) : base::P
 
     numPathsShared_ = 0;
     numStatesShared_ = 0;
-    prune_ = true;
+    focusSearch_ = true;
 
     numThreads_ = std::max(boost::thread::hardware_concurrency(), 2u);
-    Planner::declareParam<bool>("prune", this, &CForest::setPrune, &CForest::getPrune, "0,1");
+    Planner::declareParam<bool>("focus_search", this, &CForest::setFocusSearch, &CForest::getFocusSearch, "0,1");
     Planner::declareParam<unsigned int>("num_threads", this, &CForest::setNumThreads, &CForest::getNumThreads, "0:64");
 
     addPlannerProgressProperty("best cost REAL",
@@ -75,8 +75,11 @@ void ompl::geometric::CForest::addPlannerInstanceInternal(const base::PlannerPtr
     else
     {
         planner->setProblemDefinition(pdef_);
-        if (planner->params().hasParam("prune"))
-            planner->params()["prune"] = prune_;
+        if (planner->params().hasParam("focus_search"))
+            planner->params()["focus_search"] = focusSearch_;
+        else
+            OMPL_WARN("%s does not appear to support search focusing.", planner->getName().c_str());
+
         planners_.push_back(planner);
     }
 }
--- a/src/ompl/geometric/planners/rrt/RRTstar.h
+++ b/src/ompl/geometric/planners/rrt/RRTstar.h
@@ -32,7 +32,7 @@
 *  POSSIBILITY OF SUCH DAMAGE.
 *********************************************************************/
 
-/* Authors: Alejandro Perez, Sertac Karaman, Ryan Luna, Luis G. Torres, Ioan Sucan, Javier V Gomez */
+/* Authors: Alejandro Perez, Sertac Karaman, Ryan Luna, Luis G. Torres, Ioan Sucan, Javier V Gomez, Jonathan Gammell */
 
 #ifndef OMPL_GEOMETRIC_PLANNERS_RRT_RRTSTAR_
 #define OMPL_GEOMETRIC_PLANNERS_RRT_RRTSTAR_
@@ -43,7 +43,10 @@
 
 #include <limits>
 #include <vector>
+#include <queue>
+#include <deque>
 #include <utility>
+#include <list>
 
 
 namespace ompl
@@ -87,6 +90,8 @@ namespace ompl
 
             virtual void clear();
 
+            virtual void setup();
+
             /** \brief Set the goal bias
 
                 In the process of randomly selecting states in
@@ -123,6 +128,19 @@ namespace ompl
                 return maxDistance_;
             }
 
+            /** \brief Set the rewiring scale factor, s, such that r_rrg = s \times r_rrg* (or k_rrg = s \times k_rrg*) */
+            void setRewireFactor(double rewireFactor)
+            {
+                rewireFactor_ = rewireFactor;
+                calculateRewiringLowerBounds();
+            }
+
+            /** \brief Set the rewiring scale factor, s, such that r_rrg = s \times r_rrg* > r_rrg* (or k_rrg = s \times k_rrg* > k_rrg*) */
+            double getRewireFactor() const
+            {
+                return rewireFactor_;
+            }
+
             /** \brief Set a different nearest neighbors datastructure */
             template<template<typename T> class NN>
             void setNearestNeighbors()
@@ -135,7 +153,7 @@ namespace ompl
                 planner then goes through this list, starting with the lowest
                 cost, checking for collisions in order to find a parent. The planner
                 stops iterating through the list when a collision free parent is found.
-                This prevents the planner from collsion checking each neighbor, reducing
+                This prevents the planner from collision checking each neighbor, reducing
                 computation time in scenarios where collision checking procedures are expensive.*/
             void setDelayCC(bool delayCC)
             {
@@ -148,42 +166,139 @@ namespace ompl
                 return delayCC_;
             }
 
-            /** \brief Controls whether the tree is pruned during the search. */
-            void setPrune(const bool prune)
-            {
-                prune_ = prune;
-            }
+            /** \brief Controls whether the tree is pruned during the search. This pruning removes
+                a vertex if and only if it \e and all its descendents passes the pruning condition.
+                The pruning condition is whether the lower-bounding estimate of a solution
+                constrained to pass the the \e vertex is greater than the current solution.
+                Considering the descendents of a vertex prevents removing a descendent
+                that may actually be capable of later providing a better solution once
+                its incoming path passes through a different vertex (e.g., a change in homotopy class). */
+            void setTreePruning(const bool prune);
 
             /** \brief Get the state of the pruning option. */
-            bool getPrune() const
+            bool getTreePruning() const
             {
-                return prune_;
+                return useTreePruning_;
             }
 
-            /** \brief Set the percentage threshold (between 0 and 1) for pruning the tree. If the new tree has removed
-                at least this percentage of states, the tree will be finally pruned. */
-            void setPruneStatesImprovementThreshold(const double pp)
+            /** \brief Set the fractional change in solution cost necessary for pruning to occur, i.e.,
+                prune if the new solution is at least X% better than the old solution.
+                (e.g., 0.0 will prune after every new solution, while 1.0 will never prune.) */
+            void setPruneThreshold(const double pp)
             {
-                pruneStatesThreshold_ = pp;
+                pruneThreshold_ = pp;
             }
 
             /** \brief Get the current prune states percentage threshold parameter. */
-            double getPruneStatesImprovementThreshold () const
+            double getPruneThreshold() const
             {
-                return pruneStatesThreshold_;
+                return pruneThreshold_;
             }
 
-            virtual void setup();
+            /** \brief Use the measure of the pruned subproblem instead of the measure of the entire problem domain (if such an expression exists and a solution is present).
+            Currently the only method to calculate this measure in closed-form is through a informed sampler, so this option also requires that. */
+            void setPrunedMeasure(bool informedMeasure);
 
-            ///////////////////////////////////////
-            // Planner progress property functions
-            std::string getIterationCount() const
+            /** \brief Get the state of using the pruned measure */
+            bool getPrunedMeasure() const
             {
-                return boost::lexical_cast<std::string>(iterations_);
+                return usePrunedMeasure_;
             }
-            std::string getBestCost() const
+
+            /** \brief Use direct sampling of the heuristic for the generation of random samples (e.g., x_rand).
+           If a direct sampling method is not defined for the objective, rejection sampling will be used by default. */
+            void setInformedSampling(bool informedSampling);
+
+            /** \brief Get the state direct heuristic sampling */
+            bool getInformedSampling() const
             {
-                return boost::lexical_cast<std::string>(bestCost_);
+                return useInformedSampling_;
+            }
+
+            /** \brief Controls whether heuristic rejection is used on samples (e.g., x_rand) */
+            void setSampleRejection(const bool reject);
+
+            /** \brief Get the state of the sample rejection option */
+            bool getSampleRejection() const
+            {
+                return useRejectionSampling_;
+            }
+
+            /** \brief Controls whether heuristic rejection is used on new states before connection (e.g., x_new = steer(x_nearest, x_rand)) */
+            void setNewStateRejection(const bool reject)
+            {
+                useNewStateRejection_ = reject;
+            }
+
+            /** \brief Get the state of the new-state rejection option */
+            bool getNewStateRejection() const
+            {
+                return useNewStateRejection_;
+            }
+
+            /** \brief Controls whether pruning and new-state rejection uses an admissible cost-to-come estimate or not */
+            void setAdmissibleCostToCome(const bool admissible)
+            {
+                useAdmissibleCostToCome_ = admissible;
+            }
+
+            /** \brief Get the admissibility of the pruning and new-state rejection heuristic */
+            bool getAdmissibleCostToCome() const
+            {
+                return useAdmissibleCostToCome_;
+            }
+
+            /** \brief A \e meta parameter to focusing the search to improving the current solution. This is the parameter set by CFOREST.
+            For RRT*, search focusing consists of pruning the existing search and limiting future search.
+            Specifically, this is accomplished by turning on informed sampling, tree pruning and new-state rejection.
+            This flag individually sets the options described above.
+            */
+            void setFocusSearch(const bool focus)
+            {
+                setInformedSampling(focus);
+                setTreePruning(focus);
+                setPrunedMeasure(focus);
+                setNewStateRejection(focus);
+            }
+
+            /** \brief Get the state of search focusing */
+            bool getFocusSearch() const
+            {
+                return getInformedSampling() && getPrunedMeasure() && getTreePruning() && getNewStateRejection();
+            }
+
+            /** \brief Use a k-nearest search for rewiring instead of a r-disc search. */
+            void setKNearest(bool useKNearest)
+            {
+                useKNearest_ = useKNearest;
+            }
+
+            /** \brief Get the state of using a k-nearest search for rewiring. */
+            bool getKNearest() const
+            {
+                return useKNearest_;
+            }
+
+            /** \brief Set the number of attempts to make while performing rejection or informed sampling */
+            void setNumSamplingAttempts(unsigned int numAttempts)
+            {
+                numSampleAttempts_ = numAttempts;
+            }
+
+            /** \brief Get the number of attempts to make while performing rejection or informed sampling */
+            unsigned int getNumSamplingAttempts() const
+            {
+                return numSampleAttempts_ ;
+            }
+
+            unsigned int numIterations() const
+            {
+                return iterations_;
+            }
+
+            ompl::base::Cost bestCost() const
+            {
+                return bestCost_;
             }
 
         protected:
@@ -219,6 +334,12 @@ namespace ompl
                 std::vector<Motion*> children;
             };
 
+            /** \brief Create the samplers */
+            void allocSampler();
+
+            /** \brief Generate a sample */
+            bool sampleUniform(base::State *statePtr);
+
             /** \brief Free the memory allocated by this planner */
             void freeMemory();
 
@@ -243,6 +364,9 @@ namespace ompl
                 return si_->distance(a->state, b->state);
             }
 
+            /** \brief Gets the neighbours of a given motion, using either k-nearest of radius as appropriate. */
+            void getNeighbors(Motion *motion, std::vector<Motion*> &nbh) const;
+
             /** \brief Removes the given motion from the parent's child list */
             void removeFromParent(Motion *m);
 
@@ -251,18 +375,29 @@ namespace ompl
 
             /** \brief Prunes all those states which estimated total cost is higher than pruneTreeCost.
                 Returns the number of motions pruned. Depends on the parameter set by setPruneStatesImprovementThreshold() */
-            int pruneTree(const base::Cost pruneTreeCost);
+            int pruneTree(const base::Cost& pruneTreeCost);
 
-            /** \brief Deletes (frees memory) the motion and its children motions. */
-            void deleteBranch(Motion *motion);
+            /** \brief Computes the solution cost heuristically as the cost to come from start to the motion plus
+                 the cost to go from the motion to the goal. If the parameter \e use_admissible_heuristic
+                 (\e setAdmissibleCostToCome()) is true, a heuristic estimate of the cost to come is used;
+                 otherwise, the current cost to come to the motion is used (which may overestimate the cost
+                 through the motion). */
+            base::Cost solutionHeuristic(const Motion *motion) const;
 
-            /** \brief Computes the Cost To Go heuristically as the cost to come from start to motion plus
-                 the cost to go from motion to goal. If \e shortest is true, the estimated cost to come
-                 start-motion is given. Otherwise, this cost to come is the current motion cost. */
-            base::Cost costToGo(const Motion *motion, const bool shortest = true) const;
+            /** \brief Add the children of a vertex to the given list. */
+            void addChildrenToList(std::queue<Motion*, std::deque<Motion*> > *motionList, Motion* motion);
+
+            /** \brief Check whether the given motion passes the specified cost threshold, meaning it will be \e kept during pruning */
+            bool keepCondition(const Motion* motion, const base::Cost& threshold) const;
+
+            /** \brief Calculate the k_RRG* and r_RRG* terms */
+            void calculateRewiringLowerBounds();
 
             /** \brief State sampler */
             base::StateSamplerPtr                          sampler_;
+
+            /** \brief An informed sampler */
+            base::InformedSamplerPtr                       infSampler_;
 
             /** \brief A nearest-neighbors datastructure containing the tree of motions */
             boost::shared_ptr< NearestNeighbors<Motion*> > nn_;
@@ -276,6 +411,17 @@ namespace ompl
             /** \brief The random number generator */
             RNG                                            rng_;
 
+            /** \brief Option to use k-nearest search for rewiring */
+            bool                                           useKNearest_;
+
+            /** \brief The rewiring factor, s, so that r_rrg = s \times r_rrg* > r_rrg* (or k_rrg = s \times k_rrg* > k_rrg*) */
+            double                                         rewireFactor_;
+
+            /** \brief A constant for k-nearest rewiring calculations */
+            double                                         k_rrg_;
+            /** \brief A constant for r-disc rewiring calculations */
+            double                                         r_rrg_;
+
             /** \brief Option to delay and reduce collision checking within iterations */
             bool                                           delayCC_;
 
@@ -288,23 +434,55 @@ namespace ompl
             /** \brief A list of states in the tree that satisfy the goal condition */
             std::vector<Motion*>                           goalMotions_;
 
-            /** \brief If this value is set to true, tree pruning will be enabled. */
-            bool                                           prune_;
+            /** \brief The status of the tree pruning option. */
+            bool                                           useTreePruning_;
 
-            /** \brief The tree is only pruned is the percentage of states to prune is above this threshold (between 0 and 1). */
-            double                                         pruneStatesThreshold_;
+            /** \brief The tree is pruned when the change in solution cost is greater than this fraction. */
+            double                                         pruneThreshold_;
 
-            struct PruneScratchSpace { std::vector<Motion*> newTree, toBePruned, candidates; } pruneScratchSpace_;
+            /** \brief Option to use the informed measure */
+            bool                                           usePrunedMeasure_;
 
-            /** \brief Stores the Motion containing the last added initial start state. */
-            Motion *                                       startMotion_;
+            /** \brief Option to use informed sampling */
+            bool                                           useInformedSampling_;
 
-            //////////////////////////////
-            // Planner progress properties
-            /** \brief Number of iterations the algorithm performed */
-            unsigned int                                   iterations_;
+            /** \brief The status of the sample rejection parameter. */
+            bool                                           useRejectionSampling_;
+
+            /** \brief The status of the new-state rejection parameter. */
+            bool                                           useNewStateRejection_;
+
+            /** \brief The admissibility of the new-state rejection heuristic. */
+            bool                                           useAdmissibleCostToCome_;
+
+            /** \brief The number of attempts to make at informed sampling */
+            unsigned int                                   numSampleAttempts_;
+
+            /** \brief Stores the start states as Motions. */
+            std::vector<Motion*>                           startMotions_;
+
             /** \brief Best cost found so far by algorithm */
             base::Cost                                     bestCost_;
+
+            /** \brief The cost at which the graph was last pruned */
+            base::Cost                                     prunedCost_;
+
+            /** \brief The measure of the problem when we pruned it (if this isn't in use, it will be set to si_->getSpaceMeasure())*/
+            double                                         prunedMeasure_;
+
+            /** \brief Number of iterations the algorithm performed */
+            unsigned int                                   iterations_;
+
+            ///////////////////////////////////////
+            // Planner progress property functions
+            std::string numIterationsProperty() const
+            {
+                return boost::lexical_cast<std::string>(numIterations());
+            }
+            std::string bestCostProperty() const
+            {
+                return boost::lexical_cast<std::string>(bestCost());
+            }
         };
     }
 }
--- a/src/ompl/geometric/planners/rrt/src/RRTstar.cpp
+++ b/src/ompl/geometric/planners/rrt/src/RRTstar.cpp
@@ -32,26 +32,45 @@
 *  POSSIBILITY OF SUCH DAMAGE.
 *********************************************************************/
 
-/* Authors: Alejandro Perez, Sertac Karaman, Ryan Luna, Luis G. Torres, Ioan Sucan, Javier V Gomez */
+/* Authors: Alejandro Perez, Sertac Karaman, Ryan Luna, Luis G. Torres, Ioan Sucan, Javier V Gomez, Jonathan Gammell */
 
 #include "ompl/geometric/planners/rrt/RRTstar.h"
 #include "ompl/base/goals/GoalSampleableRegion.h"
 #include "ompl/tools/config/SelfConfig.h"
 #include "ompl/base/objectives/PathLengthOptimizationObjective.h"
+#include "ompl/base/Goal.h"
+#include "ompl/base/goals/GoalState.h"
+#include "ompl/util/GeometricEquations.h"
+#include "ompl/base/samplers/InformedStateSampler.h"
+#include "ompl/base/samplers/informed/RejectionInfSampler.h"
 #include <algorithm>
 #include <limits>
 #include <boost/math/constants/constants.hpp>
+#include <boost/make_shared.hpp>
+#include <vector>
 
 ompl::geometric::RRTstar::RRTstar(const base::SpaceInformationPtr &si) :
     base::Planner(si, "RRTstar"),
     goalBias_(0.05),
     maxDistance_(0.0),
+    useKNearest_(true),
+    rewireFactor_(1.1),
+    k_rrg_(0u),
+    r_rrg_(0.0),
     delayCC_(true),
     lastGoalMotion_(NULL),
-    prune_(false),
-    pruneStatesThreshold_(0.95),
-    iterations_(0),
-    bestCost_(std::numeric_limits<double>::quiet_NaN())
+    useTreePruning_(false),
+    pruneThreshold_(0.05),
+    usePrunedMeasure_(false),
+    useInformedSampling_(false),
+    useRejectionSampling_(false),
+    useNewStateRejection_(false),
+    useAdmissibleCostToCome_(true),
+    numSampleAttempts_ (100u),
+    bestCost_(std::numeric_limits<double>::quiet_NaN()),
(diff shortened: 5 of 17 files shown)


src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -123,6 +123,18 @@ namespace ompl
...
+            {
+                rewireFactor_ = rewireFactor;
+            }
+
+            /** \brief Set the rewiring scale factor, s, such that r_rrg = s \times r_rrg* > r_rrg* (or k_rrg = s \times k_rrg* > k_rrg*) */
+            double getRewireFactor() const
+            {
+                return rewireFactor_;
+            }
+
```
gammell: Since the constants in RRT* are of the form A > B, where there's (only) a closed form for B, I added the ability to set the multiplicative factor to define A in terms of B.
Inspired by what the FMT* code does.

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -148,7 +160,12 @@ namespace ompl
                 return delayCC_;
             }
 
-            /** \brief Controls whether the tree is pruned during the search. */
+            /** \brief Controls whether the tree is pruned during the search. This pruning removes
+                a vertex \e and \e its \e descendants if the lower-bounding estimate of a solution
+                constrained to pass the the \e vertex  is greater than the current solution.
+                As the ancestory of the descendent vertices can change at anytime, this means that the removed
+                descendents may actually be capable of later providing a better solution once
+                their incoming path passes through a different vertex (e.g., a change in homotopy class) */
```
gammell: Expanded documentation on the limitations of the current pruning method.

I don't mean to say that there's anything wrong with this, I just wanted to make it clear where it can be overly aggressive. I guess I could make a safer pruning function (i.e., it only removes a vertex if all its descendents are to be removed too), but I don't know if that should replace this function or live in parallel.

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -173,17 +190,57 @@ namespace ompl
...
+            {
+                useKNearest_ = useKNearest;
+            }
+
+            /** \brief Get the state of using a k-nearest search for rewiring. */
+            bool getKNearest() const
+            {
+                return useKNearest_;
+            }
+
```
gammell: I expanded RRT* to have both an r-disc and k-nearest variant.

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -173,17 +190,57 @@ namespace ompl
...
-                return boost::lexical_cast<std::string>(iterations_);
+                return iterations_;
             }
-            std::string getBestCost() const
+
+            ompl::base::Cost bestCost() const
             {
-                return boost::lexical_cast<std::string>(bestCost_);
+                return bestCost_;
             }
```
gammell: I moved the planner progress properties to private functions, as per Ioan's request, but I did make the raw data available via `numIterations()` and `getBestCost()`, as I know I have some needs for it...

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -219,6 +276,12 @@ namespace ompl
                 std::vector<Motion*> children;
             };
 
+            /** \brief Create the samplers */
+            void allocSampler();
+
+            /** \brief Generate a sample */
+            void sampleUniform(base::State *statePtr);
+
```
gammell: Helper functions to simplify the creation/use of different samplers.

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -243,6 +306,9 @@ namespace ompl
                 return si_->distance(a->state, b->state);
             }
 
+            /** \brief Gets the neighbours of a given motion, using either k-nearest of radius as appropriate. */
+            void getNeighbors(Motion *motion, std::vector<Motion*> &nbh);
+
```
gammell: Helper function to simplify the choice of r-disc/k-nearest variants.

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -256,14 +322,26 @@ namespace ompl
...
             void deleteBranch(Motion *motion);
 
-            /** \brief Computes the Cost To Go heuristically as the cost to come from start to motion plus
-                 the cost to go from motion to goal. If \e shortest is true, the estimated cost to come
-                 start-motion is given. Otherwise, this cost to come is the current motion cost. */
-            base::Cost costToGo(const Motion *motion, const bool shortest = true) const;
+            /** \brief Computes the solution cost heuristically as the cost to come from start to the motion plus
+                 the cost to go from the motion to the goal. If \e estimate is true, a heuristic estimate of the
+                 cost to come is used; otherwise, the current cost to come is used. */
+            base::Cost solutionHeuristic(const Motion *motion, const bool estimate = true) const;
```
gammell: The previous `costToGo` function was poorly named. It calculated more than cost-to-go, but in fact an estimate of an entire solution path constrained to go through that vertex by adding a cost-to-go estimate to the cost-to-come (current or estimated). Renamed as `solutionHeuristic` and cleaned up the documentation.

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -256,14 +322,26 @@ namespace ompl
...
+
+            /** \brief Computes the cost-to-go heuristically for goal states and goal regions using the motionCost given by the optimization objective.*/
+            base::Cost defaultCostToGoHeuristic(const base::State *state, const base::Goal *goal) const;
+
```
gammell: Separated the cost-to-go component from the `solutionHeuristic` function.

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -256,14 +322,26 @@ namespace ompl
...
+            /** \brief Count the number of vertices that lay outside the informed subset and store the number in numVerticesWorseThanSoln_.
+            This is a sort of virtual pruning that avoids the problem of removing promising descendents simply
+            because of the quality of their ancestors.
+            \todo Replace with an actually pruning method. */
+            void countNumberOfVerticesWorseThanSoln();
```
gammell: If I were to rewrite the pruning function as described above, this function could be removed. It's like a sort of *virtual*-prune.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -162,12 +205,6 @@ ompl::base::PlannerStatus ompl::geometric::RRTstar::solve(const base::PlannerTer
 
     Motion *solution       = lastGoalMotion_;
 
-    // \todo Make this variable unnecessary, or at least have it
-    // persist across solve runs
-    base::Cost bestCost    = opt_->infiniteCost();
-
-    bestCost_ = opt_->infiniteCost();
-
```
gammell: OK.
I would really like someone to double check me here, as I have tried to clean up this weird `bestCost`/`bestCost_` thing. Here both variables were set to the same thing.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -413,20 +458,23 @@ ompl::base::PlannerStatus ompl::geometric::RRTstar::solve(const base::PlannerTer
                 bool updatedSolution = false;
                 for (size_t i = 0; i < goalMotions_.size(); ++i)
                 {
-                    if (opt_->isCostBetterThan(goalMotions_[i]->cost, bestCost))
+                    if (opt_->isCostBetterThan(goalMotions_[i]->cost, bestCost_))
                     {
-                        bestCost = goalMotions_[i]->cost;
-                        bestCost_ = bestCost;
```
gammell: Here both variables are also set to the same thing. These are the only two places that either variable are assigned, so therefore, we can just remove `bestCost`.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -632,14 +709,112 @@ void ompl::geometric::RRTstar::deleteBranch(Motion *motion)
...
-ompl::base::Cost ompl::geometric::RRTstar::costToGo(const Motion *motion, const bool shortest) const
+ompl::base::Cost ompl::geometric::RRTstar::defaultCostToGoHeuristic(const base::State *state, const base::Goal *goal) const
+{
+    if (pdef_->getGoal()->hasType(base::GOAL_STATE) == true)
+    {
+        return opt_->motionCost(state, goal->as<base::GoalState>()->getState());
+    }
+    else if (pdef_->getGoal()->hasType(base::GOAL_REGION) == true)
+    {
+        return base::goalRegionCostToGo(state, goal);
```
gammell: My understanding from the documentation is that this function only works for `GOAL_REGION` types? Hence the switch. I guess I could handle the GOAL_STATE**S** case by finding the min over the countable set. I will update that when I push changes back from your comments. :smile:

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
jvgomez: Last summer I was trying to remove this as well but at the end I didn't. In my case I was using both for the CForest framework, but now I agree that bestCost can be simply removed. It would be nice if you can run the CForest demo/tests with these changes to be sure that nothing is broken.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
gammell: So, I see now that `GoalState` and `GoalStates` are both derived from `GoalSampleableRegion`, which is itself derived from `GoalRegion`...

I think I need some guidance.

Does that mean `base::goalRegionCostToGo` **is** valid for goals of type `GoalState`? Does it just do the same as what I was doing with the switch? In the case of `GoalStates`, does it find the min over the states or does it assume that the cost-to-go calculated with any one is the same?... I will have to look into it more, but this could really clean up some of my other code. :smile:

gammell: Got this part sorted. Updates to the PR to follow shortly.

src/ompl/geometric/planners/cforest/src/CForest.cpp
```diff
@@ -75,8 +75,11 @@ void ompl::geometric::CForest::addPlannerInstanceInternal(const base::PlannerPtr
     else
     {
         planner->setProblemDefinition(pdef_);
-        if (planner->params().hasParam("prune"))
-            planner->params()["prune"] = prune_;
+        if (planner->params().hasParam("focus_search"))
+            planner->params()["focus_search"] = prune_;
+        else
+            OMPL_WARN("%s does not appear to support search focusing.", planner->getName().c_str());
+
```
gammell: More about this change explained in the RRTstar code, but my general thought here is to turn on a generic parameter called *focus_search*, instead of something as specifically named as *prune*.

This should make it conceptually cleaner for other planners to support the CFOREST framework even if they focus their search in ways other than pruning the tree.

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -148,42 +160,125 @@ namespace ompl
...
+                return useTreePruning_;
             }
 
-            /** \brief Set the percentage threshold (between 0 and 1) for pruning the tree. If the new tree has removed
-                at least this percentage of states, the tree will be finally pruned. */
-            void setPruneStatesImprovementThreshold(const double pp)
+            /** \brief Set the fractional change in solution cost necessary for pruning to occur, i.e.,
+                prune if the new solution is at least X% better than the old solution.
+                (e.g., 0.0 will prune after every new solution, while 1.0 will never prune.) */
+            void setPruneThreshold(const double pp)
```
gammell: Updated the pruning threshold to a calculation that does not require any searching of the tree to calculate. A trade off is that this threshold will always prune on your first solution as the change in cost is relative to an infinite cost...

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -148,42 +160,125 @@ namespace ompl
...
 
             /** \brief Get the current prune states percentage threshold parameter. */
-            double getPruneStatesImprovementThreshold () const
+            double getPruneThreshold() const
+            {
+                return pruneThreshold_;
+            }
+
+            /** \brief Controls whether heuristic rejection is used on samples (e.g., x_rand) */
+            void setSampleRejection(const bool reject);
```
gammell: Added the ability to do sample rejection. Previously this was **also** done if pruning was on.
I would argue that there is no need for this to exist in parallel with informed sampling, but I didn't want to be so presumptuous as to propose that. :smile:

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -148,42 +160,125 @@ namespace ompl
...
+
+            /** \brief Get the state of the sample rejection option */
+            bool getSampleRejection() const
+            {
+                return useRejectionSampling_;
+            }
+
+            /** \brief Controls where heuristic rejection is used on new states before connection (e.g., x_new = steer(x_nearest, x_rand)) */
+            void setNewStateRejection(const bool reject)
```
gammell: Added the ability to specifically control new-state rejection. Previously this was done automatically if pruning was on.

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -148,42 +160,125 @@ namespace ompl
...
+            bool getNewStateRejection() const
+            {
+                return useNewStateRejection_;
+            }
+
+            /** \brief Controls search focusing. Search focusing consists of pruning the existing search and limiting future search.
+            In RRT* this is accomplished by turning on Informed RRT* (tree pruning and focused sampling) and new-state rejection.
+            This flag individually sets the options described above.
+            */
+            void setFocusSearch(const bool focus)
```
gammell: This is the "meta" parameter to enable the generalized concept of focused search to facilitate use with CFOREST. Apparently I was less humble here and decided that the sample focusing would come from the informed sampling instead of rejection sampling.

src/ompl/geometric/planners/cforest/src/CForest.cpp
gammell: Maybe the parameter should even be called something like *enable_cforest_support* or something?

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -148,42 +160,126 @@ namespace ompl
...
-            /** \brief Controls whether the tree is pruned during the search. */
-            void setPrune(const bool prune)
+            /** \brief Controls whether the tree is pruned during the search. This pruning removes
+                a vertex if and only if it \e and all its descendents passes the pruning condition.
+                The pruning condition is whether the lower-bounding estimate of a solution
+                constrained to pass the the \e vertex is greater than the current solution.
+                Considering the descendents of a vertex prevents removing a descendent
+                that may actually be capable of later providing a better solution once
+                its incoming path passes through a different vertex (e.g., a change in homotopy class). */
+            void setTreePruning(const bool prune)
```
gammell: So I realised that half the point of this to be able to propose changes and easily compare them. I've rewritten the pruning function to be less aggressive. Open to discussion if that's what you want.

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -148,42 +160,126 @@ namespace ompl
...
+            IROS 2014. DOI: <a href="http://dx.doi.org/10.1109/IROS.2014.6942976">10.1109/IROS.2014.6942976</a>.
+            <a href="http://www.youtube.com/watch?v=d7dX5MvDYTc">Illustration video</a>.
+            <a href="http://www.youtube.com/watch?v=nsl-5MZfwu4">Short description video</a>. */
+            void setInformedRrtStar(bool informedRrtStar);
+
+            /** \brief Get the state of informed RRT* */
+            bool getInformedRrtStar() const
+            {
+                return useInformedSampling_ && getTreePruning();
+            }
```
gammell: Informed RRT\* is (technically) the shrinking of the problem domain during search, which requires a way to generate samples in the subdomain (informed sampling) and pruning the tree. As the tree pruning is now less aggressive, I can just use that in parallel and call the combination Informed RRT\*

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -251,19 +356,22 @@ namespace ompl
...
-            void deleteBranch(Motion *motion);
+            /** \brief Computes the solution cost heuristically as the cost to come from start to the motion plus
+                 the cost to go from the motion to the goal. If \e estimate is true, a heuristic estimate of the
+                 cost to come is used; otherwise, the current cost to come is used. */
+            base::Cost solutionHeuristic(const Motion *motion, const bool estimate = true) const;
 
-            /** \brief Computes the Cost To Go heuristically as the cost to come from start to motion plus
-                 the cost to go from motion to goal. If \e shortest is true, the estimated cost to come
-                 start-motion is given. Otherwise, this cost to come is the current motion cost. */
-            base::Cost costToGo(const Motion *motion, const bool shortest = true) const;
```
gammell: This is now unnecessary, see in the .cpp file.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -101,6 +129,16 @@ void ompl::geometric::RRTstar::setup()
         {
             OMPL_INFORM("%s: No optimization objective specified. Defaulting to optimizing path length for the allowed planning time.", getName().c_str());
             opt_.reset(new base::PathLengthOptimizationObjective(si_));
+
+            if (opt_->hasCostToGoHeuristic() == false)
+            {
+                OMPL_INFORM("%s: No cost-to-go heuristic set. Defaulting to the distance from the state to the goal.", getName().c_str());
+                //Bind to the goalRegionCostToGo function.
+                opt_->setCostToGoHeuristic( boost::bind(&base::goalRegionCostToGo, _1, _2) );
```
gammell: I looked into `base::goalRegionCostToGo` and now fully understand what it's doing. It was (obviously) the correct thing to be using, but it makes even more sense to only use it if a heuristic has not otherwise been specified in `opt_`.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -207,9 +261,8 @@ ompl::base::PlannerStatus ompl::geometric::RRTstar::solve(const base::PlannerTer
             goal_s->sampleGoal(rstate);
         else
         {
-            sampler_->sampleUniform(rstate);
-
-            if (prune_ && opt_->isCostBetterThan(bestCost_, costToGo(rmotion)))
+            // Attempt to generate a sample, if we fail (e.g., too many rejection attempts), skip the remainder of this loop and return to try again
+            if (!sampleUniform(rstate))
```
gammell: The rejection sampling that was being performed by the line `opt_->isCostBetterThan(bestCost_, costToGo(rmotion))` is now being done by the default rejection sampler that was recently introduced.

`sampleUniform` looks after calling the appropriate sampler and returns false if the sampler times out.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -338,16 +394,15 @@ ompl::base::PlannerStatus ompl::geometric::RRTstar::solve(const base::PlannerTer
                 }
             }
 
-            if (prune_)
+            if (useNewStateRejection_)
             {
-                if (opt_->isCostBetterThan(costToGo(motion, false), bestCost_))
+                if (opt_->isCostBetterThan(solutionHeuristic(motion, false), bestCost_))
```
gammell: In the literature, this isn't really the same thing as pruning. This is a rejection scheme that is performed on the output of the steer function. Made it a separate setting.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -338,16 +394,15 @@ ompl::base::PlannerStatus ompl::geometric::RRTstar::solve(const base::PlannerTer
...
                 {
                     nn_->add(motion);
                     motion->parent->children.push_back(motion);
                 }
                 else // If the new motion does not improve the best cost it is ignored.
                 {
-                    --statesGenerated;
```
gammell: You can't ungenerate a state. :wink:
It wasn't used, but all the work of generating it was done... Feel free to disagree.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +661,210 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
-        si_->freeState(mto_delete->state);
-        delete mto_delete;
+            // Create the sampler
+            allocSampler();
+        }
     }
 }
 
-ompl::base::Cost ompl::geometric::RRTstar::costToGo(const Motion *motion, const bool shortest) const
+void ompl::geometric::RRTstar::allocSampler()
```
gammell: Create the requested sampler, either informed (which can be direct), rejection, or regular.
Since we're doing rejection like this anyway, I think we could do just informed or regular, but as I said, I wasn't ready to go that far.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +661,210 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
     }
 }
 
-int ompl::geometric::RRTstar::pruneTree(const base::Cost pruneTreeCost)
+int ompl::geometric::RRTstar::pruneTree(const base::Cost& pruneTreeCost)
```
gammell: This is a more 'admissible' pruning scheme.

What it does is iterate through the graph and flag every "leaf vertex" that meets the pruning condition (i.e., cannot provide a better solution). It then deletes up the branches from these leaves for as long as the vertices are (a) cannot provide a better solution and (b) a leaf.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +661,210 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
+        // The easiest way to do this is to find all the leaves of the tree that meet the pruning criterion, and ascend up the tree from each leaf until a motion is found that should not be pruned
+
+        // Variable
+        // The list of motions in the NN structure, initially full of nulls
+        std::vector<Motion*> treeVertices(nn_->size(), NULL);
+        // The list of leaves to prune
+        std::list<Motion*> leavesToPrune;
+
+        // Get the list from the NN structure
+        nn_->list(treeVertices);
```
gammell: I didn't want to start from startMotion_, since if we had multiple start states, we could have grown multiple independent trees.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +661,210 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
+        for (std::vector<Motion*>::iterator vIter = treeVertices.begin(); vIter != treeVertices.end(); ++vIter)
+        {
+            // Check if it passes the pruning condition
+            if (leafPruningCondition(*vIter, pruneTreeCost) == true)
+            {
+                // Add to the list of leaves to prune
+                leavesToPrune.push_back(*vIter);
+            }
+            // No else, does not meet the pruning criteria
+        }
```
gammell: This finds all the initial leaves to trim.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +661,210 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
+
+            // then deleting the pointer
+            delete *lpIter;
+
+            // And finally removing it from the list, erase returns the next iterator
+            lpIter = leavesToPrune.erase(lpIter);
+
+            // Update our counter
+            ++numPruned;
+        }
```
gammell: This iterates through the list of leaves to trim, working from the front, we

1. Remove the leaf from the NN struct
2. Disconnecting it from its parent
3. Check if the parent should now be added to the list, (i.e., is it now a leaf? is it's heuristic estimate too high?) and adding if necessary.
4. Freeing the state and pointer
5. Removing from the list and getting the next iterator.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +661,210 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
-        for (std::size_t i = 0; i < pruneScratchSpace_.toBePruned.size(); ++i)
-            deleteBranch(pruneScratchSpace_.toBePruned[i]);
+    return numPruned;
+}
 
-        nn_->clear();
-        nn_->add(pruneScratchSpace_.newTree);
 
-        return (tree_size - pruneScratchSpace_.newTree.size());
+bool ompl::geometric::RRTstar::leafPruningCondition(Motion *motion, const base::Cost& pruneTreeCost)
```
gammell: The pruning condition in one place for ease of use on both the vertex and the vertex's parent.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +661,210 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
+        costToCome = opt_->motionCost(startMotion_->state, motion->state); // lower-bounding cost from the start to the state
+    else
+        costToCome = motion->cost; // current cost from the state to the goal
+
+    const base::Cost costToGo = opt_->costToGo(motion->state, pdef_->getGoal().get()); // lower-bounding cost from the state to the goal
+    return opt_->combineCosts(costToCome, costToGo); // add the two costs
 }
 
-void ompl::geometric::RRTstar::deleteBranch(Motion *motion)
+void ompl::geometric::RRTstar::setSampleRejection(const bool reject)
```
gammell: If we set sample rejection after a sampler has been allocated, we need to create a new sampler.
If a sampler has not yet been allocated, we don't need to.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +661,210 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
+            infSampler_.reset();
 
-    while (!toDelete.empty())
+            // Create the sampler
+            allocSampler();
+        }
+    }
+}
+
+void ompl::geometric::RRTstar::setInformedRrtStar(bool informedRrtStar)
```
gammell: Same as setRejectionSampler, allocate if necessary, defer otherwise.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +661,210 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
+        infSampler_ = boost::make_shared<base::RejectionInfSampler>(pdef_, numSampleAttempts_);
+    }
+    else
+    {
+        // We are using a regular sampler
+        sampler_ = si_->allocStateSampler();
+    }
+}
+
+bool ompl::geometric::RRTstar::sampleUniform(base::State *statePtr)
```
gammell: Use the appropriate sampler and return the result.

gammell:
So in taking a closer look at the pruning method and the base::goalRegionCostToGo, I took it upon myself to clean somethings up about `RRT*`. I hope I didn't step on anyone's toes in doing so, if I did I apologize.

Obviously I'm expecting some back and forth on whether they're the right choices. :wink:

isucan:
My usual request for rebasing :) I know this is a pain, but it is much easier to look at one commit rather than 40 to see what was done (when looking through history)

src/ompl/geometric/planners/rrt/RRTstar.h
isucan: This sounds good, but do you need this as a PlannerProgressProperty? Properties must return string, so maybe duplicate the function? Add a suffix "Property" to the function name, leave the private version as returning string and return the actual cost / nr iterations in public functions?

src/ompl/geometric/planners/rrt/RRTstar.h
isucan: can this be const?

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
isucan: That seems to be the case, yes. +1

src/ompl/geometric/planners/cforest/src/CForest.cpp
isucan: I think I like focus_search. We just need to make sure we use the name consistently. Please add a comment explaining what the param means.

src/ompl/geometric/planners/rrt/RRTstar.h
isucan: Can you expand on this a bit? Do you mean pruning should not be needed if informed sampling is available?

src/ompl/geometric/planners/rrt/RRTstar.h
isucan: I think it is easier for users if they can construct a planner type Foo, rather than RRTstar and then set a bunch of things to make it actually be InformedRRTStar. How about we add an InformedRRTStar, which is like PRMstar -- essentially just a new name + one or two settings in the base class (so all the code remains in RRTstar)

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
isucan: Should we not move this code as part of setup() for the OptimizationObjective? Or at least for the ProblemDefinition?

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
isucan: please check where this is used. Does this not report how many states were additionally stored in the tree? I suspect we should keep the --

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
isucan: I am inclined to agree. Can you see any disadvantages to this?

src/ompl/geometric/planners/bitstar/BITstar.h
```diff
@@ -397,9 +407,6 @@ namespace ompl
             ///////////////////////////////////////
             // Planner progress property functions
             /** \brief Retrieve the best exact-solution cost found
-            as the raw data. (bestCost_) */
```
isucan: dropSample() above: please pass const& instead of by value

src/ompl/base/src/Planner.cpp
```diff
@@ -134,6 +134,11 @@ ompl::base::PlannerStatus ompl::base::Planner::solve(double solveTime)
         return solve(timedPlannerTerminationCondition(solveTime, std::min(solveTime / 100.0, 0.1)));
 }
 
+ompl::base::PlannerStatus ompl::base::Planner::solve(time::duration solveTime)
+{
+    return solve(timedPlannerTerminationCondition(solveTime));
+}
```
mamoll: I don't understand why `solve(double)` and `solve(time::duration)` are both needed and do slightly different things. If they can really have the same behavior, then you can pass in `solve(time::seconds(solveTime.seconds))` to the existing method.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -512,20 +574,37 @@ ompl::base::PlannerStatus ompl::geometric::RRTstar::solve(const base::PlannerTer
...
+{
+    if (useKNearest_)
+    {
+        //- k-nearest RRT*
+        unsigned int k = std::ceil(k_rrg_ * log((double)(nn_->size() + 1u)));
+        nn_->nearestK(motion, k, nbh);
+    }
+    else
+    {
+        double r = std::min(maxDistance_, r_rrg_*std::pow(log((double)(nn_->size() + 1u))/((double)(nn_->size() + 1u)), 1/(double)(si_->getStateDimension())));
```
isucan: white space around operators? * /

src/ompl/base/src/PlannerTerminationCondition.cpp
```diff
@@ -222,7 +222,12 @@ ompl::base::PlannerTerminationCondition ompl::base::plannerAndTerminationConditi
 
 ompl::base::PlannerTerminationCondition ompl::base::timedPlannerTerminationCondition(double duration)
 {
-    return PlannerTerminationCondition(boost::bind(&timePassed, time::now() + time::seconds(duration)));
+    return timedPlannerTerminationCondition(time::seconds(duration));
+}
+
+ompl::base::PlannerTerminationCondition ompl::base::timedPlannerTerminationCondition(time::duration duration)
+{
+    return PlannerTerminationCondition(boost::bind(&timePassed, time::now() + duration));
```
mamoll: This is also a case where we have two methods that really do the same thing. Can we just pick one way and stick with that?

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +652,322 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
+{
+    double dimDbl = static_cast<double>(si_->getStateDimension());
+
+    // k_rrg > e+e/d.  K-nearest RRT*
+    k_rrg_ = rewireFactor_*(boost::math::constants::e<double>() + (boost::math::constants::e<double>() / dimDbl));
+
+    // r_rrg > 2*(1+1/d)^(1/d)*(measure/ballvolume)^(1/d)
+    if (useInformedSampling_ == true)
+    {
+        r_rrg_ = rewireFactor_*2.0*std::pow((1.0 + 1.0/dimDbl)*(prunedInfMeasure_/unitNBallMeasure(si_->getStateDimension())), 1.0/dimDbl);
```
isucan: white space around operators?

isucan:
Pending minor issues & rebase, this is ready for merge in my opinion! Thank you for your contribution!

src/ompl/geometric/planners/rrt/RRTstar.h
mamoll: Maybe this can be done by using the code in ompl/src/ompl/geometric/planners/prm/ConnectionStrategy.h?

src/ompl/geometric/planners/rrt/RRTstar.h
mamoll: I like Ioan's idea. It'll be easier for people to find InformedRRT* in the documentation.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
mamoll: I like Ioan's idea.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -230,6 +277,7 @@ ompl::base::PlannerStatus ompl::geometric::RRTstar::solve(const base::PlannerTer
         }
 
         // Check if the motion between the nearest state and the state to add is valid
+        ++collisionChecks_;
```
mamoll: Is this really the number of calls to the state validity checker? If not, does each collision check take the same amount of time? I'm curious what the intended use is of this variable.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +652,322 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
+        // if any of the the chain Motions are now leaves, and repeat that process until done.
+        // This avoids (1) copying the NN structure into an intermediate variable and (2) the use of the expensive NN::remove() method.
+
+        // Variable
+        // The queue of Motions to process:
+        std::list<Motion*> motionQueue;
+        // The list of leaves to prune
+        std::list<Motion*> leavesToPrune;
+        // The list of chain vertices to recheck after pruning
+        std::list<Motion*> chainsToRecheck;
```
mamoll: Did you benchmark this pruning scheme against the current scheme? In general, std::list is slow as you are allocating elements one at a time. In the current scheme we use std::vector::reserve to do just one alloc. Also, better memory coherence with std::vector. I think the Scott Meyer books on effective C++ have some interesting info on this.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +652,322 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
+
+        //Clear the NN structure:
+        nn_->clear();
```
mamoll: It could be expensive to rebuild the entire nn structure each time you call pruneTree. Each insertion is probably O(d*log(n)), where d is the number of dimensions of the c-space and n the number of elements in the nn structure. Of course this depends on the nn structure (it could be even worse) and the distance function.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +652,322 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
+        // And if we're using informed sampling, the measure to which we've pruned
+        if (useInformedSampling_)
+        {
+            prunedInfMeasure_ = infSampler_->getInformedMeasure(prunedCost_);
         }
     }
 
-    // To create the new nn takes one order of magnitude in time more than just checking how many
-    // states would be pruned. Therefore, only prune if it removes a significant amount of states.
-    if ((double)pruneScratchSpace_.newTree.size() / tree_size < pruneStatesThreshold_)
```
mamoll: I seemed to remember that this was also important. @jvgomez?

src/ompl/geometric/planners/rrt/RRTstar.h
gammell: Yup, already done, though I'll clean up the names as requested.

src/ompl/geometric/planners/rrt/RRTstar.h
gammell: Good eye. :smile:

src/ompl/geometric/planners/rrt/RRTstar.h
gammell: Sorry, that statement was a bit vague. I meant there's no (long-term) need to support both rejection sampling and informed sampling of an admissible heuristic, as

1. Informed sampling is better than rejection sampling (working on a paper to show this), and
2. The way we wrote the classes, an objective without informed sampling just defaults to rejection sampling.

For now, I do think this should remain an option though in case someone wants to reject on a inadmissible heuristic.

src/ompl/geometric/planners/rrt/RRTstar.h
gammell: To my untrained eye, `prm/ConnectionStrategy.h` appears fairly tailored to PRM? There is also the problem that for the r-disc variants, the RRT\* bound is different than the PRM\* bound...

src/ompl/geometric/planners/rrt/RRTstar.h
gammell: This is up to you.
I personally just think of Informed RRT\* as a demonstration of the informed sampling (and more generally the closed-form for path-length's informed subset), but I can easily make a meta-class version.

src/ompl/geometric/planners/rrt/RRTstar.h
mamoll: Ok, just leave as-is. I thought there might have been an opportunity to unify some code.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
gammell: This variable is only used in the `OMPL_INFORM` line at the end of solve as `[...] Created %u new states [...]`. 

I guess what is the interpretation of the phrase creating a new state?

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
gammell: I'm not sure I fully understood this comment.
`goalRegionCostToGo` assumes that the cost-to-go of a state is equal to the distance the state is from the goal region, which means it's only applicable to `PathLengthOptimizationObjective`.

I've moved this call to that constructor, is that what was desired?

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
gammell: I don't think so?...

src/ompl/geometric/planners/bitstar/BITstar.h
gammell: I'm going to defer.
I can tell you a fun story sometime about a day spent learning the hard way that it is safer for delete functions for a shared pointer to be called by value. :wink:

Essentially, a delete function that does nontrivial cleanup (i.e., both accessing the pointer and removing it from owning containers) to delete the pointer before it's done with it.

I'm not sure if it's this exact function that requires the pointer by value, but one of the `BIT*` functions definitely does, so I used the same pattern for all of them to be safe.

src/ompl/base/src/Planner.cpp
gammell: I dunno, I guess I just like syntactic-sugar... Undone.

src/ompl/base/src/PlannerTerminationCondition.cpp
gammell: Well here's a case where I thought they fit nicely on top of each other.
The existing `PTC(double)` call converts the value into a `duration` internally, so all I was doing was breaking out that intermediate step. Then if one was using `duration`, they didn't have to convert to a `double` just to have the `PTC` convert back to a `duration`... That seemed likely to lose timing resolution.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
gammell: It is the number of calls to `SpaceInformation::checkMotion`.
The intent was as a planner progress property, as it can be interesting to see how many collision checks a planner is making. I thought it was useful, if your opinion differs, that's fine just let me know. :smile:

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
gammell: @jvgomez, please correct me if I'm wrong, but my understanding is this is a condition to decide whether it is worth the effort to prune or not. It is calculated from the number of vertices that will be pruned, and as a result requires evaluating the entire tree to make a prune decision on each vertex.

The proposed change replaces this condition with one that is based on the change in cost since the last time the graph was pruned.

The advantage to a cost-based method is that it doesn't require any calculation (just a comparison of `Cost`) before deciding to prune. The disadvantage is that it doesn't know how many vertices will actually be pruned from the tree.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
gammell: I kept the pattern of rebuilding the NN structure from scratch as that was was the existing prune method. The alternative would be to call `remove` on each vertex to be removed...

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
mamoll: `SpaceInformation::checkMotion` can correspond to several collision checks. If you want the number of collision checks, you'd need to count the number of calls to the various StateValidityChecker::isValid(...) methods. It's easy enough to keep track of this with an internal counter in the StateValidityChecker, but I'm not sure we should make this a requirement for a StateValidityChecker-derived class. But suppose you do this, then you can update ompl::base::Planner::getPlannerData() to add the number of collision checks as an extra property.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
gammell: No, I didn't benchmark this against the previous scheme as the previous pruning method was inadmissible. By removing a vertex independent of the heuristic-value of its children, the pruning method does not maintain uniform density in the subproblem, and more importantly can remove vertices that could be part of a better solution (e.g., in the attached picture, the highlighted vertex and its descendents would be removed by the old method, even though some of the descendents will be part of a better solution once a passage is found through the gap).

output


Regardless, you are correct that in the previous method that `std::vector` was allocated in a single call, but the difference is that it did so twice, both of the size of the NN structure (it made `std::vector`s for the vertices to keep and prune, respectively).

Alternatively, this method does not store the vertices to keep in a `std::vector`, but instead immediately adds them back to the NN structure. So no allocation for those vertices ever occurs. The other method didn't do this because it was actually counting the vertices to decide whether to go through with the prune or not (see other comment).

The only vertices it does store are ones that it cannot immediately make a decision on (`chainsToRecheck`) and those that are to be pruned (`leavesToPrune`). For these vertices, it's just using the `std::list` as a queue...

I guess the underlying assumption is that you will be pruning (significantly) less vertices than the size of the graph. Would preallocating to the entire size of the tree still be beneficial? I don't really know.

gammell:
Just about to push changes as per comments.

Is it okay if I delay rebasing until the PR is ready to be merged? I believe there are still some outstanding items to discuss.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
gammell: So's that a remove it since the information isn't meaningful?

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
gammell: In that you want them, correct?

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -573,73 +645,400 @@ void ompl::geometric::RRTstar::getPlannerData(base::PlannerData &data) const
...
+        if (useKNearest_ == false)
+        {
+            calculateRewiringLowerBounds();
         }
     }
+}
 
-    // To create the new nn takes one order of magnitude in time more than just checking how many
-    // states would be pruned. Therefore, only prune if it removes a significant amount of states.
-    if ((double)pruneScratchSpace_.newTree.size() / tree_size < pruneStatesThreshold_)
```
gammell: GitHub has changed the diff it associates for this line, unfortunately burying the previous conversation about it. This is a comment to draw attention to that.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
mamoll: I don't think it is meaningful. Worse, people might think it literally means number of collision checks, which it doesn't. In other words, I think you should remove it.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
mamoll: Ok, let's leave things as you have them for now. Try running ompl_benchmark on some simple problems with the old RRT* and new RRT* code (both with pruning on) to make sure nothing broke and, ideally, that the new version is faster.

As a general rule of thumb, it's almost always better to use std::vector over std::list *unless* you frequently need to remove elements. Instead of mentioning "memory coherence" I should said that std::vector's are more cache friendly. It may not matter much if you still have to deference Motion*'s, so I am willing to leave it the way you have it, provided it doesn't make RRT* significantly slower.

isucan: -1 for std::list. It is very rarely the case that a list is faster, even with remove operations. I am also ok to leave this is in if benchmarking shows it makes no difference.

gammell:
Ok,
Resurrecting this PR after my absence.
I've changed the `std::list` structures that were of concern to `std::queue`-wrapped `std::deque`s. From what I [understand](http://stackoverflow.com/questions/1262808/which-stl-container-should-i-use-for-a-fifo), the underlying `deque` has the cache performance that you desired, and for me it keeps the intent of the code obvious.

I wrote some code that ran normal RRT* to get a good, dense graph and then called the prune algorithm. Testing `std::queue`s with both `std::list` and `std::deque` gave average prune times of `1.0568e-3`sec and `7.965e-4`sec, respectively (on an old computer). So it appears that your intuition for `std::list` was correct.

Let me know if this addresses your concerns or if I misunderstood them. I'll get to work on fixing that unit test issue now.

mamoll:
I have checked the last couple commits and they look good to me. It's hard to navigate through all the comments on different versions of this PR, but it seems like we are now ready to merge, right?

gammell:
Yeah, is there anything I can do to make it easier to navigate through all the different versions? I've rebased it.

There's one PR I'd like to submit after this is merged, it updates Informed Sampling and BIT* to handle move-it and multiple starts/goals.

mamoll:
Once you resolve the conflicts, I'll merge this one.

gammell:
Huh, those are new...
I was just about to start packing up for the day, but I will sort this out first. Give me a second.

gammell:
I think I got it, but give me a second to the merge a once-over.

gammell:
Yup, I'm pretty sure that's it. 
The conflict was just the fix I made to the intermediateSolutionCallback in the first PR, I should have rebased this branch onto that one yesterday in preparation, sorry.

I'll get my last PR up this weekend, but right now I've got to run.

PR #58, opened 2015-07-28
Fixed setup segfault when FMT is used with MoveIt!

sonny-tarbouriech:


jvgomez:
It seems that your editor saved it as tabs instead of spaces. OMPL convention is to use 4 spaces for indenation.

sonny-tarbouriech:
Oups sorry, I rectified this.

mamoll:
Looks good. Thanks!

PR #59, opened 2015-08-01
Margin free cost comparisons

gammell:
These are the changes for option (1) as in Issue [#156](https://bitbucket.org/ompl/ompl/issues/156), moving the margin into only `isSatisfied`. I made the change faithful to the previous behaviour, though I am somewhat surprised to notice that the margin actually makes it harder to satisfy the specified threshold.

### DIFF
--- a/src/ompl/base/objectives/src/MaximizeMinClearanceObjective.cpp
+++ b/src/ompl/base/objectives/src/MaximizeMinClearanceObjective.cpp
@@ -52,7 +52,7 @@ ompl::base::Cost ompl::base::MaximizeMinClearanceObjective::stateCost(const Stat
 
 bool ompl::base::MaximizeMinClearanceObjective::isCostBetterThan(Cost c1, Cost c2) const
 {
-    return c1.value() > c2.value() + magic::BETTER_PATH_COST_MARGIN;
+    return c1.value() > c2.value();
 }
 
 ompl::base::Cost ompl::base::MaximizeMinClearanceObjective::identityCost() const
--- a/src/ompl/base/OptimizationObjective.h
+++ b/src/ompl/base/OptimizationObjective.h
@@ -80,7 +80,7 @@ namespace ompl
             /** \brief Get the description of this optimization objective */
             const std::string& getDescription() const;
 
-            /** \brief Verify that our objective is satisfied already and we can stop planning */
+            /** \brief Check if the the given cost \e c satisfies the specified cost objective, meaning we may stop planning. */
             virtual bool isSatisfied(Cost c) const;
 
             /** \brief Returns the cost threshold currently being checked for objective satisfaction */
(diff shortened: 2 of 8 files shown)


mamoll:
:+1:

gammell:
Now adding the margin by value. I was incorrect to use `combineCosts`, as there are objectives where it is not *combinatorial*.
e.g., 
```
ompl::base::Cost ompl::base::MinimaxObjective::combineCosts(Cost c1, Cost c2) const
{
    if (this->isCostBetterThan(c1, c2))
        return c2;
    else
        return c1;
}
```

src/ompl/base/objectives/src/MaximizeMinClearanceObjective.cpp
```diff
@@ -50,9 +50,14 @@ ompl::base::Cost ompl::base::MaximizeMinClearanceObjective::stateCost(const Stat
     return Cost(si_->getStateValidityChecker()->clearance(s));
 }
 
+bool ompl::base::MaximizeMinClearanceObjective::isSatisfied(Cost c) const
+{
+    return isCostBetterThan(c, Cost(threshold_.value() + magic::BETTER_PATH_COST_MARGIN));
```
isucan: should this be - ?

isucan:
I will review this shortly. Can we collapse the commits into one?

src/ompl/base/objectives/src/MaximizeMinClearanceObjective.cpp
gammell: This is consistent to what it currently is.

In both cases (`MaximizeMinClearanceObjective` and the default `OptimizationObjective`) the effect of the MARGIN is to make the threshold tougher to satisfy. Whether that's how it *should* be depends on what the original motivation was...

gammell:
Um, like rebasing? The last time I tried to do that I made a mess of all my branches and had to recreate them... :worried:

isucan:
If possible to rebase, yes, please try again. When we will look through history, it will be much more clear what was changed if the changes are collapsed into one commit.

src/ompl/base/objectives/src/MaximizeMinClearanceObjective.cpp
isucan: I agree. Looking through this, I am thinking we should just get rid of this margin completely. Thoughts?

isucan:
Mark, if you also agree, I think we should just remove the margin completely. +1 for merging;

mamoll:
I am also in favor of removing the margin and then merging.

gammell:
Ok, sounds great. I'll get you something in the next couple days on this and the other PR.

I'll get rebasing figured out, my main problem with it had been commits getting duplicated between branches, but I guess that won't be a problem if the goal is to collapse and merge. :smile:

gammell:
Ok, all rebased (this one was always going to be easy :wink:). I've completely removed `magic::BETTER_PATH_COST_MARGIN`, and everything compiles.

This is great by the way, I definitely margin-free is best for the default/low-level comparison. I just didn't know if the change would be possible given the code already written.

PR #60, opened 2015-09-07
update homebrew for travis CI

jf---:
Travis' homebrew uses boost 1.56, though homebrew users will be generally on boost 1.58
By updating brew, this will more accurately reflect the average homebrew install

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,9 @@ install:
   - if [ "$TRAVIS_OS_NAME" = "osx"   ]; then brew install boost-python ode eigen; fi
 
 script:
+  # update homebrew
+  - brew update
+
   # Create build directory
   - mkdir build
   - cd build


jf---:
awkward, still compiling with boost 1.56 after `brew update`

mamoll:
Wouldn't the build bots periodically update dependencies? I don't run `apt-get update && apt-upgrade` on the Linux build bot. Also, I don't think OMPL actually gets built on OS X. There is no OS X build log and I think there is a waiting list for open source projects.

jf---:
@mamoll , well, to me this seems like good news, but for sure osx is building, [see travis log here](https://travis-ci.org/ompl/ompl/jobs/79107400). The build is succesful, :cake: though some ( minor ) issues in the test, which is why the osx build fails. 

Still, I think that's pretty awesome though :sunglasses:

mamoll:
That's pretty cool! I think there was a wait list for OS X build bots for open source projects. If I understand you correctly, this pull request can be closed, right, since it doesn't have the desired effect?

jf---:
@mamoll, alas, correct... 
thing is that I can only invoke Travis build through this way, that is to say: sorry for your troubles :sunglasses:

PR #62, opened 2015-10-26
Informed sampling multistart/goal support, BIT* MoveIt! and infinite problem support

gammell:
This is the last PR I had in my queue. It's a couple things.

1. Support for multiple starts/goals in `PathLengthDirectInfSampler`,
2. Support for the MoveIt! workflow in `BIT*`,
3. Support for planning problems with infinite bounds in `BIT*`,
4. A couple other new ideas for `BIT*`, and
5. Documentation updates for `PathLengthDirectInfSampler,` `InformedRRT*` and `BIT*` to support the upcoming release.

Will rebase when it's ready to go.

### DIFF
--- a/src/ompl/base/goals/src/GoalStates.cpp
+++ b/src/ompl/base/goals/src/GoalStates.cpp
@@ -88,6 +88,7 @@ void ompl::base::GoalStates::sampleGoal(base::State *st) const
     samplePosition_ = samplePosition_ % states_.size();
     // Get the next state.
     si_->copyState(st, states_[samplePosition_]);
+    // Increment the counter. Do NOT roll over incase a new state is added before sampleGoal is called again.
     samplePosition_++;
 }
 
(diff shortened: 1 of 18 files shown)


mamoll:
Overall, looks good. Looks most (all?) changes only affect BIT*. Is that right? I am not a fan of std::list, and would use std::vector for the start and goal states, but performance is likely not a huge deal here.

gammell:
All the changes are to informed code, specifically either to `PathLengthDirectInfSampler` or `BIT*`.

All the other changes in files are just little documentation/consistency things, like making sure that both the direct and rejection-based informed samplers use the `virtual` keyword in the functions they overload.

Should I rebase?

mamoll:
Yes, please (if it's not too much trouble).

gammell:
Not at all, I'm getting better at it through trial and error. :wink:

gammell:
Mark,
Thanks again for all the guidance and help with integrating my code.

I actually just realised I have a small bug fix for the infinite-problem case. I'll create a PR for that once I finish testing it in an hour or so, but don't worry, this isn't the start of a thousand little PRs. :smile:

PR #63, opened 2015-10-27
BIT*: Infinite problem domain fix

gammell:
A much better default-way for `BIT*` to handle the "size" of the first batch of an infinite planning problem. Instead of assuming unit measure, estimates it as that of a (hyper)cube proportional to the distance between start and goal. Provides some relation to the order-of-magnitude of the problem, and moves the `std::isfinite` check on the measure to just `setup`.


gammell:
Again, sorry, I should have caught this before you merged the PR last night.

PR #64, opened 2015-10-28
BIT* problem measure fix and default switch to k-nearest.

gammell:


gammell:
I think this addresses the issues, let me know if it does not or if you have any further concerns.

mamoll:
:+1: looks good to me.

PR #66, opened 2015-12-31
BIT*: Correct k-nearest version

gammell:
The existing k-nearest version of BIT* is only an approximation. This is a version that is correct. Also cleaned up some other things while I was in there.

Happy new year,
Jon


gammell:
There appears to be a regression performance bug. I'm not sure if it predates this PR (i.e., whether it's already in the master or not), but I may as well get it sorted before you merge this.

PR #67, opened 2016-01-10
BIT*: Correct k-nearest version

gammell:
It turns out that the regression bug was a weird and subtle timing-related thing in my private experimental code. I rebased a commit to update a description in the header file though, and it won't let me reopen the previous commit.

The description was:
> The existing k-nearest version of BIT* is only an approximation. This is a version that is correct. Also cleaned up some other things while I was in there.

This is only changes to BIT* and a cosmetic change to Informed RRT*, so hopefully it'll be easy to merge?


mamoll:
Looks good to me.

PR #68, opened 2016-03-08
Fixed missing const of operator() in two comparator structs, ...

ralph-lange:
… which caused errors when compiling with Visual Studio 2015.


mamoll:
Looks good, thanks!

PR #69, opened 2016-03-25
Extended FMT: add resampling strategy if FMT is not able to find a solution.

jvgomez:
Resampling strategy for FMT.  Work included in the paper of the [Bidirectional FMT paper](http://arxiv.org/pdf/1507.07602.pdf(http://arxiv.org/pdf/1507.07602.pdf) (that planner will be included in another PR). This enormously increases the success rate of the planner with very little impact in the computation time.

Other minor modifications (simplifications) have been included.

### DIFF
--- a/src/ompl/geometric/planners/fmt/src/FMT.cpp
+++ b/src/ompl/geometric/planners/fmt/src/FMT.cpp
@@ -586,8 +586,8 @@ bool ompl::geometric::FMT::expandTreeFromNode(Motion **z)
 
                 // Add x to Open
                 Open_new.push_back(x);
-                // Remove x from Unvisited and add to Open
-                x->setSetType(Motion::SET_OPEN);
+                // Remove x from Unvisited
+                x->setSetType(Motion::SET_CLOSED);
             }
         } // An optimal connection from Open to x was found
     } // For each node near z and in set Unvisited, try to connect it to set Open
--- a/src/ompl/geometric/planners/fmt/src/FMT.cpp
+++ b/src/ompl/geometric/planners/fmt/src/FMT.cpp
@@ -532,6 +532,7 @@ bool ompl::geometric::FMT::expandTreeFromNode(Motion **z)
 
     // For each node near z and in set Unvisited, attempt to connect it to set Open
     std::vector<Motion*> yNear;
+    std::vector<Motion*> Open_new;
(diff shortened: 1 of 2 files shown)


mamoll:
:+1: Looks good to me. I'll give others a chance to comment before merging.

jvgomez:
@schmrlng FYI

src/ompl/geometric/planners/fmt/src/FMT.cpp
```diff
@@ -478,10 +581,12 @@ bool ompl::geometric::FMT::expandTreeFromNode(Motion **z)
                 x->setParent(yMin);
                 x->setCost(cMin);
                 x->setHeuristicCost(opt_->motionCostHeuristic(x->getState(), goalState_));
+                yMin->getChildren().push_back(x);
+
                 // Add x to Open
-                Open_new.push_back(x);
-                // Remove x from Unvisited
-                x->setSetType(Motion::SET_CLOSED);
+                Open_.insert(x);
```
schmrlng: I remember we discussed this a long time ago as a negligible difference from the paper version of FMT (see http://arxiv.org/pdf/1306.3532.pdf, footnote 2, page 8). I can't remember - was the timing difference negligible as well?

schmrlng:
Looks good to me.

src/ompl/geometric/planners/fmt/src/FMT.cpp
jvgomez: Yes, I remember that as well. But I couldn't remember what we chose. There is no real difference in time (actually I still think that it does not make any difference in the theory as well hehe). But I can leave it as it was, to avoid unnecessary changes. Another plus is that this way is closer to the paper. However in the [BFMT PR ](https://github.com/ompl/ompl/pull/70) we still use the `Open_new` set so for consistency I think  I will remove these changes.

schmrlng: Cool. But yeah, I agree with you that it really doesn't matter either way.

jvgomez: I see that in fact the footnote changed. Speed is negligible because `Open_new` is relatively small.

jvgomez:
Changes reverted. Ready again from my point of view.

PR #70, opened 2016-03-25
Add BFMT planner

jvgomez:
I acknowledge the code can be done much more elegantly. However I didn't have the time for this and it has been waiting for more than a year.

I have many improvements in mind and hopefully I will start doing them in the following days, such as:

- Code refactorization.
- A real precomputation strategy.
- Option of running both trees in separate threads.

I could not test the python bindings because in Ubuntu 14.04 is being a nightmare to update castxml and that stuff, sorry. I have run the planner several hundred times and never crashes. If you could perform a benchmark to check that results are like supposed to be that would be awesome :)


jvgomez:
@schmrlng FYI.

mamoll:
I will look at the code later, but installing Py++, pygccxml, and castxml shouldn't be that bad. See the [recently updated Py++ installation instructions](https://github.com/ompl/ompl/blob/master/doc/markdown/installPyPlusPlus.md#installation-of-py). You can download pre-compiled binaries of castxml, so that you don't have to download the whole Clang toolchain.

jvgomez:
I will try next week. Castxml is OK but pyplusplus was the problem, for some reason the solution that Dave reported didn't work for me but I didn't consider that important.

mamoll:
Code looks good to me. :+1:  I'll give @isucan a chance to give his opinion, but if he doesn't respond within a couple days, I'll merge.

PR #84, opened 2016-08-11
RRT*-Smart Implementation

fahadislam:
Added the functionality of RRT∗-Smart algorithm as an option on top of RRT∗. 
Related publications:

Nasir, J., Islam, F., Malik, U., Ayaz, Y., Hasan, O., Khan, M., & Muhammad, M. S. (2013). RRT*-SMART: A rapid convergence implementation of RRT. International Journal of Advanced Robotic Systems, 10.
Islam, F., Nasir, J., Malik, U., Ayaz, Y., & Hasan, O. (2012, August). RRT∗-smart: Rapid convergence implementation of RRT∗ towards optimal solution. In 2012 IEEE International Conference on Mechatronics and Automation (pp. 1651-1656). IEEE.

### DIFF
--- a/demos/OptimalPlanning.cpp
+++ b/demos/OptimalPlanning.cpp
@@ -48,6 +48,7 @@
 #include <ompl/geometric/planners/fmt/FMT.h>
 #include <ompl/geometric/planners/fmt/BFMT.h>
 #include <ompl/geometric/planners/rrt/InformedRRTstar.h>
+#include <ompl/geometric/planners/rrt/RRTstarsmart.h>
 #include <ompl/geometric/planners/prm/PRMstar.h>
 #include <ompl/geometric/planners/rrt/RRTstar.h>
 
@@ -74,6 +75,7 @@ enum optimalPlanner
     PLANNER_FMTSTAR,
     PLANNER_BFMTSTAR,
     PLANNER_INF_RRTSTAR,
+    PLANNER_RRTSTARSMART,
     PLANNER_PRMSTAR,
     PLANNER_RRTSTAR
 };
@@ -165,6 +167,11 @@ ob::PlannerPtr allocatePlanner(ob::SpaceInformationPtr si, optimalPlanner planne
         case PLANNER_INF_RRTSTAR:
         {
             return std::make_shared<og::InformedRRTstar>(si);
+            break;
+        }
+        case PLANNER_RRTSTARSMART:
+        {
+            return std::make_shared<og::RRTstarsmart>(si);
             break;
         }
         case PLANNER_PRMSTAR:
@@ -419,7 +426,7 @@ bool argParse(int argc, char** argv, double* runTimePtr, optimalPlanner *planner
     desc.add_options()
         ("help,h", "produce help message")
         ("runtime,t", bpo::value<double>()->default_value(1.0), "(Optional) Specify the runtime in seconds. Defaults to 1 and must be greater than 0.")
-        ("planner,p", bpo::value<std::string>()->default_value("RRTstar"), "(Optional) Specify the optimal planner to use, defaults to RRTstar if not given. Valid options are BITstar, CForest, FMTstar, BFMTstar, InformedRRTstar, PRMstar, and RRTstar.") //Alphabetical order
+        ("planner,p", bpo::value<std::string>()->default_value("RRTstar"), "(Optional) Specify the optimal planner to use, defaults to RRTstar if not given. Valid options are BITstar, CForest, FMTstar, BFMTstar, InformedRRTstar, RRTstarsmart, PRMstar, and RRTstar.") //Alphabetical order
         ("objective,o", bpo::value<std::string>()->default_value("PathLength"), "(Optional) Specify the optimization objective, defaults to PathLength if not given. Valid options are PathClearance, PathLength, ThresholdPathLength, and WeightedLengthAndClearanceCombo.") //Alphabetical order
         ("file,f", bpo::value<std::string>()->default_value(""), "(Optional) Specify an output path for the found solution path.")
         ("info,i", bpo::value<unsigned int>()->default_value(0u), "(Optional) Set the OMPL log level. 0 for WARN, 1 for INFO, 2 for DEBUG. Defaults to WARN.");
@@ -489,6 +496,10 @@ bool argParse(int argc, char** argv, double* runTimePtr, optimalPlanner *planner
     else if (boost::iequals("InformedRRTstar", plannerStr))
     {
         *plannerPtr = PLANNER_INF_RRTSTAR;
+    }
+    else if (boost::iequals("RRTstarsmart", plannerStr))
+    {
+        *plannerPtr = PLANNER_RRTSTARSMART;
     }
     else if (boost::iequals("PRMstar", plannerStr))
     {
--- a/demos/OptimalPlanning.py
+++ b/demos/OptimalPlanning.py
@@ -174,6 +174,8 @@ def allocatePlanner(si, plannerType):
         return og.BFMT(si)
     elif plannerType.lower() == "informedrrtstar":
         return og.InformedRRTstar(si)
+    elif plannerType.lower() == "rrtstarsmart":
+        return og.RRTstarsmart(si)
     elif plannerType.lower() == "prmstar":
         return og.PRMstar(si)
     elif plannerType.lower() == "rrtstar":
@@ -265,7 +267,7 @@ if __name__ == "__main__":
 
     # Add a filename argument
     parser.add_argument('-t', '--runtime', type=float, default=1.0, help='(Optional) Specify the runtime in seconds. Defaults to 1 and must be greater than 0.')
-    parser.add_argument('-p', '--planner', default='RRTstar', choices=['BITstar', 'FMTstar', 'BFMTstar', 'InformedRRTstar', 'PRMstar', 'RRTstar'], help='(Optional) Specify the optimal planner to use, defaults to RRTstar if not given.') # Alphabetical order
+    parser.add_argument('-p', '--planner', default='RRTstar', choices=['BITstar', 'FMTstar', 'BFMTstar', 'InformedRRTstar', 'RRTstarsmart', 'PRMstar', 'RRTstar'], help='(Optional) Specify the optimal planner to use, defaults to RRTstar if not given.') # Alphabetical order
     parser.add_argument('-o', '--objective', default='PathLength', choices=['PathClearance', 'PathLength', 'ThresholdPathLength', 'WeightedLengthAndClearanceCombo'], help='(Optional) Specify the optimization objective, defaults to PathLength if not given.') # Alphabetical order
     parser.add_argument('-f', '--file',  default=None, help='(Optional) Specify an output path for the found solution path.')
     parser.add_argument('-i', '--info', type=int, default=0, choices=[0, 1, 2], help='(Optional) Set the OMPL log level. 0 for WARN, 1 for INFO, 2 for DEBUG. Defaults to WARN.')
--- a/src/ompl/geometric/planners/rrt/RRTstar.h
+++ b/src/ompl/geometric/planners/rrt/RRTstar.h
@@ -40,6 +40,7 @@
 #include "ompl/geometric/planners/PlannerIncludes.h"
 #include "ompl/base/OptimizationObjective.h"
 #include "ompl/datastructures/NearestNeighbors.h"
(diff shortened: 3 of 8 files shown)


mamoll:
Have you done any comparison with @gammell's BIT\* or InformedRRT\*? RRT\*-SMART seems similar in spirit. The reason I am asking is that the RRT\* code is getting more and more bloated with bells and whistles. So while I appreciate your contribution, I am also trying to be mindful of end users who might find all the possible options just confusing.

fahadislam:
You are right. Although the key idea is to improve the convergence rate but the technique used in both the algorithms is different. Gammell also cited it in his Informed RRT* paper. The reason why I want to contribute it to OMPL is to allow more people to use it for their research and comparisons. I got more motivated because of the recent citations that it received. Most notable of which is by:
Frazzoli, Emilio. Sampling-based motion planning algorithms for dynamical systems. Diss. Massachusetts Institute of Technology, 2015. 
Another recent one is from CMU.
Choudhury, Sanjiban, et al. "Regionally accelerated batch informed trees (RABIT*): A framework to integrate local information into optimal path planning." 2016 IEEE International Conference on Robotics and Automation (ICRA). IEEE, 2016.
It has also been cited by GATECH, KAIST and many other universities.

I haven't done any comparisons with Informed-RRT* or BIT* yet. What do you suggest?

gammell:
As I understand it, the goal of RRT\*-Smart is the same as Inf. RRT\* (to use sampling to improve the convergence rate of RRT\* once a solution is found) but they do it in different ways. RRT\*-Smart biases sampling around the current solution.

@fahadislam, I wonder if it could be implemented as a "path-biasing" informed sampler? Then `RRTstar.cpp` would only need to include a flag or two *and* the technique could be (theoretically) used with other planners.

@mamoll, I do agree with the bloat concern and think it needs to be addressed. Would it be worthwhile making this code an abstract base class (e.g., `BaseRRTstar`) and then defining `RRTstar` as a derived planner that disables *all* the bell-and-whistle options? This would mirror how the derived RRT\* planners are implemented. This could be done with an eye to long-term maintenance by moving the bell-and-whistle `Planner::declareParam<...>(...)` calls to the individual related derived classes... This *will* make it harder to try 'mixed' techniques though... and probably has other issues I'm overlooking (e.g., too many derived planners?).

demos/OptimalPlanning.cpp
```diff
@@ -74,6 +75,7 @@ enum optimalPlanner
     PLANNER_FMTSTAR,
     PLANNER_BFMTSTAR,
     PLANNER_INF_RRTSTAR,
+    PLANNER_RRTSTARSMART,
```
gammell: As described by the comment in the code, I think it is best to keep this list in alphabetical order.
If not, at least remove my comment.

demos/OptimalPlanning.cpp
```diff
@@ -418,7 +425,7 @@ bool argParse(int argc, char** argv, double* runTimePtr, optimalPlanner *planner
     desc.add_options()
         ("help,h", "produce help message")
         ("runtime,t", bpo::value<double>()->default_value(1.0), "(Optional) Specify the runtime in seconds. Defaults to 1 and must be greater than 0.")
-        ("planner,p", bpo::value<std::string>()->default_value("RRTstar"), "(Optional) Specify the optimal planner to use, defaults to RRTstar if not given. Valid options are BITstar, CForest, FMTstar, BFMTstar, InformedRRTstar, PRMstar, and RRTstar.") //Alphabetical order
+        ("planner,p", bpo::value<std::string>()->default_value("RRTstar"), "(Optional) Specify the optimal planner to use, defaults to RRTstar if not given. Valid options are BITstar, CForest, FMTstar, BFMTstar, InformedRRTstar, RRTstarsmart, PRMstar, and RRTstar.") //Alphabetical order
```
gammell: Again, either keep it in alphabetical order or remove the comment saying that it is (my preference is alphabetical order but that may just be me).

demos/OptimalPlanning.py
```diff
@@ -174,6 +174,8 @@ def allocatePlanner(si, plannerType):
         return og.BFMT(si)
     elif plannerType.lower() == "informedrrtstar":
         return og.InformedRRTstar(si)
+    elif plannerType.lower() == "rrtstarsmart":
```
gammell: Same thing, alphabetical order or remove comment please.

demos/OptimalPlanning.cpp
```diff
@@ -48,6 +48,7 @@
 #include <ompl/geometric/planners/fmt/FMT.h>
 #include <ompl/geometric/planners/fmt/BFMT.h>
 #include <ompl/geometric/planners/rrt/InformedRRTstar.h>
+#include <ompl/geometric/planners/rrt/RRTstarsmart.h>
```
gammell: Again, order or remove comment.

demos/OptimalPlanning.cpp
```diff
@@ -167,6 +169,11 @@ ob::PlannerPtr allocatePlanner(ob::SpaceInformationPtr si, optimalPlanner planne
             return std::make_shared<og::InformedRRTstar>(si);
             break;
         }
+        case PLANNER_RRTSTARSMART:
```
gammell: I appreciate that these are in the same order as the enum definition. If you reorder that, please keep this the same.

demos/OptimalPlanning.py
```diff
@@ -265,7 +267,7 @@ def plan(runTime, plannerType, objectiveType, fname):
 
     # Add a filename argument
     parser.add_argument('-t', '--runtime', type=float, default=1.0, help='(Optional) Specify the runtime in seconds. Defaults to 1 and must be greater than 0.')
-    parser.add_argument('-p', '--planner', default='RRTstar', choices=['BITstar', 'FMTstar', 'BFMTstar', 'InformedRRTstar', 'PRMstar', 'RRTstar'], help='(Optional) Specify the optimal planner to use, defaults to RRTstar if not given.') # Alphabetical order
+    parser.add_argument('-p', '--planner', default='RRTstar', choices=['BITstar', 'FMTstar', 'BFMTstar', 'InformedRRTstar', 'RRTstarsmart', 'PRMstar', 'RRTstar'], help='(Optional) Specify the optimal planner to use, defaults to RRTstar if not given.') # Alphabetical order
```
gammell: Again, reorder or remove comment?

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -215,6 +216,24 @@ namespace ompl
                 return useInformedSampling_;
             }
 
+            /** \ Use intelligent sampling to bias samples towards the beacons */
+            void setIntelligentSampling(bool intelligentSampling);
```
gammell: I don't like this as a name for the option. It does not explain what it is doing. How about `setSolutionBiasedSampling` or `setBeaconSampling` or something that is a little more descriptive.

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -450,6 +469,15 @@ namespace ompl
             /** \brief Option to use informed sampling */
             bool useInformedSampling_;
 
+            /** \brief Option to use intelligent sampling */
+            bool                                           useIntelligentSampling_;
```
gammell: Same as comment re: set/get function.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -87,6 +92,10 @@ ompl::geometric::RRTstar::RRTstar(const base::SpaceInformationPtr &si)
     Planner::declareParam<bool>("pruned_measure", this, &RRTstar::setPrunedMeasure, &RRTstar::getPrunedMeasure, "0,1");
     Planner::declareParam<bool>("informed_sampling", this, &RRTstar::setInformedSampling, &RRTstar::getInformedSampling,
                                 "0,1");
+    Planner::declareParam<bool>("intelligent_sampling", this, &RRTstar::setIntelligentSampling, &RRTstar::getIntelligentSampling,
```
gammell: Option name again?

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -450,6 +469,15 @@ namespace ompl
...
+
+            /** \brief The ratio of bias towards the beacons */
+            double                                         biasingRatio_;
+
+            /** \brief Flag true after first path is found */
+            bool                                            pathFound_;
```
gammell: This shouldn't be *strictly* necessary, a solution is found if

1. `opt_->isFinite(bestCost_) == true`, or
2. ` goalMotions_.size() > 0`

If it is desirable to keep this state so explicitly stored (which it may be), then this variable needs to be always maintained. You're only setting it when RRT\*-Smart is in use, which is really confusing.

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -235,6 +244,7 @@ ompl::base::PlannerStatus ompl::geometric::RRTstar::solve(const base::PlannerTer
     const base::ReportIntermediateSolutionFn intermediateSolutionCallback = pdef_->getIntermediateSolutionCallback();
 
     Motion *solution = lastGoalMotion_;
+    Motion *solution_tmp       = lastGoalMotion_;
```
gammell: FYI, `lastGoalMotion_`, it is a misleading variable. It holds the last path to a goal region, not necessarily the best path. It is really only used for `Planner::getPlannerData`, and I am planning to remove it and just iterate over `goalMotions_` when necessary.

src/ompl/geometric/planners/rrt/RRTstar.h
```diff
@@ -478,6 +506,12 @@ namespace ompl
             /** \brief Number of iterations the algorithm performed */
             unsigned int iterations_;
 
+            /** \brief To access reduceVertices function to generate beacons */
+            PathSimplifierPtr                               psimp_;
+
+            /** \brief States used for intelligent biasing */
+            std::vector<base::State*>                       beacons;
```
gammell: Are these `State*`s being properly allocated/freed?

gammell:
I actually spent yesterday looking at `RRTstar.cpp`, so I added some comments to the proposed changes. I hope you don't mind.

src/ompl/geometric/planners/rrt/RRTstarsmart.h
```diff
@@ -0,0 +1,60 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2014, University of Toronto
```
gammell: You may want to change this?

src/ompl/geometric/planners/rrt/src/RRTstarsmart.cpp
```diff
@@ -0,0 +1,51 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2014, University of Toronto
```
gammell: This as well.

fahadislam:
@gammell Thanks for your comments and doing a detailed code review! I will make the necessary edits and make another commit. 
By adding a path-biasing informed sampler, do you mean to add a new sampler class like the informed sampler one or including a path biasing functionality in the existing informed sampler class? Can you please elaborate?

gammell:
I meant create a new class derived from `InformedSampler`, just like was done for `PathLengthDirectInfSampler` and `RejectionInfSampler`. See: http://ompl.kavrakilab.org/classompl_1_1base_1_1InformedSampler.html

A path-biased sampler wouldn't exactly meet the current description of `InformedSampler`, but it does meet the broader underlying idea of using solution information to modify sampling, so the description of the base class could be changed.

All that said, I wouldn't start making any changes until @mamoll weighs in on wanting this feature in the first place.

mamoll:
@fahadislam @gammell I can't look at this within the next couple days, but should have time next week. In the meantime, please have a look at Scott Paulin's proposed changes on BitBucket: https://bitbucket.org/ompl/ompl/issues/257/refactor-rrtstar-pulling-out-functionality

fahadislam:
@mamoll Did you get a chance to look into it? It was the quickest implementation that I could do just to convey the idea across. I will be happy to make revisions while considering the bloat concerns as per your suggestions.

mamoll:
I still haven't had time yet. The semester has just started, tomorrow is World MoveIt! Day, etc. I have been busy, but I haven't forgotten about this. Any opinion or ideas on how we could restructure RRT* so that it doesn't become even more bloated over time?

PR #86, opened 2016-08-15
Implementation of RRT# and RRTX

florianhauer:
- Implementation of RRT# and RRTX
- Adding check of max range in RRT* when using k-nearest neighbors

### DIFF
--- a/demos/Diagonal.cpp
+++ b/demos/Diagonal.cpp
@@ -1,0 +1,156 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2016, Rice University
+*  All rights reserved.
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of the Rice University nor the names of its
+*     contributors may be used to endorse or promote products derived
+*     from this software without specific prior written permission.
+*
+*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+*  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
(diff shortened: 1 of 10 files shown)


src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -368,7 +368,8 @@ ompl::base::PlannerStatus ompl::geometric::RRTstar::solve(const base::PlannerTer
                 for (std::vector<std::size_t>::const_iterator i = sortedCostIndices.begin();
                      i != sortedCostIndices.begin() + nbh.size(); ++i)
                 {
-                    if (nbh[*i] == nmotion || si_->checkMotion(nbh[*i]->state, motion->state))
+                    if (nbh[*i] == nmotion || (si_->checkMotion(nbh[*i]->state, motion->state) 
+                        && ( !useKNearest_ || si_->distance(nbh[*i]->state, motion->state) < maxDistance_) ) )
```
mamoll: Why is this change needed? Since the checkMotion call is probably more expensive, wouldn't you want to swap the order of the && operands?

src/ompl/geometric/planners/rrt/RRTX.h
```diff
@@ -0,0 +1,470 @@
...
+
+            /** \brief The number of attempts to make at informed sampling */
+            unsigned int                                    numSampleAttempts_;
+
+            ///////////////////////////////////////
+            // Planner progress property functions
+            std::string numIterationsProperty() const
+            {
+                return boost::lexical_cast<std::string>(numIterations());
+            }
```
mamoll: We're trying to use the standard C++ library when possible. In this case, you can use [std::to_string](http://en.cppreference.com/w/cpp/string/basic_string/to_string).

src/ompl/geometric/planners/rrt/src/RRTX.cpp
```diff
@@ -0,0 +1,799 @@
...
+#include "ompl/base/objectives/PathLengthOptimizationObjective.h"
+#include "ompl/base/Goal.h"
+#include "ompl/base/goals/GoalState.h"
+#include "ompl/util/GeometricEquations.h"
+#include "ompl/base/samplers/InformedStateSampler.h"
+#include "ompl/base/samplers/informed/RejectionInfSampler.h"
+#include <algorithm>
+#include <limits>
+#include <boost/math/constants/constants.hpp>
+#include <boost/make_shared.hpp>
```
mamoll: Don't need this anymore.

demos/Diagonal.cpp
```diff
@@ -0,0 +1,156 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2016, Rice University
```
mamoll: This should probably change to Georgia Tech

mamoll:
Thanks for this contribution. Overall, it looks good. There are some small things (like changing NULL to nullptr) that I can fix after merging. The main reason I'm not merging yet is because I want to understand the change to RRT\*.

PR #87, opened 2016-08-30
Renaming RRTX to RRTXstatic + Disclaimer

florianhauer:
After discussion with Michael Otte (author of the RRTX paper), we decided to rename RRTX into RRTXstatic because only the static part of the algorithm is implemented (dynamic obstacles and robot motion are not available in this implementation).
We also added a disclaimer explaining it.

It aims at reducing confusion for the user and preventing users from naively using it in dynamical environment.


mamoll:
I applied the fixes manually due to conflicts.

PR #94, opened 2016-10-27
Changed RRTXstatic to use r_rrt and k_rrt instead of r_rrg and k_rrg

florianhauer:
Similarly to the recent update on RRTstar, the constants for the radius of the ball and the number of nearest neighbors has been updated in RRTXstatic to guarantee optimality.


mamoll:
:+1:

PR #96, opened 2016-12-18
Flann CMakeModule needs to include libs in 1.1.1. 

lloydc99:
The Flann CMakeModule does not include libraries. The ROS indigo version of MoveIt won't compile from source and needs the 1.1 interface.


mamoll:
Why not simply use the current version of the Flann CMakeModule?:
https://github.com/ompl/ompl/blob/master/CMakeModules/Findflann.cmake

lloydc99:
I was going for the smallest change.
I'll try with the latest .cmake.

PR #98, opened 2017-03-15
Update Exception.h

aminaattobi:
This fixes the following compile time error: 
 'ompl::Exception::~Exception(void) throw()' : is not a special member function which can be defaulted

### DIFF
--- a/src/ompl/util/Exception.h
+++ b/src/ompl/util/Exception.h
@@ -57,7 +57,7 @@ namespace ompl
         {
         }
 
-        ~Exception() throw() override = default;
+        ~Exception() override = default;
     };
 }
 


mamoll:
What compiler and OS combination is producing a compilation error with the code in Exception.h?

aminaattobi:
Visual C++ v120 Compiler on Windows.

mamoll:
Your patch actually changes the semantics a little bit. Can you see if this also fixes the error:
```
diff -r 6c117baf4717 src/ompl/util/Exception.h
--- a/src/ompl/util/Exception.h Tue Mar 14 11:54:07 2017 -0500
+++ b/src/ompl/util/Exception.h Wed Mar 15 08:30:38 2017 -0500
@@ -57,7 +57,7 @@
         {
         }
 
-        ~Exception() throw() override = default;
+        ~Exception() noexcept override = default;
     };
 }
```

zkingston:
@aminaattobi does Mark's suggested patch fix the problem? If yes, I would like to close this pull.

PR #101, opened 2017-05-23
Fix Yocto builds

rojkov:
These two commits make OMPL buildable with the latest Yocto. Particularly it fixes a compatibility issue with gcc6 and makes setting RPATH optional to conform with Opembedded-Core's QA checks.


mamoll:
Looks good, thanks!

PR #103, opened 2017-10-20
BIT* and PathLengthDirectInfSampler fixes

gammell:
1. Allow `BITstar::setup()` to run without a problem definition and delay the necessary work until `BITstar::solve()` calls `Planner::checkValidity()` (which will then recall `BITstar::setup()`)
2. Allow `PathLengthDirectInfSampler` to support `STATE_SPACE_UNKNOWN` with a warning.
3. Initialize `BITstar` member variables in the header instead of constructor as per style.
4. Revert `BITstar::ImplicitGraph::removeSample()` and `BITstar::ImplicitGraph::removeVertex()` to taking a local copy of the removed data to avoid a low-frequency segfault. [See issue here](https://bitbucket.org/ompl/ompl/issues/364/code-cleanup-breaking-bit).
5. Fix segfault in `BITstar::ImplicitGraph::clear()` caused by an already cleared pointer.

### DIFF
--- a/src/ompl/base/samplers/informed/src/PathLengthDirectInfSampler.cpp
+++ b/src/ompl/base/samplers/informed/src/PathLengthDirectInfSampler.cpp
@@ -91,17 +91,26 @@ namespace ompl
             }
 
             // Check that the provided statespace is compatible and extract the necessary indices.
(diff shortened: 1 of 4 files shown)


mamoll:
Looks good to me. Thanks!

PR #106, opened 2018-01-04
Fix for Issue #373 on BitBucket.

gammell:
https://bitbucket.org/ompl/ompl/issues/373/informedrrtstar-segfault

### DIFF
--- a/src/ompl/geometric/planners/rrt/src/RRTstar.cpp
+++ b/src/ompl/geometric/planners/rrt/src/RRTstar.cpp
@@ -200,7 +200,7 @@ ompl::base::PlannerStatus ompl::geometric::RRTstar::solve(const base::PlannerTer
         allocSampler();
     }
 
-    OMPL_INFORM("%s: Starting planning with %u states already in datastructure", getName().c_str(), nn_->size());
+    OMPL_INFORM("%s: Started planning with %u states. Seeking a solution better than %.5f.", getName().c_str(), nn_->size(), opt_->getCostThreshold().value());
 
     if ((useTreePruning_ || useRejectionSampling_ || useInformedSampling_ || useNewStateRejection_) &&
         !si_->getStateSpace()->isMetricSpace())
@@ -213,7 +213,6 @@ ompl::base::PlannerStatus ompl::geometric::RRTstar::solve(const base::PlannerTer
 
     Motion *approxGoalMotion = nullptr;
     double approxDist = std::numeric_limits<double>::infinity();
-    bool sufficientlyShort = false;
 
     auto *rmotion = new Motion(si_);
     base::State *rstate = rmotion->state;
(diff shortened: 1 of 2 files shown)


src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -581,7 +603,8 @@ ompl::base::PlannerStatus ompl::geometric::RRTstar::solve(const base::PlannerTer
                 "%.3f",
                 getName().c_str(), statesGenerated, rewireTest, goalMotions_.size(), bestCost_.value());
 
-    return base::PlannerStatus(addedSolution, approximate);
+    // We've added a solution if newSolution == true, and it is an approximate solution if bestGoalMotion_ == false
+    return base::PlannerStatus(static_cast<bool>(newSolution), !static_cast<bool>(bestGoalMotion_));
```
mamoll: This is more of a style comment. I think this would be better:
```
return base::PlannerStatus(newSolution != nullptr, bestGoalMotion_ == nullptr);
```

src/ompl/geometric/planners/rrt/src/RRTstar.cpp
```diff
@@ -776,12 +799,17 @@ int ompl::geometric::RRTstar::pruneTree(const base::Cost &pruneTreeCost)
...
         while (leavesToPrune.empty() == false)
         {
-            // First empty the leave-to-prune
+            // First empty the current leaves-to-prune
             while (leavesToPrune.empty() == false)
             {
                 // If this leaf is a goal, remove it from the goal set
                 if (leavesToPrune.front()->inGoal == true)
                 {
+                    // Assert that it's not the _best_ goal
```
mamoll: You are not really asserting that. You are still pruning the best goal. Is that really what you want?

mamoll:
The changes look mostly ok to me. I have left some small comments. Note that the unit tests are failing, so your changes broke something else in RRT\*.

gammell:
Yeah, I noticed that the unit test was failing, but when I ran it manually on my machine without the changes it failed as well... That version of master was from before the holidays.

I haven't had a chance to look into it more, but I've made your changes and rebased to the latest master so when I get some time I'll take a closer look to see what's going on.

gammell:
When I locally run the test (an Ubuntu 16.04 VM) on the current master, the `test_2dcircles_opt_geometric` test gives an error with `RRTstar`. When I run this branch that test gives errors for both `RRTstar` and `CFOREST`. Obviously something more has broken and I'll take a closer look at the branch, but do you have any thoughts as to why the test is failing for me on master?

mamoll:
That's odd. Travis CI is running 14.04 and the latest commit on master passes the tests:
https://travis-ci.org/ompl/ompl/builds/322327633
My local build on OS X also passes the tests. Maybe your VM/host machine is just very slow?

gammell:
Hm,
So this motivated me to revisit why I was running in a 16.04 VM instead of natively in 14.04 in the first place (it had to do with conflicts with the build system here at ORI). Anywho, I got that all sorted out and am out of the VM,  but I'm still getting those RRT* failures in master... I'm going to focus on the failures I've added tho'.

gammell:
For future information:
I've figured out my problems on master. It was the difference in performance between `Debug` and `Release` modes. In `Debug` mode, the test failed 4 out of 10 runs. In `Release` mode, the test never failed.

mamoll:
Thanks for the fixes!

eirikwor:
Thanks for your efforts on this issue!

mamoll:
@eirikwor, can you confirm this fixes the issues you had, so that I can close the issue?

eirikwor:
Absolutely. This seems to have fixed the issue I was having!

PR #108, opened 2018-02-15
BIT* radius as published.

gammell:
Correct the radius used by BIT* to be the same as described in the associated papers (e.g., RRT* radius).


gammell:
I had been testing the different connection radii in the literature and appear to have accidentally changed the radius used by BIT* away from it's associated papers. This corrects that mistake.

Sorry.

PR #109, opened 2018-02-20
Fix Issue #380

gammell:
Remove shared_ptr cycle in BITstar by using raw pointers.


gammell:
On second thought, there's a more thorough fix for this.
I'll update the PR later tonight.

gammell:
On third thought I was right the first time. This is fine.

PR #110, opened 2018-05-15
Add missing parameter in TypedSpaceInformation::cloneTypedState()

ndantam:
The missing parameter is caught by newer GCC versions doing stricter template checks.


mamoll:
Thanks, Neil!

PR #112, opened 2019-01-09
Do not localize params

simonschmeisser:
This is a very basic fix for the issue described in https://bitbucket.org/ompl/ompl/issues/496/localization-errors-or-the-tale-of-decimal

Details are given in the issue but in brief it solves the following problem: after the switch from `boost::lexical_cast` to `std::stof` and `std::to_string` parameters are handled as locale dependent. I.e. "longest_valid_segment_fraction" needs to be passed in as "0,1" if you use an European locale. Also it will be returned in the same format. 

This could be improved by having a ompl::to_string that handles all types correctly (ie using std::to_string for integers and this code for real variables) but I wanted to have some feedback first. Another possibility would be to do some template specialization on `SpecificParam::getValue` or do some `enable_if` template stuff

### DIFF
--- a/src/ompl/base/GenericParam.h
+++ b/src/ompl/base/GenericParam.h
@@ -42,6 +42,7 @@
 #include <functional>
 #include <iostream>
 #include <string>
+#include <sstream>
 #include <utility>
 #include <vector>
 #include <map>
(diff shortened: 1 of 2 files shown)


simonschmeisser:
has been fixed separately, thanks!

PR #115, opened 2019-06-26
fix triangle inequality check in StateSpace::sanityChecks

juleswh:
Hi,
It seems to me that this check is wrong. It ensures that d(s1,s2) = d(s1,s3)+d(s3,s2), while it is expected to ensure that (triangle inequality) d(s1,s2) <= d(s1,s3)+d(s3,s2). Otherwise it may be the exception message that should be corrected.

Here is how I would have done it.

Hope I'm not misunderstanding something here :)

Cheers,
JW


mamoll:
Thanks!

PR #116, opened 2019-07-04
Bug/dependent parameters not updated for EST and TRRT

ChamzasKonstantinos:
Some planners have dependent parameters  on other primary parameters which can be set externally using the setParam().  However, if the primary  parameters  change and setup is not called afterwards, in the case of EST planner and the TRRT planner  the dependent parameters are not updated . This PR fixes this by updating the dependent  parameters when the setters of the original parameters are called.

### DIFF
--- a/src/ompl/geometric/planners/est/src/BiEST.cpp
+++ b/src/ompl/geometric/planners/est/src/BiEST.cpp
@@ -57,15 +57,12 @@ void ompl::geometric::BiEST::setup()
 {
     Planner::setup();
 
-    if (maxDistance_ < 1e-3)
-    {
-        tools::SelfConfig sc(si_, getName());
-        sc.configurePlannerRange(maxDistance_);
+    tools::SelfConfig sc(si_, getName());
+    sc.configurePlannerRange(maxDistance_);
 
-        // Make the neighborhood radius smaller than sampling range to
-        // keep probabilities relatively high for rejection sampling
-        nbrhoodRadius_ = maxDistance_ / 3.0;
-    }
+    // Make the neighborhood radius smaller than sampling range to
+    // keep probabilities relatively high for rejection sampling
+    nbrhoodRadius_ = maxDistance_ / 3.0;
 
     if (!nnStart_)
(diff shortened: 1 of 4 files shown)


src/ompl/geometric/planners/est/src/BiEST.cpp
```diff
@@ -57,15 +57,12 @@ void ompl::geometric::BiEST::setup()
 {
     Planner::setup();
 
-    if (maxDistance_ < 1e-3)
```
mamoll: Why remove the if statement? Now the selfconfig code is run each time setup() is called and overrides any user-specified range setting.

src/ompl/geometric/planners/rrt/TRRT.h
```diff
@@ -167,6 +167,7 @@ namespace ompl
             void setInitTemperature(double initTemperature)
             {
                 initTemperature_ = initTemperature;
+                temp_ = initTemperature_;
```
mamoll: This shouldn't be necessary, since this is also done by setup() and clear().

src/ompl/geometric/planners/est/src/BiEST.cpp
mamoll: Hm, maybe this if statement should actually be added to, e.g., ompl::geometric::EST::setup (and maybe other planners).

src/ompl/geometric/planners/rrt/TRRT.h
ChamzasKonstantinos: Yes but what if someone calls setup() then calls solve() then sets the initTemperature and then calls solve() again.

src/ompl/geometric/planners/est/src/BiEST.cpp
ChamzasKonstantinos: From my understanding selfConfig for the Range only overrides if the parameter is set at less than 0 see at [line91](http://ompl.kavrakilab.org/SelfConfig_8cpp_source.html).   This is how it is done in all the planners that use the range parameter e.g, RRT, RRTConnect  so I removed it for consistency.  It seems redundant to have this check in both places. I don't know were they should be though

PR #118, opened 2019-07-17
Serve video content from HTTPS

wxmerkt:
Nowadays browsers automatically open HTTPS if available and disallow serving content from non-secure sites if you are browsing with HTTPS. As such, the currently published documentation's gallery page does not show any embedded videos.

This PR updates the YouTube and Vimeo URLs to HTTPS and also updated some others, e.g., ompl.kavrakilab.org.


mamoll:
Thanks!

PR #119, opened 2019-09-10
Reorganize planner termination conditions and add cost convergence termination condition.

mamoll:
- Add documentation for all planner termination conditions on one page
accessible from the web site's documentation menu.
- Update Python bindings.

### DIFF
--- a/doc/markdown/plannerTerminationConditions.md
+++ b/doc/markdown/plannerTerminationConditions.md
@@ -1,0 +1,23 @@
+# Planner Termination Conditions {#plannerTerminationConditions}
+
+OMPL uses the ompl::base::PlannerTerminationCondition concept to let planning algorithms know that an attempt to solve a motion planning problem should be terminated. This is done by passing an instance of an ompl::base::PlannerTerminationCondition to a planner's solve method. A PlannerTerminationCondition instance is essentially a functor that can be called with zero arguments that returns `true` when the planner should terminate. For convenience, there is also a cast-to-bool operator defined so the following two is statements are equivalent:
+
+~~~{.cpp}
+// ptc is an instance of ompl::base::PlannerTerminationCondition
+if (ptc()) return;
+if (ptc) return;
+~~~
+
+OMPL implements several different conditions that you can use:
+
+- `timedPlannerTerminationCondition(duration)`: a **function** that returns a PlannerTerminationCondition instance that causes the planner to terminate after a fixed amount of time has been exceeded.
+- `timedPlannerTerminationCondition(duration, interval)`: a **function** that returns a PlannerTerminationCondition instance that causes the planner to terminate after a fixed amount of time has been exceeded. Unlike the previous function, this one spawns a separate thread where the time passed is checked every `interval` seconds. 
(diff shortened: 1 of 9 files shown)


mamoll:
@henningkayser can you have a look to see if this does what you had in https://github.com/ros-planning/moveit/pull/1557? I used the more general cost infrastructure, so that code will do the right thing when, e.g., *minimizing* path length or *maximizing* path clearance.

@BryceStevenWilley if you have any thoughts about this, feel free to comment as well. 

I am planning to write some unit tests for the different types of PlannerTerminationCondtions, which are completely untested at the moment.

src/ompl/base/terminationconditions/src/CostConvergenceTerminationCondition.cpp
```diff
@@ -0,0 +1,65 @@
...
+	    [this](const Planner* /*planner*/, const std::vector<const State*>& /*states*/, const Cost cost) {
+	        this->processNewSolution(cost);
+	    });
+}
+
+void ompl::base::CostConvergenceTerminationCondition::processNewSolution(const ompl::base::Cost solutionCost)
+{
+    ++solutions_;
+    size_t solutions = std::min(solutions_, solutionsWindow_);
+    Cost newCost(((solutions - 1) * averageCost_ + solutionCost.value()) / solutions);
```
BryceStevenWilley: I think averageCost_ should be updated here. I guess those unit tests would find this though :wink:

doc/markdown/plannerTerminationConditions.md
```diff
@@ -0,0 +1,23 @@
...
+
+- `timedPlannerTerminationCondition(duration)`: a **function** that returns a PlannerTerminationCondition instance that causes the planner to terminate after a fixed amount of time has been exceeded.
+- `timedPlannerTerminationCondition(duration, interval)`: a **function** that returns a PlannerTerminationCondition instance that causes the planner to terminate after a fixed amount of time has been exceeded. Unlike the previous function, this one spawns a separate thread where the time passed is checked every `interval` seconds. 
+- `IterationTerminationCondition(numIterations)`: a PlannerTerminationCondition-derived **class** that causes a planner to terminate after its `eval()` method has been called `numIterations` times.
+- `CostConvergenceTerminationCondition(pdef, solutionsWindow, convergenceThreshold)`: a PlannerTerminationCondition-derived **class** that causes an *optimizing* planner  to terminate once the cost for the best path has converged, which is defined as 
+- `exactSolnPlannerTerminationCondition()`: a **function** that returns a PlannerTerminationCondition instance that causes the planner to terminate after an exact solution has been found.
```
BryceStevenWilley:
```suggestion
- `exactSolnPlannerTerminationCondition()`: a **function** that returns a PlannerTerminationCondition instance that causes the planner to terminate after an exact (not approximate) solution has been found.
```

Without that, I always start to wonder if exact means optimal, or resolution exact, or something else.

src/ompl/base/terminationconditions/src/CostConvergenceTerminationCondition.cpp
mamoll: Fixed. Thanks!

PR #682, opened 2019-10-17
Increment the counter when directly sampling PHSs

marlinstrub:
This ensures that the `PathLengthDirectInfSampler` respects the maximum number of iterations as defined by `InformedSampler::numIters_` before giving up on sampling an informed state.


src/ompl/base/samplers/informed/src/PathLengthDirectInfSampler.cpp
```diff
@@ -451,6 +451,9 @@ namespace ompl
                     foundSample = InformedSampler::space_->satisfiesBounds(statePtr);
                 }
                 // No else
+
+                // Infrement the provided counter
```
mamoll: Small typo: "Increment"

marlinstrub: Thanks for spotting it

PR #683, opened 2019-10-17
Nonabstract goal area defined by the limits of a second StateSpace.

gammell:
Hi Mark,

This is a simple version of a goal area defined by the user via a `StateSpace`. I thought it may be useful to the main branch as, while it's limited, there's currently no nonabstract area goals in OMPL.

Jon

### DIFF
--- a/src/ompl/base/goals/src/GoalSpace.cpp
+++ b/src/ompl/base/goals/src/GoalSpace.cpp
@@ -1,0 +1,99 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2019, University of Oxford.
+*  All rights reserved.
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of the University of Oxford nor the names of its
+*     contributors may be used to endorse or promote products derived
+*     from this software without specific prior written permission.
+*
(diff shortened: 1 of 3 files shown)


src/ompl/base/goals/src/GoalSpace.cpp
```diff
@@ -0,0 +1,99 @@
...
+}
+
+void ompl::base::GoalSpace::sampleGoal(base::State *st) const
+{
+    goalSampler_->sampleUniform(st);
+}
+
+unsigned int ompl::base::GoalSpace::maxSampleCount() const
+{
+    return std::numeric_limits<unsigned int>::max();
```
gammell, line 51: @marlinstrub and I were discussing this, and we're not sure whether `maxSampleCount()` is the maximum number of possible goal (i.e., infinite for a continuous region) or the maximum number of goals the goal region can give you (i.e., the max it knows about). I think it may be the second?

mamoll: I think using `max()` is fine. `infinity()` might be "more correct", but without a specific use case where this is needed, I propose we stick with `max()`.

gammell: Right, sorry, I put this comment in a hurry last night and should have explained better. My concern is that if you look at how this function is used elsewhere, e.g., `PlannerInputStates::haveMoreGoals()`, having it be infinity will cause problems any time you seek to get all available goals since `haveMoreGoals()` returns (Line 342, [Planner.cpp](http://ompl.kavrakilab.org/Planner_8cpp_source.html)):
```
return sampledGoalsCount_ < pdef_->getGoal()->as<GoalSampleableRegion>()->maxSampleCount()
```
So any planner that gets all available goals will end up hanging indefinitely when given a `GoalSpace`...

Is that 'correct' behaviour that should be addressed with a `max_goal` parameter in the planner that limits the number of goal samples drawn?

gammell: As another example, the way MoveIt! interfaces to OMPL through `GoalStates` means that in that case this method is returning the number of goals currently available to the planner, since `GoalStates` returns the size of the internal vector, and MoveIt adds to that vector as it discovers goals...

... at least as I understand it.

PR #694, opened 2020-02-25
Fix issue in ConstrainedStateSpace geodesicInterpolate when planner range value is small

Levi-Armstrong:
I ran into an issue when trying to used constrained planning functionality in OMPL, where if you made the range value for a planner  small (0.01) it would try to index outside the `geodesic`. These changes fix the issue but not being familiar with the code I am not certain this is the correct fix and there may be some other issue elsewhere in the code.


mamoll:
Thanks, @Levi-Armstrong!

PR #695, opened 2020-02-27
Feature/deterministic sampling

palmieri:
Implementation of a deterministic state sampler. The implementation allows to load and draw samples from a precomputed sequence (saved in an external file) or from a Halton Sequence.


demos/DeterministicSampling.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include <iostream>
+
+namespace ob = ompl::base;
+
+int main(int /*argc*/, char ** /*argv*/)
+{
+    std::cout << "Testing the Halton Sequence..." << std::endl;
+
+    ob::HaltonSequence hs5d(5);
+
```
mamoll: This isn't much of a test. I'd suggest turning this into a proper unit test (and moved to `tests/base`) where the first n samples are compared with the hard-coded expected output, possibly extending the test to, say, 1d through 5d.

demos/DeterministicSamplingPlanning.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+            // set state validity checking for this space
+            ss_->setStateValidityChecker([this](const ob::State *state) { return isStateValid(state); });
+            space->setup();
+            ss_->getSpaceInformation()->setStateValidityCheckingResolution(1.0 / space->getMaximumExtent());
+            ss_->setPlanner(std::make_shared<og::PRMstar>(ss_->getSpaceInformation()));
+
+            // set the deterministic sampler
+            // 2D space, no need to specify bases specifically
+            space->setStateSamplerAllocator(std::bind(&Plane2DEnvironment::allocateHaltonStateSamplerRealVector, this, std::placeholders::_1, 2, std::vector<unsigned int>{2,3}));
```
mamoll: This is the main difference with `demos/Point2DPlanning.cpp`. Rather than duplicating the code and creating a near-identical demo, can you just extend Point2DPlanning.cpp? Use an extra boolean flag for the constructor called, say, `useDeterministicSampling` to conditionally enable this. Use some basic command line parsing to allow the user to enable this via the command line.

src/ompl/base/samplers/deterministic/src/HaltonSequence.cpp
```diff
@@ -0,0 +1,135 @@
...
+            for (auto &seq : halton_sequences_1d_)
+            {
+                samples.push_back(seq.sample());
+            }
+            return samples;
+        }
+
+        void HaltonSequence::setBasesToPrimes()
+        {
+            // set the base of the halton sequences to the first n prime numbers, where n is dimensions
```
mamoll: I'd just use boost::math::prime https://www.boost.org/doc/libs/1_72_0/libs/math/doc/html/math_toolkit/number_series/primes.html

less code is better :-)

src/ompl/base/samplers/deterministic/PrecomputedSequence.h
```diff
@@ -0,0 +1,78 @@
...
+#include "ompl/base/samplers/deterministic/DeterministicSequence.h"
+
+#include <string>
+
+namespace ompl
+{
+    namespace base
+    {
+        /** \brief General realization for a sampler of precomputed sequences or sets. */
+        class PrecomputedSequence : public DeterministicSequence
```
mamoll: Where is this class used? Why is it needed?

src/ompl/base/samplers/DeterministicStateSampler.h
```diff
@@ -0,0 +1,172 @@
...
+    */
+
+        /** \brief An abstract class for the concept of using deterministic sampling sequences
+        to decrease the dispersion of the samples. */
+        class DeterministicStateSampler : public StateSampler
+        {
+        public:
+            enum DeterministicSamplerType
+            {
+                HALTON
```
mamoll: It'd nice if there was support for at least one more deterministic sampler (e.g., Hammersley). This would help establish that the API is general enough to accommodate other types.

Why isn't there a PrecomputedStateSampler type?

src/ompl/base/samplers/DeterministicStateSampler.h
```diff
@@ -0,0 +1,172 @@
...
+
+            /** \brief Constructor, which creates the sequence internally based on the specified sequence type.
+            Uses the default constructor for the sequence.*/
+            DeterministicStateSampler(const StateSpace *space,
+                                      DeterministicSamplerType type = DeterministicSamplerType::HALTON);
+            /** \brief Constructor that takes a pointer to a DeterministicSequence and uses that object instead
+            of its own. This can be used to apply non default options to the deterministic sequence.*/
+            DeterministicStateSampler(const StateSpace *space, std::shared_ptr<DeterministicSequence> sequence_ptr);
+
+            void sampleUniform(State *state) override = 0;
```
mamoll: You should provide a default implementation that uses StateSpace::copyFromReals to copy the doubles from an std::vector into the state. You only need to override this behavior if you need to scale the coordinates.

src/ompl/base/samplers/DeterministicStateSampler.h
```diff
@@ -0,0 +1,172 @@
...
+            void sampleUniformNear(State *state, const State *near, double distance) override = 0;
```
mamoll: Provide a default implementation for this method and the next one that calls OMPL_ERROR to say this is unsupported. This will eliminate these methods in the derived classes.

demos/DeterministicSampling.cpp
```diff
@@ -0,0 +1,58 @@
...
+*  modification, are permitted provided that the following conditions
+*  are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of the Rice University nor the names of its
```
mamoll: replace "the Rice University" with "Robert Bosch GmbH" in all licenses.

src/ompl/base/samplers/deterministic/HaltonSequence.h
```diff
@@ -0,0 +1,103 @@
...
+        \ref HaltonSequence1D Realization of the Halton sequence for the generation of
+        arbitrary dimensional, low-dispersion sequences.
+        @par External documentation
+        Implementation follows https://en.wikipedia.org/wiki/Halton_sequence.
+        A more efficient implementation might be found in
+        Struckmeier, Jens. "Fast generation of low-discrepancy sequences." Journal of Computational and
+        Applied Mathematics 61.1 (1995): 29-41.
+        \brief Realization of the Halton sequence for the generation of
+        arbitrary dimensional, low-dispersion sequences.
+
```
mamoll: Your forgot `*/` here

src/ompl/base/samplers/deterministic/HaltonSequence.h
```diff
@@ -0,0 +1,103 @@
...
+        class HaltonSequence1D
```
mamoll: Can this just be a private class inside the class HaltonSequence? Or is this useful on its own?

src/ompl/base/samplers/deterministic/src/HaltonSequence.cpp
```diff
@@ -0,0 +1,135 @@
...
+        {
+            setBasesToPrimes();
+        }
+
+        HaltonSequence::HaltonSequence(unsigned int dimensions, std::vector<unsigned int> bases)
+          : DeterministicSequence(dimensions), halton_sequences_1d_(dimensions)
+        {
+            if (bases.size() != dimensions)
+            {
+                std::cout << "Warning: Number of bases does not match dimensions. Using first n primes instead.";
```
mamoll: don't use `std::cout` directly in any of your code. Use the OMPL_WARNING, OMPL_ERROR, etc. macros to print output.

src/ompl/base/samplers/deterministic/src/HaltonSequence.cpp
```diff
@@ -0,0 +1,135 @@
...
+            }
+            else
+            {
+                for (size_t i = 0; i < bases.size(); ++i)
```
mamoll: could use a range-based for-loop (`for(auto base : bases) ...`)

src/ompl/base/samplers/deterministic/src/PrecomputedSequence.cpp
```diff
@@ -0,0 +1,108 @@
...
+                sample_set_.emplace_back(new_sample);
+                if (max_samples != 0 && sample_set_.size() >= max_samples)
+                    break;
+            }
+        }
+
+        void PrecomputedSequence::shuffleSamples()
+        {
+            std::srand(0);
+            std::random_shuffle(sample_set_.begin(), sample_set_.end());
```
mamoll: please use std::shuffle instead. std::random_shuffle is deprecated: https://en.cppreference.com/w/cpp/algorithm/random_shuffle

src/ompl/base/samplers/src/DeterministicStateSampler.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+        }
+
+        void SO2DeterministicStateSampler::sampleUniform(State *state) {
+            auto sample = sequence_ptr_->sample();
+            state->as<SO2StateSpace::StateType>()->value = -boost::math::constants::pi<double>() + sample[0]*2*boost::math::constants::pi<double>();
+        }
+
+        void SO2DeterministicStateSampler::sampleUniformNear(State *state, const State *near, double distance) {
+            (void) state;
```
mamoll: I'm guessing you're doing this to suppress compiler warnings. Don't do it this way. Either omit the argument names or comment them out C-style:
```
 SO2DeterministicStateSampler::sampleUniformNear(State *, const State *, double) 
```
or
```
 SO2DeterministicStateSampler::sampleUniformNear(State */*state*/, const State */*near*/, double /*distance*/) 
```

src/ompl/base/samplers/DeterministicStateSampler.h
roym899: I'm the original author of this, but not at Bosch anymore, so I am not in the position to address the changes directly. Just wanted to comment on this. I believe the problem with the PrecomputedStateSampler type would be that there is no suitable default constructor that works out of the box (i.e., you need to specify the path for it to make sense).

So for the PrecomputedSequence you'd have to use the sequence_ptr constructor instead. Alternative would be to change the PrecomputedSequence a bit and allow access to sequence_ptr_.

src/ompl/base/samplers/deterministic/PrecomputedSequence.h
roym899: This can be used if you have an offline way to precompute samples, i.e., using some optimizer. E.g., in the Dispertio paper we generated the samples externally using Python and then used them inside of OMPL using this interface.

src/ompl/base/samplers/DeterministicStateSampler.h
palmieri: Hi Mark, can you comment a bit more on this? Or maybe provide an example? Not clear what you meant.

src/ompl/base/samplers/DeterministicStateSampler.h
mamoll: I almost forgot about this, but OMPL already has a class for sampling precomputed samples. See `ompl::base::StateStorage` and `tests/base/state_storage.cpp`. Would it be possible to eliminate the PrecomputedStateSampler class in favor of using [StateStorage::getStateSamplerAllocator](https://ompl.kavrakilab.org/classompl_1_1base_1_1StateStorage.html#a0acdbdeeb523a654afb94b9c2de9b6eb)?

src/ompl/base/samplers/DeterministicStateSampler.h
mamoll: This method need not be pure virtual. It can provide a default implementation to copy the next deterministic sample, stored in a std::vector<double>, to an abstract State*. The function StateSpace::copyFromReals is designed to do exactly this. It knows how to map the different vector values to memory locations inside the State*. This way you don't have to implement this method for every possible state space.

demos/DeterministicSamplingPlanning.cpp
palmieri: This is now implemented as requested.

src/ompl/base/samplers/deterministic/src/HaltonSequence.cpp
palmieri: Implemented

src/ompl/base/samplers/DeterministicStateSampler.h
palmieri: Implemented

demos/DeterministicSampling.cpp
palmieri: Done

src/ompl/base/samplers/deterministic/HaltonSequence.h
palmieri: We thought to keep it general and public in case any other new deterministic sampler could depend on it.

src/ompl/base/samplers/deterministic/src/HaltonSequence.cpp
palmieri: done

src/ompl/base/samplers/deterministic/src/HaltonSequence.cpp
palmieri: done

src/ompl/base/samplers/deterministic/src/PrecomputedSequence.cpp
palmieri: done

src/ompl/base/samplers/src/DeterministicStateSampler.cpp
palmieri: done

src/ompl/base/samplers/DeterministicStateSampler.h
palmieri: Implemented

demos/DeterministicSampling.cpp
palmieri: Implemented and just pushed it.

src/ompl/base/samplers/DeterministicStateSampler.h
palmieri: We would love to keep the PrecomputedSequence class. This would allow us to include easily new functionalities, for instance a method that handles the shuffling of the samples.

tests/resources/haltonXD.h
```diff
@@ -0,0 +1,94 @@
...
+    HaltonXD(unsigned int dim)
+    {
+        dimensions_ = dim;
+    }
+
+    std::vector<std::vector<double>> sequence_;
+    // Dimension of the sequence
+    unsigned int dimensions_;
+
+    HaltonXD &operator=(const HaltonXD &other)
```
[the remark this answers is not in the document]
palmieri: Yes, removed.

tests/resources/haltonXD.h
```diff
@@ -0,0 +1,94 @@
...
+
+#include <fstream>
+#include <iostream>
+#include <iterator>
+#include <vector>
+#include <sstream>
+#include <string>
+
+/** \brief Representation of XD Halton sequence read from file */
+struct HaltonXD
```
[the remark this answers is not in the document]
palmieri: Prefer a class.

tests/base/halton_deterministic_sampling.cpp
```diff
@@ -0,0 +1,167 @@
...
+    for (unsigned int i = 0; i < 5; ++i)
+    {
+        std::vector<double> sample = hs1d.sample();
+        std::vector<double> read_sample = seq[i];
+        BOOST_CHECK_EQUAL(sample.size(), read_sample.size());
+
+        for (unsigned int j = 0; j < sample.size(); j++)
+        {
+            BOOST_CHECK_CLOSE(sample[j], read_sample[j], 0.001);
+            std::cout << sample[j] << " - " << read_sample[j] << ", ";
```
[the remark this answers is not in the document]
palmieri: Done.

PR #696, opened 2020-03-02
Use optimization objective to initialize cost in FMT*

marlinstrub:
Using `std::numeric_limits<double>::infinity()` assumes that we are minimizing the objective. The fix allows to use maximization objectives (e.g., max min clearance http://ompl.kavrakilab.org/optimizationObjectivesTutorial.html).


marlinstrub:
Thanks for maintaining OMPL!

PR #715, opened 2020-05-04
Implementation of Advanced BIT* (ABIT*)

marlinstrub:
Hi

My name is Marlin and I'm currently a PhD student in Jon @gammell 's lab. We've been building on BIT*'s core ideas and one of the algorithms that we developed is Advanced BIT* (ABIT*). In order to avoid code duplication, I've modified BIT*'s implementation such that protected methods can turn it into ABIT*. This allows ABIT*'s implementation to simply inherit from BIT* and configure it appropriately. 

ABIT* is to appear in ICRA 2020 and is already available [on arxiv](https://arxiv.org/abs/2002.06589). A short video explaining the algorithm is available [on youtube](https://youtu.be/VFdihv8Lq2A).

Please let me know if there's anything I can do to make reviewing this PR easier for you. Thank you for maintaining OMPL.

### DIFF
--- a/src/ompl/geometric/planners/bitstar/BITstar.h
+++ b/src/ompl/geometric/planners/bitstar/BITstar.h
@@ -32,7 +32,7 @@
  *  POSSIBILITY OF SUCH DAMAGE.
  *********************************************************************/
 
-/* Authors: Jonathan Gammell */
+/* Authors: Jonathan Gammell, Marlin Strub */
 
 #ifndef OMPL_GEOMETRIC_PLANNERS_BITSTAR_BITSTAR_
 #define OMPL_GEOMETRIC_PLANNERS_BITSTAR_BITSTAR_
--- a/src/ompl/geometric/planners/bitstar/BITstar.h
+++ b/src/ompl/geometric/planners/bitstar/BITstar.h
@@ -315,7 +315,7 @@ namespace ompl
             /** \brief Initialize variables for a new batch. */
             void newBatch();
 
-            /** \brief Prune the problem. Returns true if pruning was done. */
+            /** \brief Prune the problem. */
             void prune();
 
             /** \brief Publish the found solution to the ProblemDefinition. */
--- a/src/ompl/geometric/planners/bitstar/BITstar.h
+++ b/src/ompl/geometric/planners/bitstar/BITstar.h
@@ -199,6 +199,12 @@ namespace ompl
             /** \brief Set the inflation factor for the initial search. */
             void setInitialInflationFactor(double factor);
 
+            /** \brief Set the parameter for the inflation factor update policy. */
+            void setInflationFactorParameter(double parameter);
+
+            /** \brief Set the parameter for the truncation factor update policy. */
+            void setTruncationFactorParameter(double parameter);
+
             /** \brief Get the inflation factor for the initial search. */
             double getInitialInflationFactor() const;
 
@@ -462,6 +468,12 @@ namespace ompl
 
             /** \brief The truncation factor for the search. */
             double truncationFactor_{1.0};
+
+            /** \brief The truncation factor parameter for the update policy. */
+            double truncationFactorParameter_{0.0};
+
+            /** \brief The inflation factor parameter for the update policy. */
+            double inflationFactorParameter_{0.0};
 
             /** \brief The goal vertex of the current best solution. */
             VertexConstPtr curGoalVertex_{nullptr};
--- a/src/ompl/geometric/planners/bitstar/BITstar.h
+++ b/src/ompl/geometric/planners/bitstar/BITstar.h
@@ -196,24 +196,6 @@ namespace ompl
             // Settings.
             // ---
 
-            /** \brief Set the inflation factor for the initial search. */
-            void setInitialInflationFactor(double factor);
-
-            /** \brief Set the parameter for the inflation factor update policy. */
-            void setInflationFactorParameter(double parameter);
-
-            /** \brief Set the parameter for the truncation factor update policy. */
-            void setTruncationFactorParameter(double parameter);
-
-            /** \brief Get the inflation factor for the initial search. */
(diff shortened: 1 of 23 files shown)


src/ompl/geometric/planners/bitstar/BITstar.h
```diff
@@ -375,118 +421,135 @@ namespace ompl
...
             std::shared_ptr<ImplicitGraph> graphPtr_{nullptr};
 
             /** \brief The queue of vertices to expand and edges to process ordered on "f-value", i.e., estimated
              * solution cost. Remaining vertex queue "size" and edge queue size are accessible via
              * vertexQueueSizeProgressProperty and edgeQueueSizeProgressProperty, respectively. */
             std::shared_ptr<SearchQueue> queuePtr_{nullptr};
 
-            /** \brief The goal vertex of the current best solution */
+            /** \brief The inflation factor for the initial search. */
+            double initialInflationFactor_{1.0};
```
mamoll, line 476: I find the distinction between fooFactor_ and fooFactorParameter_ confusing. Can you think of a better naming convention to distinguish them? Also, I guess it's not possible to hide the ABIT* parameters and getter/setter methods from BIT*?

src/ompl/geometric/planners/bitstar/datastructures/CostHelper.h
```diff
@@ -200,23 +212,21 @@ namespace ompl
...
-                return this->motionCost(edgePair.first->stateConst(), edgePair.second->stateConst());
+                return this->motionCost(edgePair.first->state(), edgePair.second->state());
             };
 
-            /** \brief Combine 3 costs */
-            inline ompl::base::Cost combineCosts(const ompl::base::Cost &a, const ompl::base::Cost &b,
-                                                 const ompl::base::Cost &c) const
+            /** \brief Combine multiple costs. */
+            template <typename... Costs>
+            inline ompl::base::Cost combineCosts(const ompl::base::Cost &cost, Costs... costs) const
```
mamoll, line 220: Nice! Shouldn't that be `const Costs... &costs` though?

mamoll:
@marlinstrub I'm trying to get a release of OMPL out soon that will be used for ROS 1 Noetic and the next ROS 2 release. Could you please address the last remaining issues? They seem relatively minor...

src/ompl/geometric/planners/bitstar/BITstar.h
marlinstrub: I agree, this is confusing. I've changed the names and added more documentation in the ABIT* class description. I also made sure ABIT*'s parameters are only declared through `declareParam` when ABIT* is instantiated (is this what you meant?).

src/ompl/geometric/planners/bitstar/datastructures/CostHelper.h
marlinstrub: It definitely should, nice catch!

marlinstrub:
@mamoll Thanks for the feedback! And sorry for the slow response, I'm pushing to create some decent videos for ICRA.

I've pushed the requested changes except for adding BIT*/ABIT* to the planner unit test. I currently get a segfault on BIT* (old and new implementation) when adding it to the test. Fixing this is at the top of my list now, you can expect to hear back from me soon.

May I ask when you plan on getting the new OMPL release out? The reason I ask is that I present a second planner at this year's ICRA and I would really like to merge it upstream as well.

mamoll:
> May I ask when you plan on getting the new OMPL release out? The reason I ask is that I present a second planner at this year's ICRA and I would really like to merge it upstream as well.

I am planning to release this in a couple of weeks. I need to time to review, test, and prepare a release, which means there isn't much time to add yet another planner. Can you create a PR within a week?

marlinstrub:
Yes, I should be able to create the PR within the week. The implementation is done and cleaned up, I just want to do a final review myself before I push.

marlinstrub:
Here is why there was a segfault in the unit tests of BIT* (and ABIT*): The unit test calls a planner's `setup` method after it sets the problem definition in the planner but before it fully sets up the problem (via `setupProblem`). BIT* calls `PlannerInputStates::nextGoal` in its `setup` method, which results in a segfault because `PlannerInputStates::nextGoal` implicitly assumes `pdef_->getGoal() != nullptr`.

I fixed it by making this assumption explicit, i.e., now `PlannerInputStates::nextGoal` returns `nullptr` if the problem definition does not have a valid pointer to a goal (instead of segfaulting). This is in the `Planner` base class, but I don't see how it could possibly have a negative effect on any planner. If you want, I could instead implement the check in BIT*, but then I'd suggest that we'd at least have an `assert(pdef_->getGoal() != nullptr)` in `PlannerInputStates::nextGoal`.

marlinstrub:
There seems to be a problem with ABIT* in the `test_2dcircles_opt_geometric` that the AppVeyor CI catches. The same error occurred on my machine before c6edf74, but with this fix I can't reproduce it on my local machine. Has a similar error occurred for any other planner on AppVeyor? I'd like to fix it, but I'm not sure where to start.

gammell:
https://ci.appveyor.com/project/mamoll/ompl/builds/32917721#L2848
> Testing ABITstar ...
> unknown location(0): fatal error: in "MyPlanTestFixture/geometric_ABITstar": memory access violation occurred at address 0x00000008, while attempting to  read inaccessible data
> C:\projects\ompl\tests\geometric\2d\2dcircles_optimize.cpp(172): last checkpoint
> *** 1 failure is detected in the test module "GeometricPlanningOpt"

Hm... Maybe try running `valgrind` on the `2dcircles_optimize.cpp` and see if it turns up anything on your machine?

marlinstrub:
Will do!

marlinstrub:
The test doesn't produce deterministic results on my machine (despite `RNG::setSeed` being called), but 3d67511 fixes a bug and it seems that was what was causing the issue.

PR #718, opened 2020-05-20
Update paper references for Informed RRT*, SORRT*, and BIT*.

gammell:
Hi @mamoll ,

I was waiting to update the paper references once I had the full publication info for the BIT* IJRR paper. I have that now, so this is just a switch from conference refs to journal refs for Informed RRT*, SORRT*, and BIT*.
(And a link to me that I found when searching for paper refs to update).

Hoping it's no work for you to merge.

Thanks,
Jon


gammell:
P.S., I have just seen @marlinstrub 's finished ICRA videos, so I think you should be getting his second PR shortly.

Thanks as always.

PR #719, opened 2020-05-22
Implementation of Adaptively Informed Trees (AIT*)

marlinstrub:
Hi Mark,

This is the other planner I was telling you about last week. I'm sorry for the delay, I hope it's not too late to get it merged before the new release. Please let me know if there's anything I can do to make merging easier for you.

AIT* is to appear in ICRA 2020. It uses an asymmetric bidirectional search on a random geometric graph and simultaneously estimates and exploits an accurate, problem-specific heuristic. The associated publication is available on [arXiv](https://arxiv.org/abs/2002.06599), and the associated ICRA presentation video is available on [YouTube](https://youtu.be/1h7ugF9F6VM). 

AIT* builds on BIT* as it also searches a random geometric graph (RGG). Unfortunately it can not reuse BIT*'s code as AIT*'s RGG vertices must store information related to AIT*'s forward and reverse search trees, and almost all other types in BIT* and AIT* make assumptions about the type of the vertices.

I'm unsure about which folder I should put AIT* in. It currently lives alongside BIT* and the other planners in `src/ompl/geometric/planners` but maybe it should be a subfolder inside BIT*'s folder? I could move BIT*/ABIT* to `.../planners/bitstar/bitstar` and AIT* to `.../planners/bitstar/aitstar` but that changes the include path of BIT* unless I create symlinks or a forward-include in `.../planners/bitstar`. Please let me know what you prefer.


gammell:
> I'm unsure about which folder I should put AIT* in. It currently lives alongside BIT* and the other planners in `src/ompl/geometric/planners` but maybe it should be a subfolder inside BIT*'s folder? I could move BIT*/ABIT* to `.../planners/bitstar/bitstar` and AIT* to `.../planners/bitstar/aitstar` but that changes the include path of BIT* unless I create symlinks or a forward-include in `.../planners/bitstar`. Please let me know what you prefer.

Some relevant information affecting this question. It's likely there we will have other planners in the same family to contribute in the future. I've been thinking about options while Marlin finished up ICRA and I've come up with the following (where the all caps are hopefully self-explanatory variables, e.g., PLANNER_NAME would currently be ABITstar, AITstar, or BITstar).

If you prefer consolidation but have no preference/desire to worry about the specific details, we're happy to pick the one we think is best. I like 2, as it seems like a nice balance between backwards-compatibility (planner includes don't move), future proofing (the planner's subfolder can have any structure necessary), and cleanliness (everything's stored away in the subfolder).

1 (Everything in subfolders)
```
planners/geometric/bitstar/
PLANNER_NAME/PLANNER_NAME.h
PLANNER_NAME/src/PLANNER_NAME.cpp
PLANNER_NAME/datastructures/CLASSES.h
PLANNER_NAME/datastructures/src/CLASSES.cpp
```

2 (Planner classes at top with self-named subfolders)
```
planners/geometric/bitstar/
PLANNER_NAME.h
src/PLANNER_NAME.cpp
PLANNER_NAME/CLASSES.h
PLANNER_NAME/src/CLASSES.cpp
```

3 (Planner classes at top with subdivided datastructure subfolder)
```
planners/geometric/bitstar/
PLANNER_NAME.h
src/PLANNER_NAME.cpp
datastructures/PLANNER_NAME/CLASSES.h
datastructures/PLANNER_NAME/src/CLASSES.cpp
```

src/ompl/geometric/planners/aitstar/AITstar.h
```diff
@@ -0,0 +1,295 @@
...
+            std::unique_ptr<EdgeQueue> forwardQueue_;
+
+            /** \brief The reverse queue. */
+            using KeyVertexPair = std::pair<std::array<ompl::base::Cost, 2u>, std::shared_ptr<aitstar::Vertex>>;
+            using VertexQueue =
+                ompl::BinaryHeap<KeyVertexPair, std::function<bool(const KeyVertexPair &, const KeyVertexPair &)>>;
+            std::unique_ptr<VertexQueue> reverseQueue_;
+
+            /** \brief The id of the current forward search .*/
+            std::shared_ptr<std::size_t> forwardSearchId_;
```
mamoll, line 245: does this really need to be a shared pointer? Same question for next two data members.

src/ompl/geometric/planners/aitstar/datastructures/Edge.h
```diff
@@ -0,0 +1,97 @@
...
+
+                /** \brief Constructs an edge from a parent, a child, and the sort key. */
+                Edge(const std::shared_ptr<Vertex> &parent, const std::shared_ptr<Vertex> &child,
+                     const std::array<ompl::base::Cost, 3u> &sortKey);
+
+                /** \brief Destructs an edge. */
+                ~Edge() = default;
+
+                /** \brief Returns the parent in this edge. */
+                std::shared_ptr<Vertex> getParent() const;
```
mamoll, line 69: presumably each vertex has only one parent, so maybe change to std::unique_ptr? same comment for every other bit of code that uses shared_ptrs to parents (and children, in cases where there can only be one child).

src/ompl/geometric/planners/aitstar/datastructures/Vertex.h
```diff
@@ -0,0 +1,371 @@
...
+
+                /** \brief Returns whether the vertex has been expanded during the current reverse search. */
+                bool hasBeenExpandedDuringCurrentReverseSearch() const;
+
+                /** \brief Returns whether the vertex has been inserted into the queue during the current reverse
+                 * search. */
+                bool hasBeenInsertedIntoQueueDuringCurrentReverseSearch() const;
+
+                /** \brief Sets the reverse queue pointer of this vertex. */
+                void setReverseQueuePointer(
```
mamoll, line 210: maybe use `using FooHeapElement = ompl::BinaryHeap<....>::Element` somewhere to simplify this and the next method. Same comment applies to long BinaryHeap types elsewhere in this code.

src/ompl/geometric/planners/aitstar/datastructures/Edge.h
marlinstrub: The ownership model I implemented is that the edges share ownership of the vertices with the `ImplicitGraph` (the NN struct in the implicit graph to be specific) and the forward and reverse search trees. An alternative model I thought a lot about was one in which the implicit graph is the sole owner. In this case I think the edges and trees should either hold `std::weak_ptr`s or raw pointers. I think I had `std::weak_ptr`s at some point but changed it to the shared ownership model because the pointers to vertices are indirectly exposed to the public (e.g.  through `AITstar::getEdgesInQueue`) and I don't want the vertices to expire while a user still holds a pointer to them (which could happen when vertices are pruned).
I would suggest keeping the current ownership model of vertices, but if you think a different model is more appropriate I'd be excited to learn about it.

src/ompl/geometric/planners/aitstar/AITstar.h
marlinstrub: They're needed in the `ImplicitGraph` but I think they can be stack allocated and passed by const reference. Thanks for pointing this out.

src/ompl/geometric/planners/aitstar/datastructures/Vertex.h
marlinstrub: Sure, I can use a `using` declaration to help with the readability.

marlinstrub:
Jon and I discussed some more and our preferred option would actually be to rename the `bitstar` folder to `informedtrees`, which would have `ABITstar.h`, `AITstar.h`, and `BITstar.h` in it. But this breaks people's include paths, and so it depends on whether you're okay with this. We think the best alternative is option 2 from Jon's earlier comment. Do you have a preference?

mamoll:
> Jon and I discussed some more and our preferred option would actually be to rename the `bitstar` folder to `informedtrees`, which would have `ABITstar.h`, `AITstar.h`, and `BITstar.h` in it. But this breaks people's include paths, and so it depends on whether you're okay with this. We think the best alternative is option 2 from Jon's earlier comment. Do you have a preference?

Putting all 3 planners under `informedtrees` using the option 2 directory layout sounds good to me.

src/ompl/geometric/planners/aitstar/datastructures/Edge.h
mamoll: Ok, let's leave as-is.

gammell:
Thanks for your input and patience Mark!

marlinstrub:
Thanks for your feedback and for merging this before the next release Mark. I really appreciate it!

mamoll:
@marlinstrub @gammell I ran into two problems (so far):
* AITstar contains a bunch of broken assert statements that cause compilation in debug mode to fail
* AITstar fails one of the assert statements when running the unit tests:
https://github.com/ompl/ompl/blob/master/src/ompl/geometric/planners/informedtrees/src/AITstar.cpp#L783-L786

Could you please fix both ASAP in a separate PR?

mamoll:
> Could you please fix both ASAP in a separate PR?

First batch of fixes is here: https://github.com/ompl/ompl/commit/e00a053b5919dd3f67d8e82f2d2d3247ebec9389

I am curious why there is an option for enabling approximate solutions that is false by default. Why shouldn't the always be on? This what every other planner does in OMPL.

I got the python bindings working, so that I can try things out in the OMPL.app GUI (a good sanity check for me). In the case of AIT*, something is going very wrong. It crashes hard with this stack trace when "find approximate solutions" is set to false:
```
Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   libompl.dylib                 	0x0000000111bf8a07 ompl::geometric::AITstar::trackApproximateSolutions(bool) + 71 (AITstar.cpp:328)
1   libompl.dylib                 	0x0000000111c22b4b void ompl::base::Planner::declareParam<bool, ompl::geometric::AITstar*, void (ompl::geometric::AITstar::*)(bool), bool (ompl::geometric::AITstar::*)() const>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, ompl::geometric::AITstar* const&, void (ompl::geometric::AITstar::* const&)(bool), bool (ompl::geometric::AITstar::* const&)() const, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)::'lambda'(bool)::operator()(bool) const + 123 (Planner.h:385)
2   libompl.dylib                 	0x0000000111c22ab7 decltype(std::__1::forward<bool>(fp)(std::__1::forward<ompl::geometric::AITstar*>(fp0)...)) std::__1::__invoke<void ompl::base::Planner::declareParam<bool, ompl::geometric::AITstar*, void (ompl::geometric::AITstar::*)(bool), bool (ompl::geometric::AITstar::*)() const>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, ompl::geometric::AITstar* const&, void (ompl::geometric::AITstar::* const&)(bool), bool (ompl::geometric::AITstar::* const&)() const, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)::'lambda'(bool)&, bool>(bool&&, ompl::geometric::AITstar*&&...) + 55 (type_traits:4425)
3   libompl.dylib                 	0x0000000111c22a52 void std::__1::__invoke_void_return_wrapper<void>::__call<void ompl::base::Planner::declareParam<bool, ompl::geometric::AITstar*, void (ompl::geometric::AITstar::*)(bool), bool (ompl::geometric::AITstar::*)() const>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, ompl::geometric::AITstar* const&, void (ompl::geometric::AITstar::* const&)(bool), bool (ompl::geometric::AITstar::* const&)() const, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)::'lambda'(bool)&, bool>(bool&&...) + 50 (__functional_base:349)
4   libompl.dylib                 	0x0000000111c22a12 std::__1::__function::__alloc_func<void ompl::base::Planner::declareParam<bool, ompl::geometric::AITstar*, void (ompl::geometric::AITstar::*)(bool), bool (ompl::geometric::AITstar::*)() const>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, ompl::geometric::AITstar* const&, void (ompl::geometric::AITstar::* const&)(bool), bool (ompl::geometric::AITstar::* const&)() const, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)::'lambda'(bool), std::__1::allocator<void ompl::base::Planner::declareParam<bool, ompl::geometric::AITstar*, void (ompl::geometric::AITstar::*)(bool), bool (ompl::geometric::AITstar::*)() const>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, ompl::geometric::AITstar* const&, void (ompl::geometric::AITstar::* const&)(bool), bool (ompl::geometric::AITstar::* const&)() const, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)::'lambda'(bool)>, void (bool)>::operator()(bool&&) + 50 (functional:1533)
5   libompl.dylib                 	0x0000000111c21723 std::__1::__function::__func<void ompl::base::Planner::declareParam<bool, ompl::geometric::AITstar*, void (ompl::geometric::AITstar::*)(bool), bool (ompl::geometric::AITstar::*)() const>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, ompl::geometric::AITstar* const&, void (ompl::geometric::AITstar::* const&)(bool), bool (ompl::geometric::AITstar::* const&)() const, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)::'lambda'(bool), std::__1::allocator<void ompl::base::Planner::declareParam<bool, ompl::geometric::AITstar*, void (ompl::geometric::AITstar::*)(bool), bool (ompl::geometric::AITstar::*)() const>(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, ompl::geometric::AITstar* const&, void (ompl::geometric::AITstar::* const&)(bool), bool (ompl::geometric::AITstar::* const&)() const, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&)::'lambda'(bool)>, void (bool)>::operator()(bool&&) + 51 (functional:1707)
6   libompl.dylib                 	0x00000001118e509d std::__1::__function::__value_func<void (bool)>::operator()(bool&&) const + 77 (functional:1860)
7   libompl.dylib                 	0x00000001118e5044 std::__1::function<void (bool)>::operator()(bool) const + 52 (functional:2419)
8   libompl.dylib                 	0x00000001118e4ab3 ompl::base::SpecificParam<bool>::setValue(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) + 115 (GenericParam.h:163)
``` 
When "find approximate solutions" is set to true, it is still unable to solve a basic problem like the cubicles problem within 30 secs. and no approximate solution is shown. The states and edges can also not be visualized, which suggests something is going wrong with getting the planner data.

marlinstrub:
I must have forgotten to switch back to debug mode for the last two commits. I'm really sorry about that, I know you're busy and I don't want this to be more work for you than it has to be. I'm looking into it now and you will hear back from me shortly.

marlinstrub:
I opened a PR that hopefully fixes the crash when setting tracking approximate solutions to false, and continuously updates approximate solutions when it is set to true (previously only updated before returning from solve).

I'm struggling to install and try out ompl app as I get errors when trying to create the python bindings but I'll report back once I can reproduce the second issue of not finding a solution/not being able to visualize states and edges.

gammell:
> I am curious why there is an option for enabling approximate solutions that is false by default. Why shouldn't the always be on? This what every other planner does in OMPL.

I would guess that this behaviour was inherited from BIT*. Back in the day I was worried about the performance hit of checking/maintaining the approximate solution on every sample, especially as I didn't really understand the use case, so I made it optional.

BIT* actually does default to it off, but I believe when they're off it tells OMPL it doesn't support approximate solutions (via the plannerspec I think it's called?).

marlinstrub:
I couldn't reproduce the issue with OMPL app (I need to sort out the python situation on my system first), but I found a similar issue when trying AIT*/(A)BIT* in MoveIt with a friend. I need to rethink how informed trees handle the situation where `solve` is called when the problem definition doesn't have a goal yet. Informed trees currently call `PlannerInputStates::nextGoal`, which doesn't return until the ptc triggers, and they never actually start planning. That would explain why we can't see any states and edges.

My current plan is to install omplapp on a fresh install of Ubuntu such that I can reproduce the problem locally.

gammell:
> Informed trees currently call `PlannerInputStates::nextGoal`, which doesn't return until the ptc triggers, and they never actually start planning. That would explain why we can't see any states and edges.

Mark can confirm, but I doubt that OMPL app is doing anything as complicated as calling solve before the problem has a goal, like MoveIt! does. Also, if BIT* still works in the app (?) then this specific problem can't be a result of a behaviour they both share.

mamoll:
See my comment on PR #721 for an easy way to reproduce another problem with AIT*.

If tracking approximate solutions causes slowdowns, then they should probably be turned off by default. In many other planners, you get an approximate solution "for free".

PR #721, opened 2020-05-29
Add nullptr check when setting AIT* to track approximate solutions

marlinstrub:
Summary:

1. Add `nullptr` check when calling `AITstar::trackApproximateSolutions`
2. Enable tracking approximate solutions by default
3. Continuously track approximate solutions (previously got best approximate solution only before `solve` returns)

Reasoning:
1. AIT* tries to initialize its approximate solution cost with `OptimizationObjective::infiniteCost`, which needs access to an optimization objective. Calling `AITstar::trackApproximateSolutions` when the planner hasn't been `setup` yet resulted in a segfault because AIT* didn't know of the optimization objective yet. This PR adds the necessary check. (Without an optimization objective, AIT* can't know how to initialize the approximate solution cost and  therefore simply refrains from doing so.)
2. This seems to be the default for other planners.
3. It seems like there is a use case where this is beneficial.

### DIFF
--- a/src/ompl/geometric/planners/informedtrees/AITstar.h
+++ b/src/ompl/geometric/planners/informedtrees/AITstar.h
@@ -272,7 +272,7 @@ namespace ompl
             bool repairReverseSearch_{true};
 
             /** \brief The option that specifies whether to track approximate solutions. */
-            bool trackApproximateSolutions_{false};
+            bool trackApproximateSolutions_{true};
 
             /** \brief The option that specifies whether to prune the graph of useless samples. */
             bool isPruningEnabled_{true};
--- a/src/ompl/geometric/planners/informedtrees/src/AITstar.cpp
+++ b/src/ompl/geometric/planners/informedtrees/src/AITstar.cpp
@@ -74,7 +74,7 @@ namespace ompl
             // Specify AIT*'s planner specs.
             specs_.recognizedGoal = base::GOAL_SAMPLEABLE_REGION;
             specs_.multithreaded = false;
-            specs_.approximateSolutions = trackApproximateSolutions_;  // Disabled by default.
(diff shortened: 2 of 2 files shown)


mamoll:
@marlinstrub this helps with the crashing, but AIT* still is unable to solve anything. To eliminate python bindings as the culprit, I tried a basic rigid body planning demo in OMPL.app. Even with 60 secs. of planning time it doesn't find anything. Even the approximate solution is just the start state.
```
diff --git a/demos/SE3RigidBodyPlanning/SE3RigidBodyPlanning.cpp b/demos/SE3RigidBodyPlanning/SE3RigidBodyPlanning.cpp
index 4f52f12..53b1e70 100644
--- a/demos/SE3RigidBodyPlanning/SE3RigidBodyPlanning.cpp
+++ b/demos/SE3RigidBodyPlanning/SE3RigidBodyPlanning.cpp
@@ -12,6 +12,7 @@
 
 #include <omplapp/apps/SE3RigidBodyPlanning.h>
 #include <omplapp/config.h>
+#include <ompl/geometric/planners/informedtrees/AITstar.h>
 
 using namespace ompl;
 
@@ -46,12 +47,13 @@ int main()
     // setting collision checking resolution to 1% of the space extent
     setup.getSpaceInformation()->setStateValidityCheckingResolution(0.01);
 
+    setup.setPlanner(std::make_shared<geometric::AITstar>(setup.getSpaceInformation()));
     // we call setup just so print() can show more information
     setup.setup();
     setup.print();
 
     // try to solve the problem
-    if (setup.solve(10))
+    if (setup.solve(60))
     {
         // simplify & print the solution
         setup.simplifySolution();
```

marlinstrub:
Thanks! I'm looking into this right now.

PR #722, opened 2020-05-30
Implement k-nearest connection model in AIT*'s RGG

marlinstrub:
Hi Mark,

The problem was that the connection radius of the edge-implicit RGG that AIT* searches was not appropriate for problems in which some dimensions are (much) larger than others (the theory holds for the unit cube). This PR implements an option that turns AIT*'s r-disc RGG into a k-nearest RGG, which seems to be more robust to dimensions of unequal extent. I've set the k-nearest model as the default option.

With this, AIT* to solved the problem referenced in #721 in less than 10 seconds 10/10 times on my machine.


mamoll:
@marlinstrub now the unit tests are failing. Can you fix those?

PR #724, opened 2020-06-01
Make unit test assumption explicit

marlinstrub:
Hi Mark,

As far as I understand, `test_2d_circles_opt_geometric` makes the implicit assumption that the planners find an exact solution to the problem. Otherwise the checks on lines 152, 156, 160, and 173 don't make sense:

- Line 152: `BOOST_CHECK(!opt->isCostBetterThan(prev_cost, new_cost));` If the solution is approximate, this should be allowed happen.
- Line 156: `BOOST_CHECK(!pdef->hasApproximateSolution());` Checks for exact solutions, but the condition of the enclosing `while` prevents this check from being reached when the planner doesn't find an exact solution and uses all the solution time.
- Line 160: `BOOST_CHECK(!opt->isCostBetterThan(ini_cost, prev_cost));` This can happen if the initial solution is approximate.
- Line 173: `BOOST_CHECK(!opt->isCostBetterThan(prev_cost, min_cost));` If `prev_cost` is the cost of an approximate solution, it can be lower that the theoretically lowest cost of an exact solution (which is `min_cost`).

This PR proposes to correct this by making the assumption of an exact solution explicit.

I think AIT* was failing on the Travis CI because there was a situation (seed?) for which AIT* did not find an exact solution to the problem in the given time. We don't expect AIT* to be great relative to other planners for problems with relatively low collision detection resolution and many small obstacles in a large configuration space. We're currently working on an algorithm with improved performance even in these scenarios.

This PR also fixes two bugs in AIT* that I found when figuring out what was going on. One is related to how AIT* handles the `plannerAlwaysTerminatingCondition`, and one how the best approximate solution is updated. I don't think either of these were the cause for the failure of the Travis build.

(Edited for clarity.)

### DIFF
--- a/src/ompl/geometric/planners/informedtrees/AITstar.h
+++ b/src/ompl/geometric/planners/informedtrees/AITstar.h
@@ -245,8 +245,11 @@ namespace ompl
             /** \brief The cost of the incumbent solution. */
             ompl::base::Cost solutionCost_;
 
-            /** \brief The cost to go to the goal of the vertex that is closest to the goal (in cost space). */
+            /** \brief The cost to come to the vertex that is closest to the goal (in cost space). */
             ompl::base::Cost approximateSolutionCost_{};
+
+            /** \brief The cost to go to the goal from the current best approximate solution. */
+            ompl::base::Cost approximateSolutionCostToGoal_{};
 
             /** \brief The increasingly dense sampling-based approximation. */
             aitstar::ImplicitGraph graph_;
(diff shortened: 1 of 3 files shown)


gammell:
I tested this branch on my machine (Ubuntu 18.04) with `valgrind` and it came back completely clean. Code was compiled in Debug mode.

I believe the CI test that failed is unrelated to these changes:
https://travis-ci.org/github/ompl/ompl/jobs/693604470#L438

mamoll:
I'm still getting memory loss errors from valgrind in AIT*, but this seems to have fixed the unit test, both on Travis and locally. Thanks!

gammell:
Huh, that's really weird.
It always takes me a while to interpret this type of valgrind error, and I haven't put the time in to understand these, but could the fact it's different mean the problem is in a linked library? Are the valgrind results coming from one of the virtual CI machines? Is it a base 18.04 machine or something with the latest 18.04.x updates?... Computers are so hard some times.

PR #739, opened 2020-07-12
Prevent RRTConnect returning approximate solutions from goal

ericpairet:
I had come across RRTConnect returning approximate solutions that start at goal state.
As I believe this is not the expected behaviour, this PR prevents this from happening.

### DIFF
--- a/src/ompl/geometric/planners/rrt/src/RRTConnect.cpp
+++ b/src/ompl/geometric/planners/rrt/src/RRTConnect.cpp
@@ -271,8 +271,13 @@ ompl::base::PlannerStatus ompl::geometric::RRTConnect::solve(const base::Planner
             if (gs != REACHED)
                 si_->copyState(rstate, tgi.xstate);
 
-            GrowState gsc = ADVANCED;
             tgi.start = startTree;
+
+            /* if initial progress cannot be done from the otherTree, restore tgi.start */
+            GrowState gsc = growTree(otherTree, tgi, rmotion);
+            if (gsc == TRAPPED)
+                tgi.start = !tgi.start;
+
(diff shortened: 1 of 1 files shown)


simonschmeisser:
This sounds very good, I see this issue a lot ( https://github.com/ros-planning/moveit/issues/1920 ) with some setups and filter solutions where it occurs but actually fixing it sounds better!

mamoll:
I cannot reproduce this behavior with the OMPL.app GUI. @ericpairet is there an easy to way to show that the approximate solution can start at the goal state?

ericpairet:
I could replicate this issue in some of the pre-defined scenarios in [OMPL Web App](http://omplapp.kavrakilab.org/) (see below a couple of examples). In these scenarios, to trigger the case of approximate solutions from the goal state, I had to set the planner's `range` to be small (0.1). However, when I first came across this issue, I did not have to modify the planner's parameters to seldomly obtain such behaviour in scenarios where expanding the tree from the start is more challenging than doing so from the goal.

_Scenario_: **BugTrap_dcar** 
_Robot_: Rigid Body Planning (2D)
_Start_: state inside the bugtrap
_Goal_: state outside the bugtrap
_Planning time_: 0.1 seconds
Screenshot from 2020-11-25 07-35-55
Screenshot from 2020-11-25 07-33-49

_Scenario_: **Twistycool** 
_Robot_: Rigid Body Planning (2D)
_Start_: state on the right-side part of the image
_Goal_: state on the left-side part of the image
_Planning time_: 0.1 seconds
Screenshot from 2020-11-25 07-48-26
Screenshot from 2020-11-25 07-45-43

mamoll (approved):

PR #762, opened 2020-11-18
Fixed missing parens in cmake

scchow:
My build of OMPL from source for use with MoveIt! from the latest master branch failed with error:

```
CMake Error at ~/uuv_ws/devel/share/ompl/cmake/omplConfig.cmake:78:
  Parse error.  Function missing ending ")".  End of file reached.
```

Tracking this down, I believe it is caused by a missing close parenthesis in Line 54
https://github.com/ompl/ompl/blob/552ee37591ddc5c8160b935085339e90dee886fc/omplConfig.cmake.in#L53-L60

I am not entirely certain what Line 54 is doing, as it seems like Line 58 resets the value of `OMPLAPP_LIBRARIES`. 
In any case, adding an extra close parenthesis at the end of Line 54 solves the build issue.


mamoll (approved):

mamoll:
Thanks for the fix!

PR #723, opened 2020-05-31
Fix AIT*'s clear method

marlinstrub:
The k-nearest RGG constant should not be reset because it is not computed when AIT* solves a problem. Sorry, this was a silly mistake.

PR #763, opened 2020-11-23
Fix/aitstar miscellaneous

marlinstrub:
Hi Mark,

I hope you are doing well.

TLDR: This PR

1. fixes a memory-leak in AIT* (hopefully this is the error you referenced in #723);
2. makes AIT* check termination conditions while sampling a new batch of states; and
3. adds AIT* to the optimal planning demo.

Please let me know if you have any questions.

Details:

1. Memory leak: AIT* caches nearest-neighbors results for each sample. It used to do so by storing the nearest-neighbors in a vector of `std::shared_ptr`s, but this could lead to a memory leak when two (or more) neighboring samples were pruned. This PR changes the nearest-neighbor cache to hold `std::weak_ptr`s instead.
2. Check termination condition while sampling: AIT* uses informed sampling to sample states. If the optimization objective does not support a direct sampler and the current solution is close to optimal, sampling informed states can take many attempts. This PR ensures AIT* is better at respecting the termination condition by checking it after every sample.
3. Add AIT* to the optimal planning demo: In trying to make this PR as little work for you as possible, I've ensured AIT* works on the optimal planning demo (`demo_OptimalPlanning`), the SE3 rigid body planning demo of ompl app (`demo_SE3RigidBodyPlanning`), as well as the 2d circles test for optimal planners (`test_2dcircles_opt_geometric`). I figured I could leave AIT* as option in the optimal planning demo.

### DIFF
--- a/demos/OptimalPlanning.cpp
+++ b/demos/OptimalPlanning.cpp
@@ -43,6 +43,7 @@
 #include "ompl/util/Console.h"
 
 // The supported optimal planners, in alphabetical order
+#include <ompl/geometric/planners/informedtrees/AITstar.h>
 #include <ompl/geometric/planners/informedtrees/BITstar.h>
 #include <ompl/geometric/planners/cforest/CForest.h>
 #include <ompl/geometric/planners/fmt/FMT.h>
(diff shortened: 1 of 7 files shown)


mamoll (approved):

mamoll:
Thanks, @marlinstrub!

PR #777, opened 2021-02-25
Add Eigen libraries to link targets for OMPL

marlinstrub:
TLDR:
I think Eigen should be added to the list of libraries the `ompl` target links to.

Details:
I don't fully understand OMPL's CMake structure, but the reason this came up is that I'm modularizing an internal tool that depends on OMPL and one component wont build because the compiler can't find the Eigen includes. This component does not directly depend on Eigen though, so I'd rather not explicitly specify it as a dependency. I'm far from being a CMake expert, but I think adding Eigen to `target_link_libraries` is the right thing to do here because that propagates OMPL's dependency on Eigen to build targets that depend on OMPL.


marlinstrub:
Something doesn't work as I expected, so this might not be necessary or sufficient. I'll investigate and come back to this.

mamoll:
There is no Eigen library; Eigen is header-only. Can I close this PR?

marlinstrub:
You're right. I was confused.

PR #778, opened 2021-02-27
Add state spaces Sphere, Torus and Mobius.

aorthey:
Hello everyone,

to make sure that a planner works correctly, I found it super useful to have some simple geometrical spaces which we can visualize. Here are three which I often use: Sphere, Torus and Mobius strip. They all come with correct uniform sampling methods and correct interpolation/distance metrics attached (see pictures to get an idea). 

I thought someone else might be interested in using them, so I wanted to share them. 

Best,
Andreas

sphere_interpolation_thumbnailsphere_samples_thumbnailtorus_interpolation_thumbnailtorus_samples_thumbnailmobius_interpolation_thumbnailmobius_samples_thumbnail

### DIFF
--- a/src/ompl/base/spaces/src/MobiusStateSpace.cpp
+++ b/src/ompl/base/spaces/src/MobiusStateSpace.cpp
@@ -1,0 +1,112 @@
+#include "ompl/base/spaces/MobiusStateSpace.h"
+#include "ompl/tools/config/MagicConstants.h"
+#include <cstring>
+#include <boost/math/constants/constants.hpp>
+
+using namespace boost::math::double_constants; //pi
+using namespace ompl::base;
+
+MobiusStateSpace::MobiusStateSpace(double intervalMax)
+{
+    setName("Mobius" + getName());
+    type_ = STATE_SPACE_UNKNOWN;
+
+    StateSpacePtr SO2(std::make_shared<SO2StateSpace>());
+    StateSpacePtr R1(std::make_shared<RealVectorStateSpace>(1));
+    R1->as<RealVectorStateSpace>()->setBounds(-intervalMax, +intervalMax);
+
+    addSubspace(SO2, 1.0);
+    addSubspace(R1, 1.0);
+    lock();
+}
+
+double MobiusStateSpace::distance(const State *state1, const State *state2) const
+{
+    double theta1 = state1->as<MobiusStateSpace::StateType>()->getS1();
+    double theta2 = state2->as<MobiusStateSpace::StateType>()->getS1();
+
+    double diff = theta2 - theta1;
+
+    if (fabs(diff) <= pi)
+    {
+        return CompoundStateSpace::distance(state1, state2);
+    }else{
+        //requires interpolation over the gluing strip 
+        const auto *cstate1 = static_cast<const CompoundState *>(state1);
+        const auto *cstate2 = static_cast<const CompoundState *>(state2);
+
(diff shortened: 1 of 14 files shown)


src/ompl/base/spaces/src/MobiusStateSpace.cpp
```diff
@@ -0,0 +1,146 @@
...
+}
+
+State *MobiusStateSpace::allocState() const
+{
+    auto *state = new StateType();
+    allocStateComponents(state);
+    return state;
+}
+
+void MobiusStateSpace::freeState(State *state) const
```
mamoll, line 143: Do you need this method? The method doesn't override anything the base class would do.

src/ompl/base/spaces/src/SphereStateSpace.cpp
```diff
@@ -0,0 +1,132 @@
...
+}
+
+ompl::base::State *ompl::base::SphereStateSpace::allocState() const
+{
+    auto *state = new StateType();
+    allocStateComponents(state);
+    return state;
+}
+
+void ompl::base::SphereStateSpace::freeState(State *state) const
```
mamoll, line 129: Not needed

src/ompl/base/spaces/src/TorusStateSpace.cpp
```diff
@@ -0,0 +1,142 @@
...
+TorusStateSampler::TorusStateSampler(const StateSpace *space) : StateSampler(space)
+{
+}
+
+void TorusStateSampler::sampleUniform(State *state)
+{
+    // https://stackoverflow.com/questions/26300510/generating-random-points-on-a-surface-of-an-n-dimensional-torus
+    // Based on publication "Random selection of points distributed on curved surfaces."
+    // Link: https://iopscience.iop.org/article/10.1088/0031-9155/32/10/009/pdf
+    const TorusStateSpace *T = static_cast<const TorusStateSpace *>(space_);
```
mamoll, line 56:
```suggestion
    const auto *T = static_cast<const TorusStateSpace *>(space_);
```

src/ompl/base/spaces/src/TorusStateSpace.cpp
```diff
@@ -0,0 +1,142 @@
...
+}
+
+State *TorusStateSpace::allocState() const
+{
+    auto *state = new StateType();
+    allocStateComponents(state);
+    return state;
+}
+
+void TorusStateSpace::freeState(State *state) const
```
mamoll, line 129: Not needed

mamoll (changes requested):
Looks good! I have left some minor comments. I hope to get to your older massive PR soon.

aorthey:
Thank you Mark! Not sure why freeState was still in there. Should be good now.

mamoll:
If it's not too much trouble, could you extend the [state space tests](https://github.com/ompl/ompl/blob/main/tests/base/state_spaces.cpp) with some basic tests for your new spaces?

marlinstrub:
Thanks for sharing these @aorthey

aorthey:
Ok I added the state space tests. I found two small bugs (Mobius used same header define as Torus, and the sphere distance function used a wrong angle). Now it looks good and tests run through.

I also added a function "toVector" which projects the parameterization of the spaces into 3D space (which would be useful if anyone wants to actually visualize it).

mamoll (approved):

PR #780, opened 2021-03-01
make PathSimplifier aware of the objective in SimpleSetup

imgeorgiev:
The `SimpleSetup` class is absolutely awesome but it currently has a limitation where even if you set a custom objective via

```cpp
simple_setup.setOptimizationObjective(my_objective);
```

it doesn't make the `PathSimplifier` aware of that objective and it simply defaults to optimizing for the shortest path length. This can be seen in the plot below

image (2)

In it, I have set a custom objective which includes optimizing for the shortest path and a potential field that should force the path to the wall on the right side. The solution states in purple are correctly optimized and sit on the right but after I run path simplification I get the path in blue which disregards my custom objective.

With the changes in this MR, running the same experiment results in 

image (3)

(the transparent line is the reference towards which the path is pulled)


mamoll (approved):
Looks good to me. Thanks!

PR #784, opened 2021-03-04
Clean up BIT* planner progress properties

### LINKED ISSUE #779: BIT* is failing when added to a benchmark
Hello @mamoll ,

I am doing some study for bench-marking of optimizing planners. 
I am getting random behavior from BIT* planner. It sometimes crashes with Seg fault. 
I couldn't understand much as I don't know about implementation details.

Please refer to gdb back trace ;
Screenshot from 2021-03-01 11-19-13

I am using source build of master at  https://github.com/ompl/ompl/commit/25c76431cef25f0100ed74d09dd88944ecca5ee1. 

Please do let me know if more info is required. In order to reproduce this adding a BIT* to bench marking should suffice but as I said I get this pretty random, so I don't know if it as actual issue.

marlinstrub:
The main purpose of this PR is to remove the planner progress properties in BIT* that are not thread-safe (namely getting number of current vertices in graph and getting number of total samples). This should fix #779, @jediofgever can you please confirm?

I also changed the default name of BIT* from `"BITstar"` to `"kBITstar"` since by default BIT* is instantiated with a k-nearest connection model (since c08da89), which silences a warning. I also took the chance to clean up some code as it came up.

### DIFF
--- a/src/ompl/geometric/planners/informedtrees/BITstar.h
+++ b/src/ompl/geometric/planners/informedtrees/BITstar.h
@@ -476,8 +476,7 @@ namespace ompl
             std::shared_ptr<ImplicitGraph> graphPtr_{nullptr};
 
             /** \brief The queue of vertices to expand and edges to process ordered on "f-value", i.e., estimated
-             * solution cost. Remaining vertex queue "size" and edge queue size are accessible via
-             * vertexQueueSizeProgressProperty and edgeQueueSizeProgressProperty, respectively. */
+             * solution cost. */
             std::shared_ptr<SearchQueue> queuePtr_{nullptr};
 
             /** \brief The inflation factor for the initial search. */
--- a/src/ompl/geometric/planners/informedtrees/BITstar.h
+++ b/src/ompl/geometric/planners/informedtrees/BITstar.h
@@ -149,7 +149,7 @@ namespace ompl
             using NameFunc = std::function<std::string()>;
 
             /** \brief Construct with a pointer to the space information and an optional name. */
-            BITstar(const base::SpaceInformationPtr &spaceInfo, const std::string &name = "BITstar");
+            BITstar(const base::SpaceInformationPtr &spaceInfo, const std::string &name = "kBITstar");
 
             /** \brief Destruct using the default destructor. */
             virtual ~BITstar() override = default;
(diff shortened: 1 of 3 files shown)


marlinstrub:
@jediofgever Oh could you please do me a favour and also test AIT* (`informedtrees/AITstar`) with your setup?

jediofgever:
@marlinstrub ,no worries about AITstar, I have included that too in my setup and it does not behave the way BITstar does(did).
 We are working on evaluation of optimizing planners for outdoor robot planning,  all major implementations of optimizing planners in OMPL are covered. We hope to publish our findings in upcoming months.

marlinstrub:
@jediofgever Great, thanks!

jediofgever:
@marlinstrub , I confirm that without setting prune to false, the changes https://github.com/robotic-esp/ompl/tree/fix/bitstar-planner-progress-properties adress seg fault. I don't see that behavior anymore.

gammell:
Thank you @jediofgever, we really appreciate you finding this and verifying the fix. Please don't hesitate to reach out directly to @marlinstrub or myself if anything else comes up in your evaluation.

src/ompl/geometric/planners/informedtrees/BITstar.h
```diff
@@ -149,7 +149,7 @@ namespace ompl
             using NameFunc = std::function<std::string()>;
 
             /** \brief Construct with a pointer to the space information and an optional name. */
-            BITstar(const base::SpaceInformationPtr &spaceInfo, const std::string &name = "BITstar");
+            BITstar(const base::SpaceInformationPtr &spaceInfo, const std::string &name = "kBITstar");
```
mamoll, line 152: was this intentional?

marlinstrub: Yes. Since c08da89 the k-nearest connectivity model is the default in BIT*, but in BIT*'s `setup` method we have the following check, which produces a warning and changes the name to `"kBITstar"` if the k-nearest model is active.
https://github.com/ompl/ompl/blob/25c76431cef25f0100ed74d09dd88944ecca5ee1/src/ompl/geometric/planners/informedtrees/src/BITstar.cpp#L240-L246
I don't know if the above check is really necessary, but I don't think its great to have the default settings result in a warning.

marlinstrub:
Yep, thanks for reporting and confirming that this fixes your problem, @jediofgever.

mamoll (approved):

PR #789, opened 2021-03-11
Fixed mismatched tags in package.xml

playertr:
The xml tags were mismatched, resulting in an error when I recently attempted to install OMPL.

I changed the close tags to be the same as the open tags and then I was able to compile OMPL.


mamoll:
didn't see the PR in time (I saw the issue first). I have already fixed this. thanks for letting me know!

PR #797, opened 2021-04-13
Update propagater of RigidBodyPlanningWithIntegrationAndControls.cpp

rmbsmit:
Propagator in demo is incorrect.

### DIFF
--- a/demos/RigidBodyPlanningWithIntegrationAndControls.cpp
+++ b/demos/RigidBodyPlanningWithIntegrationAndControls.cpp
@@ -101,19 +101,19 @@ public:
 
     void propagate(const ob::State *start, const oc::Control *control, const double duration, ob::State *result) const
     {
-        double t = timeStep_;
+        double t = 0;
         std::valarray<double> dstate;
         space_->copyState(result, start);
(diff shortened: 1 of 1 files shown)


mamoll:
The OMPL version and your version appear equivalent. Can you explain the bug (perhaps with an example)?

rmbsmit:
@mamoll 
Lets say _duration=12_, and _timestep=5_. The desired behavior is two update steps of 5 seconds, and one of 2 seconds. In the current implementation: 

1. _t_ is set to 5 (line 104)
2. First update of 5 seconds. (line 109&110)
3. _t_ is set to 10 (line 111)
4. Second update of 5 seconds (line 109&110)
5. _t_ is set to 15 (line 111)

At this point, the **while**-loop is invalid and the **if**-statement is called. At this point, _t_ = 15 and 2x5=10 seconds of propagation is done. 

6. If loop is entered (_t_=15 > duration=12)
7. Third update for (15-12=3) seconds (line 116)

This results in a total propagation of 13 seconds, instead of 12 seconds.

mamoll (approved):

PR #805, opened 2021-05-13
Print upper and lower bound if bounds check fails

5yler:


src/ompl/base/spaces/src/RealVectorBounds.cpp
```diff
@@ -89,9 +90,32 @@ void ompl::base::RealVectorBounds::resize(std::size_t size)
...
+        {
+            std::cout << low[i] << " ";
+        }
+        std::cout << std::endl << "high: ";
+        for (unsigned int i = 0; i < high.size(); ++i)
+        {
+            std::cout << high[i] << " ";
+        }
+        std::cout << std::endl;
+        throw Exception("Bounds for real vector space seem to be incorrect (lower bound must be stricly less than "
```
mamoll, line 118: ... lower bound must be less than **or equal to** upper bound

mamoll (commented):
Looks good to me. I have just one small rewording suggestion for an error message.

PR #813, opened 2021-06-04
Refactor AIT*

marlinstrub:
This PR refactors AIT* and implements a slight improvement regarding the stopping condition of AIT*'s reverse search. The forward search and the resulting solutions are equivalent to before but may be discovered slightly faster.


gammell (approved):

gammell:
These changes to  AIT* include changes to internal APIs, is that constant from an ABI perspective?

mamoll:
That is my understanding: if the public API changes, we should increase the ABI version number (upon release).

PR #820, opened 2021-07-09
override the validSegmentCount function of the ReedsSheppState Space

ZhangSongyi:
the inherited validSegmentCount is calculated by 2D distances between states, which causes the undersampling issue to the ReedsShepp curves connecting two states near in 2D positions.


zkingston (approved):
LGTM. Thank you for the contribution.

PR #827, opened 2021-08-17
Small fixes to make lightning and thunder work with MoveIt and robowflex

ChamzasKonstantinos:
These are some minor changes needed to make Lightning and Thunder work with MoveIt and [Robowflex](https://kavrakilab.github.io/robowflex/), Some changes are also required from the Move-It side, I will probably make a PR there soon.

### DIFF
--- a/src/ompl/tools/lightning/src/Lightning.cpp
+++ b/src/ompl/tools/lightning/src/Lightning.cpp
@@ -129,7 +129,6 @@ void ompl::tools::Lightning::setup()
             else
             {
                 experienceDB_->load(filePath_);  // load from file
-                return; 
             }
         }
         else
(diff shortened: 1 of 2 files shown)


src/ompl/tools/lightning/src/Lightning.cpp
```diff
@@ -124,10 +124,12 @@ void ompl::tools::Lightning::setup()
             if (filePath_.empty())
             {
                 OMPL_ERROR("No file path has been specified, unable to load experience DB");
+                return; 
             }
             else
             {
                 experienceDB_->load(filePath_);  // load from file
+                return;
```
zkingston, line 132: Shouldn't the function drop out of this conditional and set the `configured_` flag here? It appears to successfully load.

src/ompl/tools/thunder/src/Thunder.cpp
```diff
@@ -70,6 +70,12 @@ void ompl::tools::Thunder::initialize()
 
 void ompl::tools::Thunder::setup()
 {
+    if (!filePath_.compare("unloaded"))
+    {
+        OMPL_WARN("Database was not set setup has to be called with a database different than %s", filePath_.c_str());
```
zkingston, line 75: Should be "Database filepath has not been set. Unable to setup.". You can also just use the `==` operator instead of compare.

zkingston (approved):
LGTM

PR #853, opened 2021-12-07
Optimization Objectives: Add support for control-space planners

whoenig:
OptimizationObjective was designed for geometric planning and cannot
associate costs with ompl::control::Control. This addresses this short-
coming in the following way:

* New controlCost(control, steps) function (defaults to identityCost)
* New ControlDurationObjective, which minimizes the total time needed
* Support for controlCost() in the control-space version of SST

Other typical control-space costs include control-integral costs, but
are application-specific and can be now defined by the user by creating
a derived class inheriting from OptimizationObjective().


zkingston (approved):
Looks good, thanks for the contribution!

PR #854, opened 2021-12-07
SST: Add support for IntermediateSolutionCallback of problem definition

whoenig:
SST currently just prints an information when a new solution is found.
This change calls the (optional) callback that a user can set to get
informed about new solutions, similar to BIT*, RRT* etc.


zkingston (approved):
LGTM. Thank you!

PR #863, opened 2021-12-31
Ensure RRT-Connect alternates expansion of trees

marlinstrub:
Hi Mark

A small change for RRT-Connect:

I think it's important that planners can be visualized iteration by iteration. One way to achieve this is to repeatedly call solve with an always-terminating ptc and get the planner data after each call. This small change ensures that RRT-Connect alternates between expanding the start and goal trees in this scenario.

cc: @gammell

### DIFF
--- a/src/ompl/geometric/planners/rrt/RRTConnect.h
+++ b/src/ompl/geometric/planners/rrt/RRTConnect.h
@@ -177,6 +177,9 @@ namespace ompl
             /** \brief The goal tree */
             TreeData tGoal_;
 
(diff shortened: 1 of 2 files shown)


gammell (approved):

gammell:
@mamoll I am sure I have asked you this before, but if all checks have passed, am I clear to just squash and merge without your input?

zkingston:
@gammell go ahead and squash & merge if the checks have passed! Thanks for helping out.

Issue #872, opened 2022-01-18
cpack version not correctly set


BeatScherrer:
Steps to reproduce:
when generating a package with cpack the version of the generated package is not entirely correct: The name shows the correct version number while when installing the reported version is `0.1.1` which indicates the version is not set at all.

I'm opening a pull request with the corresponding fix. If you want to configure it another way (which certainly is possible) just propose the solution.

PR #873, opened 2022-01-18
Set cpack version number

BeatScherrer:
https://github.com/ompl/ompl/issues/872


BeatScherrer:
Also when installing and trying to use the library with regular `find_package(...)` the set variables seem off: 
```
CMake Error at /usr/local/share/ompl/cmake/omplConfig.cmake:24 (message):
  File or directory /usr/local/include/ompl-1.5 referenced by variable
  OMPL_INCLUDE_DIR does not exist !
Call Stack (most recent call first):
  /usr/local/share/ompl/cmake/omplConfig.cmake:35 (set_and_check)
  CMakeLists.txt:334 (FIND_PACKAGE)
```
the headers are installed at `/usr/include/` and not `/usr/local/include`

BeatScherrer:
Not sure why the state space test fails in CI after changing a cmake variable

mamoll (approved):

PR #878, opened 2022-01-31
fix compiler warnings and python bindings for multilevel code

mamoll:


aorthey (approved):
Thanks for the fix. Looks good to me!

PR #879, opened 2022-01-31
fix spacetime planning compiler warnings

mamoll:


demos/SpaceTimePlanning.cpp
```diff
@@ -95,10 +95,10 @@ class SpaceTimeMotionValidator : public ob::MotionValidator {
         return true;
     }
 
-    bool checkMotion(const ompl::base::State *s1, const ompl::base::State *s2,
-                     std::pair<ob::State *, double> &lastValid) const override
+    bool checkMotion(const ompl::base::State *, const ompl::base::State *,
+                     std::pair<ob::State *, double> &) const override
     {
-        return checkMotion(s1, s2);
```
mamoll, line 101: The current demo doesn't do the right thing with the lastValid argument, so it's better to throw an exception rather than potentially have undefined behavior upstream.

aorthey (approved):
Looks good to me!

PR #881, opened 2022-01-31
add citation cff

mamoll:
Per @zkingston's suggestion, I added a citation file. It doesn't seem to do anything with the other OMPL-related citations.


zkingston (approved):
LGTM - works for the default citation, and maybe they will add more features that use this file later.

Issue #883, opened 2022-03-02
Conditionally use ament_cmake to register package for colcon workspaces environment


ruffsl:
Despite https://github.com/ompl/ompl/pull/754 , it seems that ompl still doesn't completely register as an installed package when building it from source in an isolated underlay colcon workspace. Although the current cmake does register ompl via a marker file in the ament package index, it does not populate the necessary hooks/scripts used to update the `AMENT_PREFIX_PATH` environment used to find the ament package index. This path environment is normally updated when sourcing the setup scripts of a colcon workspace, and is particularly relevant when the workspace is built as isolated (colcon default).

For example, when building and sourcing ompl from an underlay workspace, tools such as rosdep for installing overlay dependencies will still fail to find ompl due to the exclusion of it's respective ament package index folder from `AMENT_PREFIX_PATH`. This is what I just recently encountered when building ompl from source to help bootstrap Nav2's CI when migrating ROS2 Rolling from Ubuntu Focal to Jammy. 

Rather than continuing to reverse engineer a cmake workaround to replicate the ament logic, perhaps we could conditionally use ament if ament_cmake is found to register the package directly? This would be reminiscent of other ROS agnostic CMake projects used in the ROS ecosystem, such as BehaviorTree.CPP:

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/d0d1177ce1b0b4a4d0aff647c7af867acd9417d1/package.xml#L30

https://github.com/BehaviorTree/BehaviorTree.CPP/blob/392f8a7e9bb1163991d7275ac6d6ebc835b76661/CMakeLists.txt#L120-L127

cc @SteveMacenski @tylerjw , as I assume MoveIt2 would like to quickly migration with Rolling to Ubuntu Jammy as well.

ruffsl:
@mamoll , I'm not entirely familiar with all the cmake build_type features and configure install scripts being used here at present.
Any insites?

PR #884, opened 2022-03-02
Conditionally use ament_cmake to register package for colcon workspaces environment

ruffsl:
Fixes https://github.com/ompl/ompl/issues/883 by mimicking the approach  from BehaviorTree.CPP to conditionally use ament_cmake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -245,6 +245,7 @@ endif()
 
 find_package(ament_cmake QUIET)
 if(ament_cmake_FOUND)
+    ament_export_include_directories(include)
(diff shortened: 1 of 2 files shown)


mamoll:
Is it perhaps this code that needs to change for downstream packages?: https://github.com/ompl/ompl/blob/main/CMakeLists.txt#L202-L217
I believe this is the fairly "standard" way to install the cmake config files that let cmake find ompl in downstream packages.

Ryanf55:
The recent CMake changes to support target-based properties on ompl should make this easier.

PR #906, opened 2022-05-26
create a copy of captured TerminationCondition to avoid dangling refe…

### LINKED ISSUE #830: Wrong handling of derived PlannerTerminationConditions
When using PlannerTerminationCondition in the following context:
```cpp
auto c = ompl::base::plannerOrTerminationCondition(
  ob::CostConvergenceTerminationCondition(...),
  ob::IterationTerminationCondition(...)
);
```
the resulting condition `c` is broken: To be passed into [`plannerOrTerminationCondition()`](https://github.com/ompl/ompl/blob/f255ddd60c0ed278b7d7df2b3e37de5e872bac23/src/ompl/base/src/PlannerTerminationCondition.cpp#L197-L204), the _temporary_ sub conditions are auto-converted to `PlannerTerminationConditions` via the defined type-conversion operators, e.g.:
https://github.com/ompl/ompl/blob/f255ddd60c0ed278b7d7df2b3e37de5e872bac23/src/ompl/base/terminationconditions/src/IterationTerminationCondition.cpp#L56-L59

However, the `this` pointer becomes dangling as soon as the temporaries are destroyed, causing [illegal memory access](https://github.com/ros-planning/moveit/pull/2836).

To fix the issue, the type-conversion operator needs to store an instance of the actual condition by capturing a copy:
```cpp
ompl::base::IterationTerminationCondition::operator ompl::base::PlannerTerminationCondition() 
{ 
    auto &c = *this;
    return PlannerTerminationCondition([c] { return c.eval(); }); 
}
``` 

Also, you should explicitly suppress deriving from `PlannerTerminationCondition` but enforce implementing the type-conversion operator instead. Otherwise, a derived condition is passed by reference into a `PlannerTerminationCondition`, which is subsequently copied by value - as the _base_ class, thus missing all members of the derived instance.

simonschmeisser:
…rence

Is this what you had in mind @rhaschke in #830 ?  Fixes #830

### DIFF
--- a/src/ompl/base/terminationconditions/src/IterationTerminationCondition.cpp
+++ b/src/ompl/base/terminationconditions/src/IterationTerminationCondition.cpp
@@ -55,5 +55,6 @@ void ompl::base::IterationTerminationCondition::reset()
 
 ompl::base::IterationTerminationCondition::operator ompl::base::PlannerTerminationCondition()
 {
-    return PlannerTerminationCondition([this] { return eval(); });
+    auto &c = *this;
+    return PlannerTerminationCondition([c]() mutable { return c.eval(); });
 }


v4hn:
ping @mamoll

zkingston (approved):
LGTM, thanks for the fix.

Issue #902, opened 2022-05-02
memory access violation when using informed trees planner (on windows, in debug mode)


christian-scheurer:
I am facing memory access violations when using informed trees planner (BITstar, ABITstar and AITstar) in debug mode only.

memory access violation: iterator is null
in ompl\geometric\planners\informedtrees\bitstar\src\Vertex.cpp, line 716

System description:
ompl 1.5.2
Windows 10
compiled with msvc 14.2 toolset with Visual Studio 2019

I adapted the 2denvs.cpp test case so that these planner can be tested.
See here
[2denvs.zip](https://github.com/ompl/ompl/files/8603623/2denvs.zip)

stack trace: 	test_2denvs_geometric.exe!std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<ompl::BinaryHeap<std::pair<std::array<ompl::base::Cost,3>,std::pair<std::shared_ptr<ompl::geometric::BITstar::Vertex>,std::shared_ptr<ompl::geometric::BITstar::Vertex>>>,std::function<bool __cdecl(std::pair<std::array<ompl::base::Cost,3>,std::pair<std::shared_ptr<ompl::geometric::BITstar::Vertex>,std::shared_ptr<ompl::geometric::BITstar::Vertex>>> const &,std::pair<std::array<ompl::base::Cost,3>,std::pair<std::shared_ptr<ompl::geometric::BITstar::Vertex>,std::shared_ptr<ompl::geometric::BITstar::Vertex>>> const &)>>::Element *>>>::operator++() Line 75	C++
 	test_2denvs_geometric.exe!std::_Vector_iterator<std::_Vector_val<std::_Simple_types<ompl::BinaryHeap<std::pair<std::array<ompl::base::Cost,3>,std::pair<std::shared_ptr<ompl::geometric::BITstar::Vertex>,std::shared_ptr<ompl::geometric::BITstar::Vertex>>>,std::function<bool __cdecl(std::pair<std::array<ompl::base::Cost,3>,std::pair<std::shared_ptr<ompl::geometric::BITstar::Vertex>,std::shared_ptr<ompl::geometric::BITstar::Vertex>>> const &,std::pair<std::array<ompl::base::Cost,3>,std::pair<std::shared_ptr<ompl::geometric::BITstar::Vertex>,std::shared_ptr<ompl::geometric::BITstar::Vertex>>> const &)>>::Element *>>>::operator++() Line 285	C++
[...]

PR #908, opened 2022-06-02
fix encounter of possibly null iterator in for loops

gabq1:
This is a bug fix concerning issue #902.

The iterator can become null and so even if the found flag is set to true, it crashes in the next pass before exiting the loop...


zkingston (approved):
LGTM, thanks for the fix.

PR #923, opened 2022-07-27
Fix #include for Eigen headers in demos

lpea:
This is a minor fix that makes inclusion of Eigen headers more consistent across the code base.


mamoll (approved):
:+1:

PR #940, opened 2022-09-14
fix python bindings

mamoll:
Yet another attempt to keep the Python bindings alive for a bit longer. Works on macOS 12.5, clang14 & boost 1.80. Also tested with Dockerfiles in next PR.


zkingston (approved):
Tested locally, works! Thanks Mark.

Issue #907, opened 2022-05-27
How to use "modern" PlannerTerminationCoditions?


simonschmeisser:
I am experimenting around a bit with how the results obtained from OMPL could be improved. Currently we just use RRTConnect as it has a very high rate of finding a solution within a tolerable time. However the quality of solutions can be quite bad occasionally when it flips the robot to some different configuration. Those cases will then also fail to be simplified significantly.

I would like to have something similar to STOMP's "iterations after collision free". Ie once a solution has been found, invest some finite amount of future iterations in hope of better alternatives and then return the result.

The various PlannerTerminationConditions sound promising but unfortunately I could not get any of the optimizing planners to obey to them. I think it was @henningkayser implementing them, which planner did you use? I looked a bit at RRT* and it seems that it would need to call `addSolutionPath` whenever it calls `intermediateSolutionCallback` for the termination conditions to work? Also it seems that the OptimizationObjective is used in a similar manner here?

henningkayser:
just saw this now... I implemented part of it, @mamoll has completed the work and got it merged. If I recall correctly, we have used this with PRM* and AnytimePathShortening successfully. If you look at the CostConvergenceTerminationCondition, you'll see that the callback is being [registered on construction](https://github.com/ompl/ompl/blob/main/src/ompl/base/terminationconditions/src/CostConvergenceTerminationCondition.cpp#L46), and the condition itself doesn't have to depend on the solution path. If `addSolutionPath()` is missing, I'd expect the planner to still terminate while it may not provide the last intermediate solution.

DanyResasco:
Hi,
I have the same problem. I can trigger the processNewSolution(). Can you provide an example on how to use CostConvergenceTerminationCondition with prm* ? 

I was trying to use it with 
```
auto convergenceCondition = aidtr_ompl::CostConvergenceTerminationCondition(pdef, 1, 0.1);
 auto terminationCondition = ompl::base::plannerOrTerminationCondition(timeElapsedCondition,
    convergenceCondition); 
ompl::base::PlannerStatus solved = m_omplSimpleSetup->solve(terminationCondition);
```
    
    but the callback is never called even if I use windows = 1

DanyResasco:
I think is missing something like this 

```
                    std::vector<const ompl::base::State *> dummy;
                    pdef_->getIntermediateSolutionCallback()(nullptr,dummy,  pathCost);
```
in maybeConstructSolution()

simonschmeisser:
[...]

PR #961, opened 2022-12-11
avoid dangling reference to CostConvergence parameters

simonschmeisser:
Currently the captured `this` pointer becomes a dangling reference as soon the original `CostConvergenceTerminationCondition` goes out of scope. This lead to a `solutionsWindow` of 0 which made this quite useless, current average being always `inf`.

Relates to #907 

@rhaschke tagging you due to initiating #906

### DIFF
--- a/src/ompl/base/terminationconditions/src/CostConvergenceTerminationCondition.cpp
+++ b/src/ompl/base/terminationconditions/src/CostConvergenceTerminationCondition.cpp
@@ -43,9 +43,10 @@ ompl::base::CostConvergenceTerminationCondition::CostConvergenceTerminationCondi
       solutionsWindow_(solutionsWindow),
       epsilon_(epsilon)
 {
+    auto& c = *this;
     pdef_->setIntermediateSolutionCallback(
-	    [this](const Planner* /*planner*/, const std::vector<const State*>& /*states*/, const Cost cost) {
-	        this->processNewSolution(cost);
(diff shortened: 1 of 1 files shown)


rhaschke (commented):
> Currently the captured `this` pointer becomes a dangling reference as soon the original `CostConvergenceTerminationCondition` goes out of scope.

Are you sure? Shouldn't a dangling pointer trigger a segfault (or at least `asan` error)?
Your change seems reasonable to keep the termination condition alive (or better a copy of it).
However, I'm not sure the copy operation will actually create a copy of the derived class. Did you check that?
Usually, copying a base class, just calls the base class copy-constructor, which would strip off all interesting stuff from derived classes (https://stackoverflow.com/questions/274626/what-is-object-slicing).

simonschmeisser:
My understanding is that the original code creates a copy of the base class when used in `plannerOrTerminationCondition` (https://github.com/ompl/ompl/blob/d8375d842a7f016d64f27a81d6f95eaa83fbe595/src/ompl/base/src/PlannerTerminationCondition.cpp#L201)
The lambda here in the constructor stores a (this) pointer to the original CostConvergence-object in the original code which then goes out of scope and becomes a dangling pointer.

In my experiments `solutionsWindow` was always 0 instead of the expected value but that is afaik just UB coincidence.

rhaschke:
You are right. Should work then (and follows the same scheme as #906).

mamoll (approved):

PR #783, opened 2021-03-03
Add Experience-driven Random Trees (ERT and ERTConnect)

ericpairet:
This PR contains the Experience-driven Random Trees presented in our recent work on experience-based path planning.

Publication (https://arxiv.org/abs/2103.00448):
Èric Pairet, Constantinos Chamzas, Yvan Petillot, Lydia Kavraki, Path Planning for Manipulation using Experience-driven Random Trees, IEEE Robotics and Automation Letters (RA-L) - IEEE International Conference on Robotics and Automation (ICRA), 2021.

PR #962, opened 2022-12-14
test: add test for experience based planner

HiroIshida:
## what's this
This PR add test for experience base planner.
I think this test is useful for ensuring the core behavior of newly added algorithm is same as the other experience-based planner. 

For example, in my opinion, for https://github.com/ompl/ompl/pull/783 to be merged, adding the test using the same interface with lightning and thunder, though some of the common test could be changed, is expected. 

## Note
1. The test tests both case where algorithm is specified and not. This is because, algorithm's initialization/seutp timing is different for both case.

## Some issue found by adding test
Actually by writing this test, I found behavior mismatches between lightning and thunder, and, the same tests for Thunder will not pass.  I will fix these issues/difference after this PR is merged and this PR is the first step for that.

### difference of when path is stored to the databaes
lightning store the solution to the database right after problem solved in `solve()` function. While in Thunder, it is not automatically stored, rather user explicitly call `doPostProcessing` to move queue to the database. I guess, this implementation is probably because Thunder use more complex database than lightning so if database-storing time is count up in `solve()` time, fair algorithm performance comparison cannot be done. That's why, for the fair comparison, I guess they are intentionally separated. @davetcoleman 

So, in my opinion, Lightning should also have `doPostProcessing` function, and Lightning also should require user to explicitly call this function after `solve()` called.

### segmentation fault when calling `getExperienceCount()`
Thunder initialize its `sparsedb_` only in the end of `solve()` function. Thus, it causes error when `getExperienceCount` called before solving, because at that time `sparsedb_` is `nullptr` at that time.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -53,6 +53,9 @@ if (OMPL_BUILD_TESTS)
     add_ompl_test(test_2dmap_control control/2dmap/2dmap.cpp)
     add_ompl_test(test_planner_data_control control/planner_data.cpp)
 
+    # Test experience based planning
+    add_ompl_test(test_experience_planning tools/test_experience_planning.cpp)
+
     # Test planning via MORSE extension
     if(OMPL_EXTENSION_MORSE)
         add_ompl_test(test_morse_extension extensions/morse/morse_plan.cpp)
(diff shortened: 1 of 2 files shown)


zkingston (approved):
LGTM - thanks for the contribution. Changing the behavior of Lightning to also require an explicit post-process step would also be an acceptable change - this could also then be moved to ExperienceSetup.

PR #965, opened 2022-12-20
fix detection of python module path and version

mamoll:
`inspect.getfile` doesn't seem to work with python3.10 and higher for modules like `PyQt5`:
```
  File "<string>", line 1, in <module>
  File "/opt/local/Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/inspect.py", line 778, in getfile
    raise TypeError('{!r} is a built-in module'.format(object))
TypeError: <module 'PyQt5' (<_frozen_importlib_external._NamespaceLoader object at 0x1100dd660>)> is a built-in module
```


mamoll:
That error is actually caused by some weirdness on my machine unrelated to python version. Nevertheless, this PR still seems like an improvement over the old code.

PR #970, opened 2023-01-18
Fix for RRTConnect with intermediate states

### LINKED ISSUE #712: Potential Bug in RRTConnect (intermediate states)
Hi there! I believe that there is a bug in the `growTree` method of `RRTConnect` where an invalid path is added to the tree in the event that the goal tree is being expanded and `addIntermediateStates_` is set to true.

https://github.com/ompl/ompl/blob/003fc6f29da3b44f9406193fe90905b6d3933f51/src/ompl/geometric/planners/rrt/src/RRTConnect.cpp#L150

The current logic flips the direction of the path being extended based on the current tree. I believe that the current behavior is correct when `tgi.start` is true. However, when `tgi.start` is false, the direction of the motion is flipped: instead of going from the nearest-neighbor in the goal tree (`nmotion->state`) towards the target state (`dstate`), the configurations will be returned in the opposite direction from `dstate` towards `nmotion->state`.

I've drawn a few examples out by hand, and this seems incorrect to me. In the event that `tgi.start` is false, the first motion added to the tree by the loop on line `159` corresponds to `dstate`, which immediately has `nmotion` as its parent. This shortcuts all of the intermediate states between them, which are also added backwards by the same loop.

I think the fix here would actually be not having a special case on line 150, but instead always extending from `nmotion->state` towards the target state `dstate`, which is currently done only if  `tgi.start` is true.

Thanks for your time OMPL dev team!

zkingston:
Fixes #712 - adding states should be reversed when extending goal tree.

### DIFF
--- a/src/ompl/geometric/planners/rrt/src/RRTConnect.cpp
+++ b/src/ompl/geometric/planners/rrt/src/RRTConnect.cpp
@@ -154,21 +154,30 @@ ompl::geometric::RRTConnect::GrowState ompl::geometric::RRTConnect::growTree(Tre
         const unsigned int count = si_->getStateSpace()->validSegmentCount(astate, bstate);
 
         if (si_->getMotionStates(astate, bstate, states, count, true, true))
-            si_->freeState(states[0]);
+        {
+            // if coming from start, don't add the start state (start->goal)
+            if (tgi.start)
+                si_->freeState(states[0]);
+            // if coming from the goal, don't add the start state (goal->start)
+            else
+                si_->freeState(states[states.size() - 1]);
+        }
 
         // Add states forwards if from start, backwards if from goal
-        for (std::size_t i = tgi.start ? 1 : states.size() - 1;  //
-             (tgi.start) ? i < states.size() : i > 0;            //
-             i += 2 * tgi.start - 1)
+        const auto &add_state = [&](const auto &state)
         {
             auto *motion = new Motion;
-            motion->state = states[i];
+            motion->state = state;
             motion->parent = nmotion;
             motion->root = nmotion->root;
             tree->add(motion);
-
             nmotion = motion;
-        }
+        };
+
+        if (tgi.start)
+            std::for_each(++std::begin(states), std::end(states), add_state);
+        else
+            std::for_each(++std::rbegin(states), std::rend(states), add_state);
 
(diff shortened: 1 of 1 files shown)


src/ompl/geometric/planners/rrt/src/RRTConnect.cpp
```diff
@@ -156,7 +156,10 @@ ompl::geometric::RRTConnect::GrowState ompl::geometric::RRTConnect::growTree(Tre
         if (si_->getMotionStates(astate, bstate, states, count, true, true))
             si_->freeState(states[0]);
 
-        for (std::size_t i = 1; i < states.size(); ++i)
+        // Add states forwards if from start, backwards if from goal
+        for (std::size_t i = tgi.start ? 1 : states.size() - 1;  //
+             (tgi.start) ? i < states.size() : i > 0;            //
+             i += 2 * tgi.start - 1)
```
mamoll, line 162: That seems rather obfuscated. Since `i` isn't actually used, wouldn't this be cleaner?:
```
            for (const auto& state : tgi.start ? states : (states | std::views::reverse))
            {
                ...
                motion->state = state;
                ...
            }
```
This requires C++20, but this can probably also be done in C++17 by using range-v3 or boost.

zkingston: I changed up the implementation to use something more intuitive than arcane indexes, not as slick as the suggested but I think will be clear.

src/ompl/geometric/planners/rrt/src/RRTConnect.cpp
```diff
@@ -154,18 +154,30 @@ ompl::geometric::RRTConnect::GrowState ompl::geometric::RRTConnect::growTree(Tre
...
         if (si_->getMotionStates(astate, bstate, states, count, true, true))
-            si_->freeState(states[0]);
+        {
+            // if coming from start, don't add the start state (start->goal)
+            if (tgi.start)
+                si_->freeState(states[0]);
+            // if coming from the goal, don't add the start state (goal->start)
+            else
+                si_->freeState(states[states.size() - 1]);
+        }
```
mamoll, line 164: It seems like this is a separate bug. Nice catch!

mamoll (approved):

PR #980, opened 2023-01-25
Added try/except for string SpecificParam

zkingston:
For some reason, `SpecificParam<std::string>` appears as either `SpecificParam<std::string>` or `SpecificParam<std::basic_string<char>>` - there is some code to try and figure out what strings are in castxml, but for some reason it is inconsistent for `SpecificParam`, particularly on Ubuntu 20.04.

### DIFF
--- a/py-bindings/generate_bindings.py
+++ b/py-bindings/generate_bindings.py
@@ -191,7 +191,10 @@ class ompl_base_generator_t(code_generator_t):
         self.ompl_ns.class_('SpecificParam< float >').rename('SpecificParamFloat')
         self.ompl_ns.class_('SpecificParam< double >').rename('SpecificParamDouble')
         self.ompl_ns.class_('SpecificParam< long double >').rename('SpecificParamLongDouble')
-        self.ompl_ns.class_(f'SpecificParam< {self.string_decl} >').rename('SpecificParamString')
+        try:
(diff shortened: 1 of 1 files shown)


mamoll:
[This](https://github.com/ompl/ompl/blob/main/py-bindings/bindings_generator.py.in#L143-L148) is supposed to correctly guess the "real" type that castxml detects for strings on macOS and Linux. Are you saying this is not working for you? Which version of Linux?

zkingston:
Yeah, I checked the output on Ubuntu 20.04 and it was `std::basic_string<char>`, not `std::string`, but only for `SpecificParam` as far as I could tell.

mamoll:
I can reproduce the problem on Ubuntu 20.04 and on Ubuntu 22.04. This is weird, because I have successfully built the ompl Docker images, which include the Python bindings.

mamoll (approved):

1850298154:
I sovled above problem. But I meet a new problem.
How can I solve it？ thank you！

ubuntu22   anaconda 2023-07
```shell


File /mnt/MPC-MOTION-PLANING/github-share/MPC-MP-zyt-note/IMPC-zOB/ompl/util/__init__.py:4
      2 from ompl import dll_loader
      3 dll_loader('ompl', dirname(abspath(__file__)))
----> 4 from ompl.util._util import *
      5 import inspect
      7 def OMPL_DEBUG(text):

SystemError: initialization of _util raised unreported exception

```

PR #990, opened 2023-02-15
fix: AIT* infinite loop

haryngod:
Executing AIT* with my environment is in an infinite loop.
In a simple environment, it doesn't have any problems, but in a high-complexity environment planner is stuck.

A similar issue was reported in the moveit repo [[moveit#3329](https://github.com/ros-planning/moveit/issues/3329)](https://github.com/ros-planning/moveit/issues/3329).

I added a termination condition to prevent this situation.

### DIFF
--- a/src/ompl/geometric/planners/informedtrees/aitstar/src/ImplicitGraph.cpp
+++ b/src/ompl/geometric/planners/informedtrees/aitstar/src/ImplicitGraph.cpp
@@ -336,7 +336,7 @@ namespace ompl
 
                         // Count how many states we've checked.
                         ++numSampledStates_;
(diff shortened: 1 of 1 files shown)


zkingston (approved):
LGTM, thanks for the fix!

marlinstrub:
The problem with this change is that AIT* can now have invalid samples in its graph.

gammell:
I agree with @marlinstrub s concern. The only way to get stuck in that `while` loop was if the sampler was only returning invalid samples, which is really a problem with the sampling... This feels like changing AIT* to fix a problem elsewhere.

PR #992, opened 2023-02-16
Fix Python versions <= 3.7 finding module path

### LINKED ISSUE #983: New Way of Checking Python Bindings Incompatible with Python < 3.8
For various (annoying) reasons, I have to build OMPL using Python 3.7. A change introduced a month ago breaks the dependency checking when building the Python bindings in 3.7. Here is the change:
https://github.com/ompl/ompl/commit/30f3a5bb90587c133973c42c2b6f26f252e7baea#diff-812c4ae87316942a30217e350bf50f289a827f09b952507df19832d0a9556a1bR128

Note the use of `importlib.metadata`. In Python 3.7, this is an elective package called `importlib_metadata` which must be installed manually. The install script should probably be modified to install this package if necessary and the FindPython.cmake file could use a try/except one-liner to try both packages in order to work with both Python versions. 

I'm happy to make these changes if they seem reasonable. As far as I can tell, the current release (1.6.0) is broken with Python 3.7.

zkingston:
Fixes #983

### DIFF
--- a/CMakeModules/FindPython.cmake
+++ b/CMakeModules/FindPython.cmake
@@ -115,20 +115,35 @@ function(find_python_module module)
         # A module's location is usually a directory, but for binary modules
         # it's a .so file.
         if (_minversion STREQUAL "")
-            execute_process(COMMAND "${PYTHON_EXEC}" "-c"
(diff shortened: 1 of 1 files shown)


mamoll (approved):
LGTM. Sorry for not testing my changes on python<=3.7!

PR #994, opened 2023-02-16
Fix EIT* failing to sample the goal region

yurirocha15:
I am currently trying to use the informed tree planners in MoveIt, however, when testing EIT*/EIRM* the planner is failing to sample the goal region. I get the following error:

```shell
[ERROR] [1676514489.653650246]: arm/arm[EITstar]: No solution can be found as no goal states are available
[ERROR] [1676514489.653739781]: arm/arm[EITstar]: No solution can be found as no goal states are available
[ERROR] [1676514489.653818728]: arm/arm[EITstar]: No solution can be found as no goal states are available
[ERROR] [1676514489.653867118]: arm/arm[EITstar]: No solution can be found as no goal states are available
[ WARN] [1676514489.653914669]: ParallelPlan::solve(): Unable to find solution by any of the threads in 0.000329 seconds
[ WARN] [1676514489.653941393]: Timed out
[ WARN] [1676514489.662968340]: Goal sampling thread never did any work.
[ WARN] [1676514489.806632313]: Fail: ABORTED: TIMED_OUT
```

After analyzing the AIT* code I noticed that [AIT* forwards the termination condition to the goal sampler](https://github.com/ompl/ompl/blob/main/src/ompl/geometric/planners/informedtrees/aitstar/src/ImplicitGraph.cpp#L172), while EIT* create a new `plannerAlwaysTerminatingCondition`. After performing this change, the planner was able to successfully sample the goal region. The terminating condition sent by the parent caller is also a `plannerAlwaysTerminatingCondition`, so I am not sure why sending a new object causes the sampler to fail.

I tested this changes using this code: https://github.com/yurirocha15/testing_informed_planners_example

### DIFF
--- a/src/ompl/geometric/planners/informedtrees/eitstar/src/RandomGeometricGraph.cpp
+++ b/src/ompl/geometric/planners/informedtrees/eitstar/src/RandomGeometricGraph.cpp
@@ -494,7 +494,7 @@ namespace ompl
                 whitelistedStates_.push_back(state);
             }
 
-            std::shared_ptr<State> RandomGeometricGraph::getNewSample()
+            std::shared_ptr<State> RandomGeometricGraph::getNewSample(const ompl::base::PlannerTerminationCondition& terminationCondition)
             {
                 // Allocate a new state.
                 auto state = std::make_shared<State>(spaceInfo_, objective_);
@@ -510,6 +510,11 @@ namespace ompl
                 {
                     do  // Sample randomly until a valid state is found.
                     {
+                        if (!terminationCondition)
+                        {
+                            // We've been asked to stop.
+                            return nullptr;
+                        }
                         if (isMultiqueryEnabled_)
                         {
                             // If we are doing multiquery planning, we sample uniformly, and reject samples that can
@@ -555,11 +560,11 @@ namespace ompl
                 {
                     do
                     {
-                        const auto state = getNewSample();
+                        const auto state = getNewSample(terminationCondition);
 
                         // Since we do not do informed sampling, we need to check if the sample could improve
                         // the current solution.
-                        if (!canBePruned(state))
+                        if (state && !canBePruned(state))
                         {
                             newSamples_.emplace_back(state);
 
--- a/src/ompl/geometric/planners/informedtrees/eitstar/src/RandomGeometricGraph.cpp
+++ b/src/ompl/geometric/planners/informedtrees/eitstar/src/RandomGeometricGraph.cpp
@@ -508,13 +508,9 @@ namespace ompl
                 }
                 else
                 {
+                    bool foundValidSample = false;
                     do  // Sample randomly until a valid state is found.
                     {
-                        if (terminationCondition)
-                        {
-                            // We've been asked to stop.
-                            return nullptr;
-                        }
                         if (isMultiqueryEnabled_)
                         {
                             // If we are doing multiquery planning, we sample uniformly, and reject samples that can
@@ -530,7 +526,15 @@ namespace ompl
                         }
 
                         ++numSampledStates_;
-                    } while (!spaceInfo_->isValid(state->raw()));
+                        // Check if the sample is valid.
+                        foundValidSample = spaceInfo_->isValid(state->raw());
+                    } while (!foundValidSample && !terminationCondition);
+
+                    // The sample is invalid, but we have to return to respect the termination condition.
+                    if (!foundValidSample)
+                    {
+                        return nullptr;
+                    }
 
                     // We've found a valid sample.
                     ++numValidSamples_;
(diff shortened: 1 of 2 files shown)


yurirocha15:
When the environment is complex, the sampling can fail several times before finding a valid sample. Therefore, I also passed the termination condition to the `getNewSample` function, so it can stop trying to get a new sample when the time is up.

src/ompl/geometric/planners/informedtrees/eitstar/src/RandomGeometricGraph.cpp
```diff
@@ -510,6 +510,11 @@ namespace ompl
                 {
                     do  // Sample randomly until a valid state is found.
                     {
+                        if (terminationCondition)
+                        {
+                            // We've been asked to stop.
+                            return nullptr;
+                        }
```
marlinstrub, line 517: We have to be careful here to sample the state before checking the termination condition. If the user passes a condition that always terminates (e.g., to run EIT* iteration by iteration), then this change will prevent EIT* from making any progress. I think we could do something like this:

```cpp
std::shared_ptr<State>
RandomGeometricGraph::getNewSample(const ompl::base::PlannerTerminationCondition &terminationCondition)
{
    // Allocate a new state.
    auto state = std::make_shared<State>(spaceInfo_, objective_);

    // Create the requested number of new states.
    if (currentNumSamples_ < buffer_.size())
    {
        // This does at the moment not deal with the fact that samples might become invalid
        // also, these samples might not be informed samples
        state = buffer_[currentNumSamples_];
    }
    else
    {
        bool foundValidSample = false;
        do  // Sample randomly until a valid state is found.
        {
            if (isMultiqueryEnabled_)
            {
                // If we are doing multiquery planning, we sample uniformly, and reject samples that can
                // not improve the solution. This means that we need to sample the whole space, and add
                // the samples to the buffer
                sampler_->sampleUniform(state->raw(), objective_->infiniteCost());
            }
            else
            {
                // In case we are not doing multiquery planning, we can still directly sample the informed
                // set.
                sampler_->sampleUniform(state->raw(), solutionCost_);
            }

            // Keep track of how many states we sampled.
            ++numSampledStates_;

            // Check if the sample is valid.
            foundValidSample = spaceInfo_->isValid(state->raw());
        } while (!foundValidSample && !terminationCondition);

        // The sample is invalid. We have to return to respect the termination condition.
        if (!foundValidSample)
        {
            return nullptr;
        }

        // We've found a valid sample, let's add it to the buffer.
        ++numValidSamples_;
        buffer_.emplace_back(state);
    }
    ++currentNumSamples_;
    return state;
}
```
Not the most elegant code, but hopefully it fixes the issue you see. Would you mind testing this version with your setup?

marlinstrub (changes requested):
Thanks for raising the issue and offering a fix. Please see the comments for a suggested change.

src/ompl/geometric/planners/informedtrees/eitstar/src/RandomGeometricGraph.cpp
yurirocha15: @marlinstrub 
I added the requested changes in [da49a1c](https://github.com/ompl/ompl/pull/994/commits/da49a1c6550100db6302b01be8359e9be15c263b)

To check if it is working as intended, I created a almost impossible to sample environment and ran the code with and without this PR fixes.
image

### AS IS
- max_planning_time: 5s
```shell
[ WARN] [1676860730.860321117]: ParallelPlan::solve(): Unable to find solution by any of the threads in 7.615564 seconds
[ WARN] [1676860730.860351988]: Timed out
[ WARN] [1676860730.866058622]: Fail: ABORTED: TIMED_OUT
[ WARN] [1676860743.322497691]: ParallelPlan::solve(): Unable to find solution by any of the threads in 8.275939 seconds
[ WARN] [1676860743.322526399]: Timed out
[ WARN] [1676860743.338020238]: Fail: ABORTED: TIMED_OUT
[ WARN] [1676860752.686450662]: ParallelPlan::solve(): Unable to find solution by any of the threads in 7.840804 seconds
[ WARN] [1676860752.686479757]: Timed out
```

### TO BE
- max_planning_time: 5s
```shell
[ WARN] [1676861034.370784561]: ParallelPlan::solve(): Unable to find solution by any of the threads in 5.000077 seconds
[ WARN] [1676861034.370829586]: Timed out
[ WARN] [1676861034.382284010]: Fail: ABORTED: TIMED_OUT
[ WARN] [1676861042.756183926]: ParallelPlan::solve(): Unable to find solution by any of the threads in 5.000095 seconds
[ WARN] [1676861042.756211490]: Timed out
[ WARN] [1676861042.773767959]: Fail: ABORTED: TIMED_OUT
[ WARN] [1676861049.156938698]: ParallelPlan::solve(): Unable to find solution by any of the threads in 5.000158 seconds
[ WARN] [1676861049.156962079]: Timed out
[ WARN] [1676861049.157320775]: Fail: ABORTED: TIMED_OUT
```

marlinstrub (approved):
Thanks for the fixes and for testing them.

PR #995, opened 2023-02-20
Add multi-goal planning to LazyPRM/LazyPRM*

### LINKED ISSUE #776: LazyPRM family does not properly plan against multiple goals
I recently encountered what I think is a bug, but might simply be based on a wrong assumption on my part. I attach the following MRE:
https://gist.github.com/haudren/32fb94dbb3b64dd5eb394eddb74ba62c

Compile with:
```
g++ `pkg-config --cflags ompl` test.cpp `pkg-config --libs ompl`
```

This creates a simple 2D grid, and attempts to plan to both [1, 0] and [0,1] using a `GoalStates`, starting from `[0, 0.7]`. The expected result is that the optimal solution will reach `[0, 1]` as it's much closer than `[1, 0]`. However, the result is that the LazyPRM and LazyPRMstart algorithms seem to always target the first goal.

Indeed, running it with LazyPRMstar results incorrectly in finding a path to the far goal, irrespective of the time given for planning:
```
Debug:   LazyPRMstar: Planner range detected to be 0.282843
Info:    LazyPRMstar: Starting planning with 2 states already in datastructure
Info:    LazyPRMstar: Created 10969 states
Exact solution
Reached:
Geometric path with 2 states
RealVectorState [0 0.7]
RealVectorState [1 0]
```

But the PRMstar algorithm correctly finds the path to the closest goal:
```
Info:    PRMstar: Starting planning with 2 states already in datastructure
Info:    PRMstar: Created 34037 states
Exact solution
Reached:
Geometric path with 2 states
RealVectorState [0 0.7]
RealVectorState [0 1]
```

Am I doing something wrong? Is there a bug in the LazyPRM algorithms? At a cursory glance [this call](https://github.com/ompl/ompl/blob/25c76431cef25f0100ed74d09dd88944ecca5ee1/src/ompl/geometric/planners/prm/src/LazyPRM.cpp#L370) to `solutionComponent` seems to be at fault, as `solutionComponent` will always return the same start & end goal pair. I think that patching it so that every call returns another pair would fix the issue, but I'm really not sure if that's the correct approach.

Thanks a lot for any pointers!

zkingston:
Fixes #776. Before, LazyPRM would simply use the first sampled goal. This should enable LazyPRM to plan to multiple goals.

### DIFF
--- a/src/ompl/geometric/planners/prm/src/LazyPRM.cpp
+++ b/src/ompl/geometric/planners/prm/src/LazyPRM.cpp
@@ -1,36 +1,36 @@
 /*********************************************************************
-* Software License Agreement (BSD License)
-*
-*  Copyright (c) 2013, Willow Garage
-*  All rights reserved.
-*
-*  Redistribution and use in source and binary forms, with or without
-*  modification, are permitted provided that the following conditions
-*  are met:
-*
-*   * Redistributions of source code must retain the above copyright
-*     notice, this list of conditions and the following disclaimer.
-*   * Redistributions in binary form must reproduce the above
-*     copyright notice, this list of conditions and the following
-*     disclaimer in the documentation and/or other materials provided
-*     with the distribution.
-*   * Neither the name of Willow Garage nor the names of its
-*     contributors may be used to endorse or promote products derived
-*     from this software without specific prior written permission.
-*
-*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
-*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
-*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
-*  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
-*  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
-*  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
-*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
-*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
-*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
-*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
-*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
-*  POSSIBILITY OF SUCH DAMAGE.
-*********************************************************************/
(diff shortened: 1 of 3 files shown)


src/ompl/geometric/planners/prm/src/LazyPRM.cpp
```diff
@@ -357,7 +334,9 @@ ompl::base::PlannerStatus ompl::geometric::LazyPRM::solve(const base::PlannerTer
 
     bestCost_ = opt_->infiniteCost();
     base::State *workState = si_->allocState();
-    std::pair<std::size_t, std::size_t> startGoalPair;
+
+    Vertex startPair;
```
mamoll, line 338:
```suggestion
    Vertex startVertex;
```

src/ompl/geometric/planners/prm/src/LazyPRM.cpp
```diff
@@ -357,7 +334,9 @@ ompl::base::PlannerStatus ompl::geometric::LazyPRM::solve(const base::PlannerTer
...
+    std::set<Vertex> goalPair;
```
mamoll, line 339:
```suggestion
    std::set<Vertex> goalSet;
```

src/ompl/geometric/planners/prm/src/LazyPRM.cpp
```diff
@@ -367,10 +346,19 @@ ompl::base::PlannerStatus ompl::geometric::LazyPRM::solve(const base::PlannerTer
...
+            const base::State *st = pis_.nextGoal(ptc);
+            if (st != nullptr)
+                goalM_.push_back(addMilestone(si_->cloneState(st)));
+        }
+
         sampler_->sampleUniform(workState);
         Vertex addedVertex = addMilestone(si_->cloneState(workState));
 
-        const long int solComponent = solutionComponent(&startGoalPair);
+        const long int solComponent = solutionComponent(startPair, goalPair);
```
mamoll, line 360:
```suggestion
        const long int solComponent = solutionComponent(startVertex, goalSet);
```

src/ompl/geometric/planners/prm/src/LazyPRM.cpp
```diff
@@ -467,25 +456,28 @@ void ompl::geometric::LazyPRM::markComponent(Vertex v, unsigned long int newComp
...
+    goals.clear();
+
+    // TODO: Still need to handle multiple starts, will choose first start that works
+    for (const auto &s : startM_)
     {
-        long int startComponent = vertexComponentProperty_[startM_[startIndex]];
-        for (std::size_t goalIndex = 0; goalIndex < goalM_.size(); ++goalIndex)
+        auto startComponent = vertexComponentProperty_[s];
+        std::copy_if(std::begin(goalM_), std::end(goalM_), std::inserter(goals, std::begin(goals)),
+                     [&](const auto &g) { return vertexComponentProperty_[g] == startComponent; });
```
mamoll, line 468: The code does more work than necessary, since you can stop as soon as you find a goal state that is in the same component as the start state. You don't have to iterate over all goal states. No need to change, since optimizing this part of the code is probably not worth it.

mamoll (approved):
Looks good to me, modulo some minor changes.

zkingston:
Thanks for the review - just tested locally to double check and the LazyPRM is now throwing an exception in some cases, I'll change this to a WIP and request review when I've kicked the tires a bit more.

PR #996, opened 2023-02-21
Fix AIT* segfault when calling clear before setup

yurirocha15:
Currently, the `objective_` pointer is only set when calling the `setup()` method. However, this pointer is referenced in the `clear()` method, which means that if the user calls `AITstar::clear()` before calling `AITstar::setup()` there will be a segmentation fault.

As the costs will be set to the `infinity cost` when the `setup()` method is called, I changed the `clear` method to set the costs as NaN.

### DIFF
--- a/src/ompl/geometric/planners/informedtrees/src/AITstar.cpp
+++ b/src/ompl/geometric/planners/informedtrees/src/AITstar.cpp
@@ -203,9 +203,9 @@ namespace ompl
             graph_.clear();
             forwardQueue_.clear();
             reverseQueue_.clear();
-            solutionCost_ = objective_->infiniteCost();
-            approximateSolutionCost_ = objective_->infiniteCost();
-            approximateSolutionCostToGoal_ = objective_->infiniteCost();
+            solutionCost_ = ompl::base::Cost(std::numeric_limits<double>::signaling_NaN());
+            approximateSolutionCost_ = ompl::base::Cost(std::numeric_limits<double>::signaling_NaN());
+            approximateSolutionCostToGoal_ = ompl::base::Cost(std::numeric_limits<double>::signaling_NaN());
             numIterations_ = 0u;
             numInconsistentOrUnconnectedTargets_ = 0u;
             Planner::clear();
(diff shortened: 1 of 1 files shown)


src/ompl/geometric/planners/informedtrees/src/AITstar.cpp
```diff
@@ -203,9 +203,9 @@ namespace ompl
             graph_.clear();
             forwardQueue_.clear();
             reverseQueue_.clear();
-            solutionCost_ = objective_->infiniteCost();
-            approximateSolutionCost_ = objective_->infiniteCost();
-            approximateSolutionCostToGoal_ = objective_->infiniteCost();
+            solutionCost_ = ompl::base::Cost(std::numeric_limits<double>::signaling_NaN());
+            approximateSolutionCost_ = ompl::base::Cost(std::numeric_limits<double>::signaling_NaN());
+            approximateSolutionCostToGoal_ = ompl::base::Cost(std::numeric_limits<double>::signaling_NaN());
```
zkingston, line 208: Side-stepping the actually objective's `infiniteCost()` is no good. I would instead check if `objective_` exists and then reset these costs if it exists. These are reset in `setup()` as well. That is:
```cpp
if (objective_)
{
    solutionCost_ = objective_->infiniteCost();            
    approximateSolutionCost_ = objective_->infiniteCost();
    approximateSolutionCostToGoal_ = objective_->infiniteCost();
}
```

zkingston (changes requested):
Thanks for reporting, can you change and check if this fixes the problem?

src/ompl/geometric/planners/informedtrees/src/AITstar.cpp
yurirocha15: I added the requested changes in [f742b9c](https://github.com/ompl/ompl/pull/996/commits/f742b9c43220db42416dae315e32a70fcc846708)

The reason I didn't go for this solution initially is that if the pointer does not exist, the cost won't be reset. However, as you said, those values are also set in the setup function, so I guess it is better to have a valid, outdated cost than a potentially invalid one.

yurirocha15: @zkingston I tested the suggested code, and there were no segmentation faults.

zkingston (approved):
LGTM, thank you.

Issue #997, opened 2023-02-21
AIT* can have invalid samples in its graph


marlinstrub:
The change in https://github.com/ompl/ompl/pull/990 means that AIT* can now have invalid samples in its graph (if the loop exits because the termination condition triggers).

I do think AIT* should be better at respecting termination conditions though. I'm going to look for a different solution, but PRs that fix this are obviously welcome too.

cc @zkingston @haryngod

PR #998, opened 2023-02-23
Fix adding invalid states to AIT* graph

yurirocha15:
@marlinstrub This is a quick fix for #997.

This solution is similar to the one added to EIT* in #994. I could not think of a better way to fix it, so any suggestions are welcome as well.


marlinstrub (approved):
Looks good to me, thanks.

PR #1001, opened 2023-03-09
Fix incorrect validSegmentCount in DubinsStateSpace

### LINKED ISSUE #811: Longest Valid Segment Length and State validity checking resolution
The description on the OMPL website seems a bit vague for me. Sorry, but as I understand, this is a percentage of the maximumExtent of the state space right? 

Therefore, if I want my path segments to be utmost 0.1 m apart (and given that I can guarantee collission-free segments of this length), `StateSpace::validSegmentCount()` should return the number 10 for two states a metre apart, correct?

Therefore, when the distances between every two path points for a particular solution is computed, these should all be the same number: the number returned by `getLongestValidSegmentLength()` correct? 

Am I missing something here? 
I can also see that `StateSpace::validSegmentCount()` returns a seemingly very random number. When I printout `space.distance(state1, state2) / space.getLongestValidSegmentLength()`, this doesn't seem to be anywhere close to what `validSegmentCount()` returns.

I am using a Dubins State Space.

rconnorlawson:
`validSegmentCount` is used to determine the number of interpolation points to use for collision checking a segment.

`DubinsStateSpace` incorrectly delegated to the default `CompoundStateSpace::validSegmentCount` implementation, which evaluates each subspace independently and returns the highest value.

This patch returns to the basic `StateSpace` behavior, computing a count based on edge length -- in this case, the length of the Dubins path.

Fixes #811.

### DIFF
--- a/src/ompl/base/spaces/DubinsStateSpace.h
+++ b/src/ompl/base/spaces/DubinsStateSpace.h
@@ -128,6 +128,8 @@ namespace ompl
                 return isSymmetric_;
             }
 
+            unsigned int validSegmentCount(const State *state1, const State *state2) const override;
+
             void sanityChecks() const override
             {
                 double zero = std::numeric_limits<double>::epsilon();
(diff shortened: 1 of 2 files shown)


zkingston (approved):
LGTM. Thanks for the fix.

PR #1008, opened 2023-03-27
Fix for constrained planning python bindings

### LINKED ISSUE #1007: Error in Python Bindings when Running Constrained Planning Examples
When running the example `ConstrainedPlanningSphere.py` (and commenting out the tools import, as suggested in #1006), I get the following error:
```
$ python3 ConstrainedPlanningSphere.py 
Info:    Using Projection-Based State Space!
Debug:   RRT: Planner range detected to be 1.385641
Info:    RRT: Starting planning with 1 states already in datastructure
  File ".../ompl/demos/constraint/ConstrainedPlanningSphere.py", line 160, in <module>
    spherePlanning(parser.parse_args())
  File ".../ompl/demos/constraint/ConstrainedPlanningSphere.py", line 143, in spherePlanning
    spherePlanningOnce(cp, planners[0], options.output)
  File ".../ompl/demos/constraint/ConstrainedPlanningSphere.py", line 98, in spherePlanningOnce
    stat = cp.solveOnce(output, "sphere")
  File ".../ompl/demos/constraint/ConstrainedPlanningCommon.py", line 207, in solveOnce
    stat = self.ss.solve(self.options.time)
TypeError: No Python class registered for C++ class Eigen::Ref<Eigen::Matrix<double, -1, 1, 0, -1, 1> const, 0, Eigen::InnerStride<1> >

```

zkingston:
- Update to the signature needed to find & replace in Python binding generation needed for constrained planning.
- Optional tool import

Fixes #1007

### DIFF
--- a/demos/constraint/ConstrainedPlanningCommon.py
+++ b/demos/constraint/ConstrainedPlanningCommon.py
@@ -60,6 +60,7 @@ except ImportError:
         from ompl import tools as ot
     except ImportError:
         pass
+
 import datetime
 
 
--- a/demos/constraint/ConstrainedPlanningCommon.py
+++ b/demos/constraint/ConstrainedPlanningCommon.py
@@ -41,7 +41,11 @@ try:
     from ompl import util as ou
     from ompl import base as ob
     from ompl import geometric as og
-    from ompl import tools as ot
+    try:
+        from ompl import tools as ot
+    except ImportError:
+        pass
+
 except ImportError:
     # if the ompl module is not in the PYTHONPATH assume it is installed in a
     # subdirectory of the parent directory called "py-bindings."
@@ -52,7 +56,10 @@ except ImportError:
     from ompl import util as ou
     from ompl import base as ob
     from ompl import geometric as og
-    from ompl import tools as ot
+    try:
+        from ompl import tools as ot
+    except ImportError:
+        pass
 import datetime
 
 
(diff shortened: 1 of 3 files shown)


zkingston:
Thanks for the feedback, that's been fixed.

demos/constraint/ConstrainedPlanningCommon.py
```diff
@@ -254,6 +262,10 @@ def solveOnce(self, output=False, name="ompl"):
         return stat
 
     def setupBenchmark(self, planners, problem):
+        if ot:
```
mamoll, line 265:
```suggestion
        if not ot:
```

demos/constraint/ConstrainedPlanningCommon.py
```diff
@@ -254,6 +262,10 @@ def solveOnce(self, output=False, name="ompl"):
...
+            print("Benchmarking not available, no ompl.tools")
+            return
```
mamoll, line 267:
```suggestion
            exit(0)
```
If you don't exit, run_benchmark will still get called, which produces an error.

mamoll (changes requested):
Can confirm this works when not running benchmarking. When benchmarking is not available, setupBenchmark returns and runBenchmark still gets called (which will produce an error). My two suggested changes will fix this. With my fixes, I can benchmark on macOS (where ompl.tools is available).

mamoll (approved):

PR #1011, opened 2023-03-28
Build tools module for Python bindings

zkingston:
Key fix was making `RunProperties` a subclass of `std::map<std::string, std::string>`, rather than just an alias.


mamoll (approved):
Works for me on macOS. It would have been easier to review if unrelated changes (formatting changes, and moving code from headers to cpp) would have been in a separate PR.

zkingston:
Thanks for reviewing Mark!

cohnt:
Can confirm that this works on my end (Ubuntu 22.04)

PR #1013, opened 2023-03-31
fix compiler warnings

mamoll:
Fixes for compiler warnings (tested on macOS with clang-16++ and Boost 1.81)


zkingston (approved):
Adding the `_HAS_AUTO_PTR_ETC=0` definition works for `clang` but not for `gcc`, unfortunately. LGTM otherwise.

PR #1015, opened 2023-04-03
Use more conservative types in PLY files

sotpapathe:
In PLY files it is more common and well supported to use a `uchar` for the number of face vertices and an `int` for the vertex indices. These are also the types used in the PLY format description. Certain software (e.g. [Meshlab](https://www.meshlab.net/)) doesn't support using a `uint` for either type.

Although this is arguably a bug in Meshlab, it doesn't hurt if OMPL outputs PLY files in a more widely supported format, especially considering there are no drawbacks for doing so.

http://paulbourke.net/dataformats/ply/


zkingston (approved):
LGTM. There may eventually be an issue with a chart in an atlas being constructed with over 255 neighbors, but compatibility with actual software to look at PLY files is more important.

PR #1022, opened 2023-04-20
Support for Python binding installation through pip & pypy version checking

zkingston:
- Added a setup.py to the `py-bindigns` directory to enable `pip install .`
- Added checks to verify `pypy` version matches desired Python executable version


mamoll:
What are users supposed to do with the output of `pip install .`? I suspect this is not targeted at OMPL users, but more something targeted at OMPL maintainers. In that case, this probably needs more work:
* You'll want to add build targets to [create python wheels](https://realpython.com/python-wheels) and maybe document the process of uploading wheels to Pypi.
* You may want to make this part of CI, so that you automatically get build artifacts in clean build environments. Since generating Python bindings with Py++ exceeds the build time limits of the free plan, this is currently not an option.
* A lot of people seem to like `conda`, so maybe prepare packaging instructions for Conda?

wbthomason:
The intent of these initial changes is to enable building for `conda`, as well as easier local development with Python (e.g. using `pip install -e .`. We have a working `conda` recipe ready, but it requires these changes to build and install correctly.

From discussions with @zkingston, while we still want a PyPI upload (and CI integration) to be feasible in the near future, we need to do more work to get wheels built (and potentially deal with the fact that the wheels may depend on specific Boost versions on users' computers).

zkingston:
As @wbthomason says, the goal of this PR is to support a `conda` package and local development of the Python bindings in other, local environments. I'll change the name of the PR to better reflect what this actually achieves.

CMakeModules/Findpypy.cmake
```diff
@@ -5,8 +5,18 @@ include(FindPackageHandleStandardArgs)
 find_program(PYPY NAMES pypy${PYTHON_VERSION_MAJOR} pypy)
 
 if(PYPY)
+  execute_process(
+    COMMAND ${PYPY} "-c" "import platform; print(platform.python_version())"
+    OUTPUT_VARIABLE PYPY_PYTHON_VERSION
+    OUTPUT_STRIP_TRAILING_WHITESPACE)
+  message(WARNING ${PYPY_PYTHON_VERSION})
```
mamoll, line 12: this should probably be removed

CMakeModules/Findpypy.cmake
```diff
@@ -5,8 +5,18 @@ include(FindPackageHandleStandardArgs)
...
+  if(${PYPY_PYTHON_VERSION} VERSION_GREATER_EQUAL ${PYTHON_VERSION_FULL})
```
mamoll, line 13: Is this necessary? I typically install python and pypy via a package manager. Pypy's version is typically older than the python version and this hasn't caused issues for me.

zkingston: When we were developing a `conda` target we ran into issues with respect to Python versions in the `conda` environment, and `pypy` in the base environment leaking through `conda`'s path (I believe with Python 3.10 in the `conda` env). @wbthomason can elaborate more.

CMakeModules/Findpypy.cmake
wbthomason:
```suggestion
```

CMakeModules/Findpypy.cmake
wbthomason: Yes - `conda` seems to be a leakier abstraction than we hoped, and having `pypy` for Python 2 installed through the system package manager while trying to build the bindings for Python 3 caused failures that motivated this check. I forget why we're testing against `PYTHON_VERSION_FULL` and not just `PYTHON_VERSION_MAJOR`, though...perhaps because there's no equivalent for `pypy`? I can experiment and check if this can be relaxed.

@mamoll, when you say that your `pypy` version is usually older than the Python version, do you mean different minor versions (e.g. `pypy` at 3.9, Python at 3.10) or major versions?

wbthomason:
@zkingston Nitpick, but we should rename this branch, as it no longer does anything with `pybind`.

CMakeModules/Findpypy.cmake
wbthomason: Actually, I think @mamoll is right here - we already restrict to `pypy3` on line 4. I've just run an experiment building the bindings with and without `pypy`, with this conditional removed, inside `conda`, and it seems to have worked.

@zkingston if you can confirm, I think we can remove this check.

zkingston: @mamoll @wbthomason I changed the check to instead just check the major version - there is an issue if pypy does not match (which can happen if pypy2 is installed with a Python 3 executable).

mamoll: @wbthomason yes, I think major version match between pypy and python is sufficient (at least, that has been my experience).

zkingston:
@mamoll comments have been addressed, is this branch good to merge now?

mamoll (approved):

Issue #874, opened 2022-01-19
Make more planners able to inform of intermediate solutions


Hugal31:
Currently, only the RRTstar, BITstar, RRTXstatic and the CForest uses `ProblemDefinition::getIntermediateSolutionCallback`. It would be great to support this feature in other planners, such as the PRMs, SST, etc...

PR #854 is working on implementation for SST.

PR #1024, opened 2023-04-26
AIT*/EIT*: Add support for IntermediateSolutionCallback

yurirocha15:
Related to #874 

This PR adds a call to `getIntermediateSoluitionCallback` to AIT* and EIT* planners, so the `CostConvergenceTermination`  (and `CForest`) can be used with them.

On a side note, the planner specs were already setting `canReportIntermediateSolutions` as true.


marlinstrub:
Looks good to me as well. Thanks!

PR #1028, opened 2023-04-28
fix: Issue error if matching Boost to Python/Numpy versions fails

wbthomason:
As noted by @roberto-martinmartin, the macros OMPL uses to find the right `Boost.Python`/`Boost.Numpy` versions for given Python/Numpy versions can silently fail, leading to confusion about whether or not the Python bindings have been succesfully built.

This PR emits non-fatal error messages if this matching fails. The errors are non-fatal because I think these macros may run even for users who do not care about building the Python bindings; as an alternative, we could change the `CMakeLists` so that this is not the case, and issue a fatal error instead.


wbthomason:
Apparently `SEND_ERROR` still results in a non-zero exit code. I'll change the message type to `WARNING` for now, but this may motivate changing the CMakeLists logic to only search for matching Boost libraries if the Python bindings are to be built (i.e. either the relevant option has been set to `ON` or all other required dependencies have been found).

mamoll (approved):

roberto-martinmartin:
Hi!
Thanks for taking care of this issue.
I'm going again over some issues with this, now on my Mac. 
Just to confirm: should I be seeing any warning if the target update_bindings is not created or not yet?
I'm using the main branch, which I think contains now what was before in the branch. I 'cmake' and 'make update_bindings' and get the error I had before (No rule to make target `update_bindings') but do not see the warning.
Thanks!

PR #1048, opened 2023-05-26
Add a special case to checkAndRepair for the two state case

### LINKED ISSUE #1026: PathGeometric::checkAndRepair problem with path composed of two states
I am using PathGeometric::checkAndRepair method to fix a path that becomes invalid due to an obstacle. The path has only two states - start and goal. When I call checkAndRepair on this path, I always get (true, true) as output. 

Looking through the code, it looks like the error is originating due to the condition used in the for loop on line 234 (PathGeometric.cpp on main, last commit [a02c63f]). When a path has only two states, the for loop on line 234 never gets executed, and hence path does not get repaired with the method returning (true, true). If I add an intermediate state to the path, then checkAndRepair works as expected.

I would appreciate any input in this regard!

Thank you and regards!

zkingston:
Fixes #1026. Special case for two state case to validate motion, as it won't be checked otherwise.

### DIFF
--- a/src/ompl/geometric/PathGeometric.h
+++ b/src/ompl/geometric/PathGeometric.h
@@ -177,8 +177,7 @@ namespace ompl
                 \return A pair of boolean values is returned. The first
                 value represents the validity of the path before any
                 change was made. The second value represents the
-                validity of the path after changes were attempted. If
-                no changes are attempted, the both values are true.
+                validity of the path after changes were attempted.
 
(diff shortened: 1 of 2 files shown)


mamoll (approved):

mamoll:
LGTM, but update the documentation in the PathGeometric.h header file by removing this line in the `checkAndRepair` documentation:
```
If no changes are attempted, the both values are true.
```

Issue #1065, opened 2023-06-23
Bug: Ubuntu 23.04 failure to build OMPL - Could NOT find Boost (missing: filesystem system) 


Ryanf55:
OS: Ubuntu 23.04
Boost: `Version: 1.74.0.3ubuntu7` installed through `apt` as `libboost-dev`
g++ version: `Ubuntu 12.2.0-17ubuntu1`
CMake version: `3.27.0-rc3`

```
--- stderr: ompl
CMake Error at /usr/local/share/cmake-3.27/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Boost (missing: filesystem system) (found suitable version
  "1.74.0", minimum required is "1.58")
Call Stack (most recent call first):
  /usr/local/share/cmake-3.27/Modules/FindPackageHandleStandardArgs.cmake:600 (_FPHSA_FAILURE_MESSAGE)
  /usr/local/share/cmake-3.27/Modules/FindBoost.cmake:2378 (find_package_handle_standard_args)
  CMakeLists.txt:52 (find_package)
```

Reference: https://stackoverflow.com/questions/71610458/could-not-find-boost-missing-system-found-suitable-versihow-on-1-78-0-min

zkingston:
I believe you need to install `libboost-all-dev`. `libboost-dev` does not pull in all components of boost.

Ryanf55:
Yea, perhaps they changed what's installed by default... Here's what's on my system with `libboost-dev`. 

```
ryan@ryan-B650-970:~/Development/nav2_ws/src/ompl/ompl$ find /usr  -name "*.cmake" | grep -v 3.27 | grep -v 3.25 | grep boost
/usr/lib/x86_64-linux-gnu/cmake/boost_wserialization-1.74.0/boost_wserialization-config-version.cmake
/usr/lib/x86_64-linux-gnu/cmake/boost_wserialization-1.74.0/boost_wserialization-config.cmake
/usr/lib/x86_64-linux-gnu/cmake/boost_wserialization-1.74.0/libboost_wserialization-variant-shared.cmake
/usr/lib/x86_64-linux-gnu/cmake/boost_wserialization-1.74.0/libboost_wserialization-variant-static.cmake
/usr/lib/x86_64-linux-gnu/cmake/boost_regex-1.74.0/boost_regex-config.cmake
/usr/lib/x86_64-linux-gnu/cmake/boost_regex-1.74.0/libboost_regex-variant-static.cmake
[...]

PR #1066, opened 2023-06-23
CMake: Switch to exported namespace targets and prefer config mode

Ryanf55:
This allows use of config mode in boost to find it. Regardless of whether boost is found via config or module, the alias targets can, and are preferred, to be used in CMake. 

Reference: https://cmake.org/cmake/help/latest/module/FindBoost.html#boost-cmake
Relates to https://github.com/ompl/ompl/issues/1065

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,9 +48,7 @@ set_package_properties(Boost PROPERTIES
     URL "https://www.boost.org"
     PURPOSE "Used throughout OMPL for data serialization, graphs, etc.")
 set(Boost_USE_MULTITHREADED ON)
-set(Boost_NO_BOOST_CMAKE ON)
-find_package(Boost 1.58 QUIET REQUIRED COMPONENTS serialization filesystem system program_options)
-include_directories(SYSTEM ${Boost_INCLUDE_DIR})
+find_package(Boost 1.58 REQUIRED COMPONENTS serialization filesystem system program_options)
 
 # Add support in Boost::Python for std::shared_ptr
 # This is a hack that replaces boost::shared_ptr related code with std::shared_ptr.
--- a/src/ompl/CMakeLists.txt
+++ b/src/ompl/CMakeLists.txt
@@ -36,15 +36,20 @@ source_group("OMPL Headers" FILES "${OMPL_HEADERS}")
 # build the library
 if(MSVC)
     add_library(ompl STATIC ${OMPL_SOURCE_CODE})
-else(MSVC)
+else()
     add_library(ompl SHARED ${OMPL_SOURCE_CODE})
-endif(MSVC)
+endif()
 target_link_libraries(ompl
     ${OMPL_LINK_LIBRARIES}
     ${Boost_SERIALIZATION_LIBRARY}
     ${Boost_FILESYSTEM_LIBRARY}
     ${Boost_SYSTEM_LIBRARY}
-    ${CMAKE_THREAD_LIBS_INIT})
+    ${CMAKE_THREAD_LIBS_INIT}
+    Boost::filesystem
+    Boost::serialization
+    Boost::filesystem
+    Boost::system
+    Boost::program_options)
 
 if (OMPL_HAVE_SPOT)
     target_link_libraries(ompl ${SPOT_LIBRARIES})
--- a/src/ompl/CMakeLists.txt
+++ b/src/ompl/CMakeLists.txt
@@ -41,15 +41,11 @@ else()
 endif()
 target_link_libraries(ompl
     ${OMPL_LINK_LIBRARIES}
-    ${Boost_SERIALIZATION_LIBRARY}
-    ${Boost_FILESYSTEM_LIBRARY}
-    ${Boost_SYSTEM_LIBRARY}
     ${CMAKE_THREAD_LIBS_INIT}
     Boost::filesystem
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -48,9 +48,7 @@ set_package_properties(Boost PROPERTIES
     URL "https://www.boost.org"
     PURPOSE "Used throughout OMPL for data serialization, graphs, etc.")
 set(Boost_USE_MULTITHREADED ON)
-set(Boost_NO_BOOST_CMAKE ON)
-find_package(Boost 1.58 QUIET REQUIRED COMPONENTS serialization filesystem system program_options)
```
Ryanf55, line 52: I can keep QUIET if you want...

src/ompl/CMakeLists.txt
```diff
@@ -36,15 +36,20 @@ source_group("OMPL Headers" FILES "${OMPL_HEADERS}")
 # build the library
 if(MSVC)
     add_library(ompl STATIC ${OMPL_SOURCE_CODE})
-else(MSVC)
```
Ryanf55, line 39: This is no longer recommended in CMake btw

src/ompl/CMakeLists.txt
```diff
@@ -36,15 +36,20 @@ source_group("OMPL Headers" FILES "${OMPL_HEADERS}")
...
-endif(MSVC)
+endif()
 target_link_libraries(ompl
     ${OMPL_LINK_LIBRARIES}
     ${Boost_SERIALIZATION_LIBRARY}
     ${Boost_FILESYSTEM_LIBRARY}
     ${Boost_SYSTEM_LIBRARY}
-    ${CMAKE_THREAD_LIBS_INIT})
+    ${CMAKE_THREAD_LIBS_INIT}
+    Boost::filesystem
```
Ryanf55, line 48: Linking alias targets ensures they actually exist, vs variables which can silently fail.

CMakeLists.txt
```diff
@@ -48,9 +48,7 @@ set_package_properties(Boost PROPERTIES
     URL "https://www.boost.org"
     PURPOSE "Used throughout OMPL for data serialization, graphs, etc.")
 set(Boost_USE_MULTITHREADED ON)
-set(Boost_NO_BOOST_CMAKE ON)
```
Ryanf55, line 51: I cannot tell what reason a user would enable this. CMake developers recommend the use of CONFIG mode, whenever possible, but this disables it.

mamoll: This was probably put in when the Boost CMake support was still experimental. I'm in favor of this change.

CMakeLists.txt
```diff
@@ -48,9 +48,7 @@ set_package_properties(Boost PROPERTIES
...
-find_package(Boost 1.58 QUIET REQUIRED COMPONENTS serialization filesystem system program_options)
-include_directories(SYSTEM ${Boost_INCLUDE_DIR})
```
mamoll, line 53: Why is this line deleted? If I have set the environment variable `BOOST_ROOT` to some non-standard boost installation, will the correct include path be used?

mamoll (commented):
I have one question, but I'm ok with this change.

CMakeLists.txt
Ryanf55: The includes are brought in through the call to `target_link_libraries`. The exported namespace targets that are found above already have the properties set. 

According to the [docs](https://cmake.org/cmake/help/latest/module/FindBoost.html#hints), it looks like `BOOST_ROOT` is used by the module as a hint of where to find boost. Therefore, this would be my expected behavior if you have installed boost in a custom location.
1. Cmake tries to find it boostConfig.cmake from find_package and fails since it's not on cmake prefix path
2. You gave a hint for the module to find boost from `BOOST_ROOT`
3. The find succeeds, and populates the imported namespace targets
4. `target_link_libraries` functions the same.

If you have a system without boost installed in system headers and instead in a non-standard location, please try. 

If you think this is a common way people find boost, perhaps one could add it to CI?

mamoll: > The includes are brought in through the call to target_link_libraries. The exported namespace targets that are found above already have the properties set.

Neat! I learnt something today. I tried it and it works with boost in a non-standard location.

src/ompl/CMakeLists.txt
mamoll: Don't you need to add Boost::serialization and delete ${Boost_SERIALIZATION_LIBRARY} and ${Boost_FILESYSTEM_LIBRARY}?

Ryanf55: `Boost::serialization` is already there, but yes, I forgot to remove the old properties. Great catch. 
While I'm at it, I'll sort them alphabetically. 

Fixed in a9c2ba50184365cec6599e011346fa6b30ce329f

Ryanf55:
Much like here I try to get OMPL to consume the namespaced exported targets from boost, `ompl` doesn't install an export set. 
https://cmake.org/cmake/help/latest/guide/tutorial/Adding%20Export%20Configuration.html

I can either add that functionality to this PR, or do it in a follow up. Let me know which you prefer.

zkingston (approved):
LGTM, thanks for the change. Tried on my machine with defaults and with boost in a non-standard location.

PR #1070, opened 2023-07-10
Only prune goal in BIT* if it is not part of the current best solution

### LINKED ISSUE #949: Segfault in ABIT*: graph reduced to zero or one vertex
ABIT* sometimes segfaults because the graph is reduced to zero vertices (or one). Here's the terminal output for zero vertices (shortened down for clarity):

```
informedtrees/src/BITstar.cpp:993 - robot[ABITstar]: Finished with a solution of 
cost 6.7633 (2 vertices) found from 73 samples by processing 162 edges (135 collision checked) to 
create 18 vertices and perform 0 rewirings. The final graph has 0 vertices.
BITstar.cpp:753 - curVertex: vertex = goal, depth: 0
Stack trace (most recent call last) in thread 1374030:
#31   Object "", at 0xffffffffffffffff, in 
#30  Source "../sysdeps/unix/sysv/linux/x86_64/clone3.S", line 81, in __clone3 [0x7f46cd0139ff]
```
It seems this is caused by pruning:
```
BITstar.cpp:993 - robot[ABITstar]: Finished with a solution of cost 6.7633 (2 vertices)
 found from 73 samples by processing 162 edges (135 collision checked) to create 18 vertices
 and perform 0 rewirings. The final graph has 0 vertices.
```

And it can be fixed by setting `use_graph_pruning` in ABIT* to `false`.

marlinstrub:
This hopefully fixes #949.

The errors reported in #949 suggest that the start and goal of a trivial solution (direct start to goal connection) are sometimes pruned. If there are multiple goals and if the heuristic is perfect for a trivial solution (e.g., Euclidean distance heuristic for path length objective in an environment that allows for the trivial solution), then the current code will indeed prune the start and goal. The proposed change directly prevents this by checking the cost of the goal and by adding previously pruned starts back to the samples if necessary.

### DIFF
--- a/src/ompl/geometric/planners/informedtrees/bitstar/src/ImplicitGraph.cpp
+++ b/src/ompl/geometric/planners/informedtrees/bitstar/src/ImplicitGraph.cpp
@@ -1093,8 +1093,14 @@ namespace ompl
                 // Run until at the end:
                 while (goalIter != goalEnd)
                 {
-                    // Check if this goal has met the criteria to be pruned
-                    if (this->canSampleBePruned(*goalIter))
+                    // Check if this goal has met the criteria to be pruned, but make sure it is not the goal that is
+                    // the current best solution. The current goal can satisfy the condition for pruning if the
+                    // heuristic is perfect, since samples are pruned unless the combined costs of the heuristic
+                    // cost-to-come and heuristic cost-to-go is (strictly) lower than the current solution cost. If the
+                    // heuristic is perfect, then the heuristic cost-to-come is equal to the solution cost (and the
+                    // heuristic cost-to-go is zero, since this is a goal).
+                    if (this->canSampleBePruned(*goalIter) &&
+                        costHelpPtr_->isCostNotEquivalentTo((*goalIter)->getCost(), solutionCost_))
                     {
                         // It has, remove the goal vertex completely
                         // Check if this vertex is in the tree
--- a/src/ompl/geometric/planners/informedtrees/bitstar/src/ImplicitGraph.cpp
+++ b/src/ompl/geometric/planners/informedtrees/bitstar/src/ImplicitGraph.cpp
@@ -527,6 +527,9 @@ namespace ompl
                         // Add it back to the vector
                         startVertices_.push_back(*prunedStartIter);
 
+                        // Add as a sample
+                        this->addToSamples(*prunedStartIter);
+
                         // Add this vertex to the queue.
                         // queuePtr_->enqueueVertex(*prunedStartIter);
 


Hugal31:
This fixes the bug for me. Thanks!

marlinstrub:
Thanks for confirming!

andreas-botbuilt:
Same here. The fix works!

zkingston (approved):
LGTM, thanks for the fix.

src/ompl/geometric/planners/informedtrees/bitstar/src/ImplicitGraph.cpp
```diff
@@ -527,6 +527,9 @@ namespace ompl
                         // Add it back to the vector
                         startVertices_.push_back(*prunedStartIter);
 
+                        // Add as a sample
+                        this->addToSamples(*prunedStartIter);
```
gammell, line 531: This part of the code is "reactivating" previously removed starts that, thanks to the addition of a new goal, now could be useful.

But (if I remember correctly) everything in `startVertices_` is considered part of the search tree by definition (even if it's just a singular root), so I don't think they should also be added to the samples...

gammell: I'm pretty sure not, Line 537 right below has:
```
// Add the vertex to the set of vertices.
this->registerAsVertex(*prunedStartIter);
```
and I'm pretty sure the Vertices and Samples are supposed to be mutually exclusive sets... I think.

src/ompl/geometric/planners/informedtrees/bitstar/src/ImplicitGraph.cpp
```diff
@@ -1093,8 +1096,14 @@ namespace ompl
...
-                    // Check if this goal has met the criteria to be pruned
-                    if (this->canSampleBePruned(*goalIter))
+                    // Check if this goal has met the criteria to be pruned, but make sure it is not the goal that is
+                    // the current best solution. The current goal can satisfy the condition for pruning if the
+                    // heuristic is perfect, since samples are pruned unless the combined costs of the heuristic
+                    // cost-to-come and heuristic cost-to-go is (strictly) lower than the current solution cost. If the
+                    // heuristic is perfect, then the heuristic cost-to-come is equal to the solution cost (and the
+                    // heuristic cost-to-go is zero, since this is a goal).
+                    if (this->canSampleBePruned(*goalIter) &&
+                        costHelpPtr_->isCostNotEquivalentTo((*goalIter)->getCost(), solutionCost_))
```
gammell, line 1106: Expanding the comparison inequality for goals from `<` to `<=` seems like a harmless and way to check if the goal-considered-for-removale is part of the current solution. It's slightly indirect, but I don't know if there's a direct way to check if a vertex is part of the current solution. And I don't see any harm in the rare (impossible?) case of keeping a goal that meets equality but is not part of the current solution.

src/ompl/geometric/planners/informedtrees/bitstar/src/ImplicitGraph.cpp
gammell: Oh, [Lines 430-431](https://github.com/robotic-esp/ompl/blob/eeebb7e3a29ee50057e41cea573815c1f6121bf3/src/ompl/geometric/planners/informedtrees/bitstar/src/ImplicitGraph.cpp#L430-L431) add a new start to both samples and vertices,
```
this->addToSamples(startVertices_.back());
this->registerAsVertex(startVertices_.back());
```
so that would suggest this is right...? I don't remember this code very well anymore.

gammell: Looking at what when a sample is first connected as a new vertex in [`BITstar::addEdge`](https://github.com/robotic-esp/ompl/blob/eeebb7e3a29ee50057e41cea573815c1f6121bf3/src/ompl/geometric/planners/informedtrees/src/BITstar.cpp#L808C16-L848) and what [`BITstar::ImplicitGraph::registerAsVertex`](https://github.com/robotic-esp/ompl/blob/eeebb7e3a29ee50057e41cea573815c1f6121bf3/src/ompl/geometric/planners/informedtrees/bitstar/src/ImplicitGraph.cpp#L736C14-L756) actually does, it looks to me like all vertices should be, and stay, in the set of samples.

So this is in fact right, sorry for the confusion!

gammell (commented):
Looks good to me with more thinking!

PR #1074, opened 2023-07-12
fix: EIGEN3_INCLUDE_DIR is not defined when Eigen3 is installed from source code.

daizhirui:
The latest Eigen3 version 3.4.90 does not define `EIGEN3_INCLUDE_DIR` but embeds everything in an imported target `Eigen3::Eigen`.


zkingston (approved):
LGTM, thanks for the contribution.

Issue #1014, opened 2023-04-02
DubinsState space implementation not using set classification


Jaeyoung-Lim:
**Problem Description**
While the DubinsStateSpace implementation mentions the Dubins set classification method(A.M. Shkel and V. Lumelsky, “Classification of the Dubins set,” Robotics and Autonomous Systems, 34(4):179-202, 2001), the implementation computes the Dubins distance exhaustively.

https://github.com/ompl/ompl/blob/55588baaccbf4b48da926939e87f76d2f717ff70/src/ompl/base/spaces/DubinsStateSpace.h#L56-L62

@mamoll Was there a specific reason that the Dubins set classification scheme was not implemented? or is this a matter of a missing implementation? If it is the latter, I would be interested in contributing the implementation for the potential speedup of a factor of 2-3.

mamoll:
No, there was no reason I didn't do it. Implementing this state space was a bit of side project without an immediate need, so I considered the current implementation "good enough" at the time. If you implement this, please do measure the speed up. I'm curious to see how much of a difference it makes.

Jaeyoung-Lim:
@mamoll Thanks! These are some primary results on my side constrained to the long path case

![bitmap](https://user-images.githubusercontent.com/5248102/229501342-96b2f9d8-8c4a-471b-b6fa-7749b94d15bb.png)

zkingston:
Wow, great speed up. I would definitely review a PR that implements these changes.

Jaeyoung-Lim:
Closing as addressed in #1080

PR #1080, opened 2023-07-24
Use of Dubins Set Classification for faster distance computations in DubinsStateSpace

Jaeyoung-Lim:
**Problem Description**
Currently, the Dubins distance computation is done by going through all six path types exhaustively. This is quite inefficient for applications such as sampling-based planners, as the Dubins distance need to be computed many times repeatedly.

It has been proposed by Shkel(2001) that certain Dubins path types can be excluded by looking at the start and end configurations of the path. While the current OMPL implementation of DubinsStateSpace references the paper from Shkel(2001), the implementation goes through every Dubins path type exhaustively. This was discussed previously in an issue: https://github.com/ompl/ompl/issues/1014 with @zkingston and @mamoll 

**Proposed Solution**
The proposed solution is to implement the Dubins Set Classification theme in the `DubinsStateSpace`.

While trying to implement the Dubins Set Classification method, I ran into some issues where the Dubins Set Classification results were not correct for some cases.

I realized these were related to some issues with the original Dubins set classifications scheme proposed in Shkel(2001), and presented the corrections in the [Workshop on Energy Efficient Robotic Systems](https://aerial-robotics-workshop-icra2023.com/agenda/) at [ICRA 2023](https://www.icra2023.org/)
- Paper: [Paper](https://www.research-collection.ethz.ch/handle/20.500.11850/615185)

This work was done with @acfloria @nrjl @rikba

This PR adds an implementation of the DubinsSetClassification including the corrections regarding the long path case classification and the classification of Dubins Path types for equivalency group a_12, a_21, a_34, a_43

**Evaluation**
The following shows a comparison between the Dubins Set Classification and exhaustive Dubins calculations. (Ours being the Dubins set classification)
image

The results show that the set classification takes 42% of the time compared to exhaustive Dubins calculations, while always returning the same answer with the exhaustive Dubins path calculations

### DIFF
--- a/src/ompl/base/spaces/DubinsStateSpace.h
+++ b/src/ompl/base/spaces/DubinsStateSpace.h
@@ -71,6 +71,26 @@ namespace ompl
                 DUBINS_STRAIGHT = 1,
                 DUBINS_RIGHT = 2
             };
+
+            enum DubinsClass
+            {
+                A11 = 0,
+                A12 = 1,
+                A13 = 2,
+                A14 = 3,
+                A21 = 4,
+                A22 = 5,
+                A23 = 6,
+                A24 = 7,
+                A31 = 8,
+                A32 = 9,
+                A33 = 10,
+                A34 = 11,
+                A41 = 12,
+                A42 = 13,
+                A43 = 14,
+                A44 = 15
+            };
             /** \brief Dubins path types */
             static const DubinsPathSegmentType dubinsPathType[6][3];
             /** \brief Complete description of a Dubins path */
--- a/src/ompl/base/spaces/DubinsStateSpace.h
+++ b/src/ompl/base/spaces/DubinsStateSpace.h
@@ -72,25 +72,6 @@ namespace ompl
                 DUBINS_RIGHT = 2
             };
 
-            enum DubinsClass
-            {
-                A11 = 0,
-                A12 = 1,
-                A13 = 2,
-                A14 = 3,
-                A21 = 4,
-                A22 = 5,
-                A23 = 6,
-                A24 = 7,
-                A31 = 8,
-                A32 = 9,
-                A33 = 10,
-                A34 = 11,
-                A41 = 12,
-                A42 = 13,
-                A43 = 14,
-                A44 = 15
-            };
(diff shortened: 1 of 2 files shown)


mamoll:
I am running `demo_GeometricCarPlanning --dubins --distance` to visualize the results as a basic sanity check. There are several erroneous results. Can you confirm that the distance between (0,0,0) and (1.9,.2,-pi) is 1.79769e+308 (or something like that)? My first guess would be a divide by 0 somewhere, but I don't see anything obvious. Let me know if you also see weird values in the output of `demo_GeometricCarPlanning --dubins --distance`.

Jaeyoung-Lim:
@mamoll Indeed, I could reproduce the issues. The problem was that in some cases, alpha and beta were not in the range of [0, 2pi]. I have hardened this by running a mod2pi in the dubins calculation and it seems to be fixed on myside. Would you able to check the fix? Thanks!

src/ompl/base/spaces/DubinsStateSpace.h
```diff
@@ -71,6 +71,26 @@ namespace ompl
                 DUBINS_STRAIGHT = 1,
                 DUBINS_RIGHT = 2
             };
+
+            enum DubinsClass
```
mamoll, line 75: I'd move this to DubinsStateSpace.cpp inside an anonymous namespace, since it is only used internally and not in the public API.

mamoll (approved):
This is great work. I did some testing to confirm that the old and new code are input/output equivalent and this appears to be the case. I also confirmed that the new code provides a significant performance improvement. Thank you for your contribution.

I have one small stylistic comment. Let me know if you agree.

src/ompl/base/spaces/DubinsStateSpace.h
Jaeyoung-Lim: Done!

mamoll (approved):

mamoll:
Thanks again, @Jaeyoung-Lim!

Jaeyoung-Lim:
@mamoll Happy to contribute! Thanks for the merge!

PR #1085, opened 2023-08-02
build: use CMake to export boost dependency

felixf4xu:
Hi,

Currently ompl exports boost's target at https://github.com/ompl/ompl/blob/main/omplConfig.cmake.in#L43 but ompl does not find boost, this leads to an error for other project which uses ompl but does not directly use boost like this:

>CMake Error at CMakeLists.txt:75 (add_library):
  Target "nav2_smac_planner" links to target "Boost::serialization" but the
  target was not found.  Perhaps a find_package() call is missing for an
  IMPORTED target, or an ALIAS target is missing?

This PR use `find_dependency` to fix it.

PS:
There is another draft PR (#1067) also trying to fix this, but that PR seems to be on hold for a while.


tylerjw (approved):
This looks like a helpful fix. Thank you!

ChrisThrasher (approved):

Ryanf55:
If the maintainers are happy with this change, then I'll get #1067 rebased, finished and ready for merge.

zkingston:
I'm happy with this change. Sorry to leave #1067 in the lurch for a while, but all the changes listed here and there are all great for modernizing the build system. @Ryanf55 if you could finish the rebase and ping me when I should do a final review, that would be great. 

I'm would like to close this PR in favor of finishing up #1067 if possible, but I'll leave it open until everything is done.

Ryanf55:
> I'm happy with this change. Sorry to leave #1067 in the lurch for a while, but all the changes listed here and there are all great for modernizing the build system. @Ryanf55 if you could finish the rebase and ping me when I should do a final review, that would be great.
> 
> I'm would like to close this PR in favor of finishing up #1067 if possible, but I'll leave it open until everything is done.

Will do!

Ryanf55:
Upon further inspection, I disagree with this PR, because the demos `are` installed. See below:
https://github.com/ompl/ompl/blob/1aed811a0f23023246c576338aab1d45d2ebc517/demos/CMakeLists.txt#L91

Ryanf55:
> [zkingston](/zkingston)

Ready for review! 

Note - I totally hosed my system by accidentally installing flann with both debian and a cmake system. May take a bit to recover.

felixf4xu:
> Upon further inspection, I disagree with this PR, because the demos `are` installed. See below:
> 
> https://github.com/ompl/ompl/blob/1aed811a0f23023246c576338aab1d45d2ebc517/demos/CMakeLists.txt#L91
> 
> 
> Consumers should be required to have boost program options.

demo src files are installed does not mean demo src will be used by _Consumers_ as a cmake library.

If you check the current installation files, demo .cpp files are installed to `share` folder, generally speaking, _Consumers_ are not supposed to include files from `share`. Actually ompl only install demo .cpp files not .h file, this is usually for users who wants to build demo from source. It's fine to include any headers there, of course demos can use `program_options`.

Back to ompl as a library, only `include` folder and `lib` are necessary for `Consumers`. If `Boost::system` is used, then it's `ompl`'s responsibility to export it. In those include .h files, `program_options` are not used.

Ryanf55:
> > Upon further inspection, I disagree with this PR, because the demos `are` installed. See below:
> > https://github.com/ompl/ompl/blob/1aed811a0f23023246c576338aab1d45d2ebc517/demos/CMakeLists.txt#L91
> > 
> > Consumers should be required to have boost program options.
> 
> demo src files are installed does not mean demo src will be used by _Consumers_ as a cmake library.
> 
> If you check the current installation files, demo .cpp files are installed to `share` folder, generally speaking, _Consumers_ are not supposed to include files from `share`. Actually ompl only install demo .cpp files not .h file, this is usually for users who wants to build demo from source. It's fine to include any headers there, of course demos can use `program_options`.
> 
> Back to ompl as a library, only `include` folder and `lib` are necessary for `Consumers`. If `Boost::system` is used, then it's `ompl`'s responsibility to export it. In those include .h files, `program_options` are not used.

Got it, I was accidentally linking `boost::program_options` in the wrong spot. It's fixed on my branch in f140fd1f6fad732f38a7a48a4725384a62e1bebb. 

Want to let me know if my branch also solves your issue?

felixf4xu:
> Want to let me know if my branch also solves your issue?

My case is simple: to compile `nav2_smac_planner` (https://github.com/ros-planning/navigation2/tree/main/nav2_smac_planner).

zkingston:
Does the now merged #1067 address this issue? If so, will close.

Ryanf55:
> Does the now merged #1067 address this issue? If so, will close.

Yep! I believe this can be closed now.

PR #1089, opened 2023-08-14
Allow allocating custom state sampler in rejection sampler

marlinstrub:
The rejection sampler currently uses the default state sampler of the state space as base sampler even if the user provided the space with a custom sampler via `StateSpace::setStateSamplerAllocator`. I think it would be more intuitive for the rejection sampler to only fall back to the default state sampler if the user didn't provide a custom one. This would also make it easier to get BIT* et al. to use a custom (uninformed) sampler. Does anyone have any concerns with the proposed change? @gammell?

### DIFF
--- a/src/ompl/base/samplers/informed/src/RejectionInfSampler.cpp
+++ b/src/ompl/base/samplers/informed/src/RejectionInfSampler.cpp
@@ -46,7 +46,7 @@ namespace ompl
           : InformedSampler(probDefn, maxNumberCalls)
         {
             // Create the basic sampler
-            baseSampler_ = InformedSampler::space_->allocDefaultStateSampler();
+            baseSampler_ = InformedSampler::space_->allocStateSampler();
 
             // Warn if a cost-to-go heuristic is not defined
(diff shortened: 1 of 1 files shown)


gammell:
Thanks @marlinstrub! I'm travelling at the moment, but this interface has never worked as I intended, so am happy to look at your improvements when I have a chance in the coming weeks.

zkingston (approved):
LGTM - @gammell & @marlinstrub. Tested and looks good.

gammell:
This is still in my queue. I will be able to take a look soon.

gammell:
I had no idea it was as simple as this! :laughing: That's both this specific PR and the desired API in the first place. Very embarrassed now that it took me so long, I should have understood it was a one-liner and moved faster.

I hope everything else on my list today will be like this. :smile:

gammell (approved):

zkingston:
Nope, just squash and merge. Thanks for taking a look both @marlinstrub and @gammell!

PR #1100, opened 2023-09-22
Python bindings for ARM and M1 Mac

zkingston:
WIP PR for Python bindings on arm64/aarch64 platforms, primarily addressing issues M1 Mac users have had with compiling the Python bindings. I have gotten the bindings to generate and compile on an M1 Mac Mini with some changes to:
- CastXML invocation, to include the default system includes that are missing. This uses the default CastXML available from homebrew.
- Remove Eigen vectorization in the binding generation (not during the actual compilation) to deal with some issues with Neon intrinsics.
- Variety of changes to the bindings to handle platform differences with strings and declarations.
- A mysterious load-bearing include in `ompl_py_tools.h` that otherwise results in a compilation error in CastXML with including `boost/python.hpp` - there probably is some include order shenanigans going on, but I didn't want to spend time to figure out the precise file necessary to fix.

Marking this PR as WIP until I have confirmation of it working on other setups and machines.

### DIFF
--- a/CMakeModules/Findcastxml.cmake
+++ b/CMakeModules/Findcastxml.cmake
@@ -20,11 +20,11 @@ if (CASTXML)
     endif()
 
     # workaround for problem between Xcode and castxml on Mojave
-    if (APPLE AND CMAKE_CXX_COMPILER MATCHES "/Applications/Xcode.app/Contents/Developer/Toolchains/.*")
-        set(CASTXMLCOMPILER_PATH "/usr/bin/clang++")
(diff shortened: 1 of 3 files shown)


zkingston:
Note: you will also need to update `pyplusplus`: https://github.com/ompl/pyplusplus/pull/1

wbthomason:
Note that this PR may also be necessary for x86 systems - it fixed an error I was encountering, probably due to a new version of `clang`.

PR #1101, opened 2023-09-24
Use full install paths for pkg-config

hacker1024:
According to the [CMake documentation](https://cmake.org/cmake/help/latest/module/GNUInstallDirs.html), there are some special cases where `CMAKE_INSTALL_<dir>` is not the same as `${prefix}/@CMAKE_INSTALL_<dir>`.

Such a case is where `CMAKE_INSTALL_<dir>` is an absolute path, which is common on NixOS: https://github.com/NixOS/nixpkgs/issues/144170


lopsided98:
Thank you for working on this; I've been meaning to fix this for a long time.

Unfortunately, the correct fix is a bit more complicated than this. pkg-config files are supposed to use the `${prefix}` variable where possible, so it is not enough to just use `CMAKE_INSTALL_FULL_*DIR`. See here for more info: https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files 

On the other hand, I think the solution proposed on that page is over-complicated, particularly if you can't use the CMake 3.20 `cmake_path(APPEND)` version. I prefer an approach along these lines (from memory, not tested, may need tweaking):
```
string(REPLACE "${CMAKE_INSTALL_PREFIX}" \${prefix} PKG_LIBDIR "${CMAKE_INSTALL_FULL_LIBDIR}")
```
Then you use `@PKG_LIBDIR@` in the pkg-config file.

hacker1024:
I see. Is such flexibility really necessary here, though? The linked page links to a blog post that uses custom GIO modules as an example where mirroring the directory structure of another project is needed. In this case, however, only libraries and headers are installed - when would you ever need to do something like that for those?

lopsided98:
I guess not. Variables are also required to make the package relocatable, but that requires other changes and is not currently supported by this pkg-config file.

lopsided98 (approved):

zkingston (approved):
LGTM, thanks for the contribution!

lepalom:
I think that this change is bad resolved. The result is:
```
includedir=/usr/local/include
```
when, it should be:
```
includedir=/usr/local/include/ompl-2.0
```
As, by default, OMPL_VERSIONED_INSTALL in ON, and then:
set(CMAKE_INSTALL_INCLUDEDIR "include/ompl-${PROJECT_VERSION_MAJOR}.${PROJECT_VERSION_MINOR}")

Maybe it is a question of where is set. Maybe it should be defined before:
include(GNUInstallDirs)
In any case, now it generates a wrong pkg-config file.

PR #1103, opened 2023-10-22
Add support for building Python wheels for Linux & MacOS

kylc:
This PR adds support for building Python wheels for the OMPL package, along with Github CI workflows to do so automatically. The result is that users can install the OMPL Python bindings, with precompiled OMPL shared libraries and all dependencies, like this:

`pip install https://github.com/ompl/ompl/releases/download/prerelease/ompl-1.6.0-cp310-cp310-manylinux_2_28_x86_64.whl`

Along with just about any CPython version [from the cibuildwheel table](https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip), supported build targets are:
- Linux x86_64 (via `manylinux_2_28_x86_64`)
- MacOS (both `x86_64` and `arm64`)

This is done using the [cibuildwheel](https://cibuildwheel.readthedocs.io/en/stable/) package to automate building for various operating systems, CPU architectures, and Python versions. On Linux, the official [manylinux](https://github.com/pypa/manylinux) Docker images are used to ensure glibc backwards compatibility in accordance with standard [PyPA packaging rules](https://peps.python.org/pep-0600/).

## Usage

Users can build wheels locally by installing `cibuildwheel` and invoking it from the `ompl/` root directory like this:

    CIBW_BUILD="cp312-manylinux_x86_64" cibuildwheel --platform linux py-bindings/

Correct format of the `CIBW_BUILD` variable can be found in the [cibuildwheel docs](https://cibuildwheel.readthedocs.io/en/stable/options/#build-skip). If cross-compiling on MacOS, you must also set `OMPL_BUILD_ARCH` to the target architecture, e.g. `arm64`.

**Note that `cibuildwheel` will make changes to your local system on MacOS. On Linux it uses Docker to containerize the build, but on MacOS the commands are performed on the host system. Be careful!**

## Changes

**NOTE: I don't have access to a MacOS development machine, so it's entirely possible that some of the Mac-specific changes below aren't required. Additional testing on that platform would be appreciated.**

- `py-bindings/setup.py` is modified to invoke CMake and build the OMPL shared libraries. This allows the `cibuildwheel` tool (or users) to `pip install .` to build locally.
- Github workflow files are added under `.github` to automatically run wheel build jobs on new commits.
- `CMakeLists.txt` changes:
  - Custom shared library outputs paths are removed to make it easier for `setup.py` to find the resulting shared libraries
  - We remove direct linking to `${PYTHON_LIBRARIES}` in accordance with [PEP513 recommendations](https://peps.python.org/pep-0513/#libpythonx-y-so-1); it will be loaded at runtime anyway.
  - When building on Mac, we pass extra linker arguments `-undefined suppress -flat_namespace` to prevent the linker complaints about missing Python symbols (due to the above).
  - When building on Mac, we define `_LIBCPP_ENABLE_CXX17_REMOVED_UNARY_BINARY_FUNCTION=1` to bring back some deprecated stdlib types which are removed in recent XCode versions (see [Xcode 15 Release Notes](https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes)).

## Known Issues

- CRITICAL: On MacOS, there seems to be some issue with multiple definitions of the same symbol. This can be observed by importing multiple OMPL namespaces and watching for a warning to the effect, e.g. `import ompl.base; import ompl.util`. These symbols aren't compatible with each other even though it is plain to see that they refer to the same thing. This breaks some Python code with cryptic errors like "expected argument LogLevel, got LogLevel".

  `RuntimeWarning: to-Python converter for ompl::msg::LogLevel already registered; second conversion method ignored.`

## Future work

- Reformat some of what is written above into a documentation page in the OMPL repository.
- Submit to PyPI so that users can simply `pip install ompl` (or similar).
- Build `universal2` wheels for MacOS, which bundle both x86_64 and arm64 compatible binaries into a single wheel.
- Bundle optional dependencies like FLANN, if worthwhile?
- Windows builds?

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -24,10 +24,9 @@ jobs:
         env:
           CIBW_ARCHS_MACOS: ${{ matrix.arch }}
           OMPL_BUILD_ARCH: ${{ matrix.arch }}
-          CIBW_BUILD: cp310-macosx_{x86_64,arm64} cp310-manylinux_x86_64
           # NOTE: Many combinations of OS, arch, and Python version can be built
           # depending on your patience. For example:
-          # CIBW_BUILD: cp3{10,11,12}-macosx_{x86_64,arm64} cp3{9,10,11,12}-manylinux_x86_64
+          CIBW_BUILD: cp3{10,11,12}-macosx_{x86_64,arm64} cp3{7,8,9,10,11,12}-manylinux_x86_64
           CIBW_BUILD_VERBOSITY: 1
       - uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -1,0 +1,68 @@
+name: Build Wheels
+on: push
+
+jobs:
+  build_wheels:
+    name: Build wheels on ${{ matrix.os }}-${{ matrix.arch }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-22.04, macos-13]
+        arch: [x86_64]
+        include:
+          - os: macos-13
+            arch: arm64
+    steps:
+      - name: Check out repository code
+        uses: actions/checkout@v4
+        with:
+          submodules: 'recursive'
+      - name: Build wheels
+        uses: pypa/cibuildwheel@v2.16.1
+        with:
+          package-dir: py-bindings
+        env:
+          CIBW_ARCHS_MACOS: ${{ matrix.arch }}
+          OMPL_BUILD_ARCH: ${{ matrix.arch }}
+          CIBW_BUILD: cp310-macosx_{x86_64,arm64} cp310-manylinux_x86_64
+          # NOTE: Many combinations of OS, arch, and Python version can be built
+          # depending on your patience. For example:
+          # CIBW_BUILD: cp3{10,11,12}-macosx_{x86_64,arm64} cp3{9,10,11,12}-manylinux_x86_64
+          CIBW_BUILD_VERBOSITY: 1
+      - uses: actions/upload-artifact@v3
+        with:
+          name: wheels
+          path: wheelhouse
+
+  pre-release:
+    name: Pre Release
+    concurrency:
+      group: push-${{ github.ref_name }}-prerelease
+      cancel-in-progress: true
+    needs: [build_wheels]
+    runs-on: "ubuntu-latest"
+    steps:
+      - uses: dev-drprasad/delete-tag-and-release@v0.2.1
+        with:
+          tag_name: prerelease
+          delete_release: true
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - uses: actions/download-artifact@v3
+        with:
+          name: wheels
+          path: wheelhouse
+
+      # Create the actual prerelease
+      # https://github.com/ncipollo/release-action
+      - name: GitHub Release
+        uses: ncipollo/release-action@v1.12.0
+        with:
+          prerelease: true
+          tag: "prerelease"
+          name: "Development Build"
+          allowUpdates: true
+          removeArtifacts: true
+          replacesArtifacts: true
+          artifacts: "wheelhouse/*"
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -34,20 +34,20 @@ jobs:
           name: wheels
           path: wheelhouse
 
-  pre-release:
-    name: Pre Release
+  prerelease:
+    if: github.ref == 'refs/heads/main'
+    runs-on: ubuntu-latest
     concurrency:
       group: push-${{ github.ref_name }}-prerelease
       cancel-in-progress: true
     needs: [build_wheels]
-    runs-on: "ubuntu-latest"
     steps:
       - uses: actions/download-artifact@v3
         with:
           name: wheels
           path: wheelhouse
 
-      - name: Publish artifacts
+      - name: GitHub release
         uses: ncipollo/release-action@v1.12.0
         with:
           prerelease: true
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -42,21 +42,12 @@ jobs:
     needs: [build_wheels]
     runs-on: "ubuntu-latest"
     steps:
-      - uses: dev-drprasad/delete-tag-and-release@v0.2.1
-        with:
-          tag_name: prerelease
-          delete_release: true
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-
       - uses: actions/download-artifact@v3
         with:
           name: wheels
           path: wheelhouse
 
-      # Create the actual prerelease
-      # https://github.com/ncipollo/release-action
-      - name: GitHub Release
+      - name: Publish artifacts
         uses: ncipollo/release-action@v1.12.0
         with:
           prerelease: true
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,9 +23,6 @@ include(FeatureSummary)
 include(CompilerSettings)
 include(OMPLUtils)
 
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin")
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
-
 set(OMPL_CMAKE_UTIL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules"
     CACHE FILEPATH "Path to directory with auxiliary CMake scripts for OMPL")
 set(OMPL_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/src;${CMAKE_CURRENT_BINARY_DIR}/src")
--- a/py-bindings/ompl/base/__init__.py
+++ b/py-bindings/ompl/base/__init__.py
@@ -1,2 +1,1 @@
-from ompl import util
 from ompl.base._base import *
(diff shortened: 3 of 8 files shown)


.github/workflows/wheels.yaml
```diff
@@ -0,0 +1,68 @@
...
+        uses: pypa/cibuildwheel@v2.16.1
+        with:
+          package-dir: py-bindings
+        env:
+          CIBW_ARCHS_MACOS: ${{ matrix.arch }}
+          OMPL_BUILD_ARCH: ${{ matrix.arch }}
+          CIBW_BUILD: cp310-macosx_{x86_64,arm64} cp310-manylinux_x86_64
+          # NOTE: Many combinations of OS, arch, and Python version can be built
+          # depending on your patience. For example:
+          # CIBW_BUILD: cp3{10,11,12}-macosx_{x86_64,arm64} cp3{9,10,11,12}-manylinux_x86_64
```
kylc, line 30: Which combinations should we build for in CI? It's only a matter of patience and not exceeding Github's CI job timeout.

Maybe:
- Linux cp3{10,11,12} to support Ubuntu 22.04+
- Mac OS cp12 to match the latest Python version in Brew

CMakeLists.txt
```diff
@@ -24,9 +24,6 @@ include(FeatureSummary)
 include(CompilerSettings)
 include(OMPLUtils)
 
-set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin")
-set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib")
-
```
kylc, line 29: This might break some existing expectations, but it also makes it easier for the `setup.py` script to find the resulting binaries.

https://github.com/ompl/ompl/pull/1103/files#diff-60f7d754a0c159e4789a1ac359dc6552c6187ec903af6e6c3af43232be7449a8R102-R104

py-bindings/pyproject.toml
```diff
@@ -0,0 +1,21 @@
+[build-system]
+requires = [
+    "setuptools>=42",
+    "cmake>=3.18",
+    "ninja",
+    "pygccxml",
+    "numpy",
+    # Need latest commit for this PR with Mac fixes:
+    # https://github.com/ompl/pyplusplus/pull/1
+    "pyplusplus @ git+https://github.com/ompl/pyplusplus",
```
kylc, line 10: This URL can go away if a new pyplusplus release is made to PyPI.

py-bindings/setup.py
```diff
@@ -1,13 +1,168 @@
...
+            # Multi-config generators have a different way to specify configs
+            if not single_config:
+                cmake_args += [
+                    f"-DCMAKE_LIBRARY_OUTPUT_DIRECTORY_{cfg.upper()}={extdir}"
+                ]
+                build_args += ["--config", cfg]
+
+        if sys.platform.startswith("darwin"):
+            # TODO: Move these out to configuration
+            cmake_args += ["-DCMAKE_CXX_COMPILER=/usr/local/opt/llvm@16/bin/clang++"]
```
kylc, line 109: I don't have a Mac machine to test changing this, but this path probably shouldn't be hardcoded. Maybe someone can experiment with setting this via an environmental variable in `before_build.sh`? Maybe `brew` provides a method for identifying this path?

py-bindings/setup.py
```diff
@@ -1,13 +1,168 @@
...
+            cmake_args += ["-DCMAKE_OSX_DEPLOYMENT_TARGET=13.0"]
```
kylc, line 110: Not sure how to correctly set this value.

py-bindings/ompl/base/__init__.py
```diff
@@ -1,2 +1 @@
-from ompl import util
```
kylc, line 1: This change "avoids" the major issue on Mac I mention in the PR description, but doesn't solve it if the user imports `ompl.util` or other packages themselves.

I didn't notice any negative side effects from removing this.

kylc (commented):
Pre-review

.github/workflows/wheels.yaml
wbthomason: Maybe also Linux cp39? Some Noetic environments still use Python 3.9 iirc.

zkingston: cp38 and I believe ISAAC Sim is tied to cp37 may be worthwhile targets... but that may be pushing the amount of time we have on the CI job.

py-bindings/setup.py
zkingston: I have a Mac to test against, I'll see what needs to be changed here.

zkingston:
Thanks a ton for making a PR for this! I would like to make sure #1100 gets merged first as I believe it fixes some of the issues for MacOS and modern Linux systems.

kylc:
Thanks for testing it out @wbthomason, apologies for not noticing your post earlier. It seems like there is some odd behavior depending on the ordering of imports. For example, if I `import ompl.tools` in a fresh interpreter then I get an error like yours, but if I first `import ompl.util; import ompl.tools` then it runs fine.

Removing the import change [here](https://github.com/ompl/ompl/pull/1103/files/9c3d3d92f5efab538f80d46487849d9f11792833#diff-807ae1414f6c6420b1e3a8ec42e1dfe866e3534980ba05e2df70fa65d42be67eL1
) seems to resolve the issue for me, but I thought it was causing some issues on MacOS relating to py-converters already being registered. For now I've reverted this change and I guess we'll see.

If you want to test updated wheels then you can find them on my OMPL fork after CI finishes running: https://github.com/kylc/ompl/tree/build-wheels

Regarding the libboost shared library issue, can you provide some more details about your Linux setup? Did you build the wheels locally or install from the CI job? It seems odd that it's looking for 1.78 since we build a much more recent version and link to that in the CI job. Maybe you can unzip the whl you installed and see which libboost shared libraries are bundled in there?

For example, if I unzip the latest build, I see that `ompl.libs/libboost_serialization-15a60f1e.so.1.83.0` exists and `ldd` shows that OMPL is linking directly to that file.

wbthomason:
I unfortunately deleted the actual `.whl` file I was testing with before, but I'll try to reproduce the issue in the next few days. In the meanwhile:
- I'm on Arch Linux, using Python 3.11.5
- I installed the wheel from the CI job on the day I filed the issue, using `pip` 23.3.1

I'll try re-installing and (assuming the issue repros) unzipping/checking with `ldd` later this week.

wbthomason:
I did a bit more digging here - I found the original `.whl` file. The errors in https://github.com/kylc/ompl-wheels/issues/1 may have been red herrings - I had another version of OMPL installed on my system, and this was found first by the call `dll_loader('ompl', dirname(abspath(__file__)))` in `ompl/__init__.py`, line 17. This old version of OMPL linked to the old, missing version of Boost.

So, there's still a bug with the library loading, but I don't think it's related to this PR - at the least, I definitely don't think that it should block this.

I'm closing https://github.com/kylc/ompl-wheels/issues/1 for now; feel free to re-open if you think that this is a problem that should be addressed by the wheel build process/in order for there to be wheels.

py-bindings/setup.py
mamoll: This is typically set to the lowest supported OS version. The host OS needs to have the SDK for that version. Do you need to set this at all?

py-bindings/pyproject.toml
zkingston: A new pyplusplus tag has been made, 1.8.6

.github/workflows/wheels.yaml
kylc: cp37-cp312 might be possible. I think the default Github CI limit is 6hrs. I'll hold off until we're done with reviews and then test it out in CI to avoid making the builds take forever during the review process.

py-bindings/pyproject.toml
kylc: PyPI is still showing 1.8.5. I think the release needs to be made there before updating this line.

py-bindings/setup.py
kylc: Without it I was seeing lots of `dylib was built for newer macOS version than being linked` warnings on the macos-13 Github runner. Maybe it's OK to leave it as-is until someone has a chance to verify that it does indeed build with a lower version.

zkingston:
Before finishing this PR, I'd like to get these issues with building on Ubuntu 20/22 fixed, will require a modest reversion of some of the CMake logic to pull in the old versions of libraries:
- https://github.com/ompl/ompl/issues/1107
- https://github.com/ompl/ompl/issues/1105

It would also be nice to get the PR which fixes compilation for modern builds/Apple M-series chips in as well: https://github.com/ompl/ompl/pull/1100, but that can wait until after.

zkingston:
Sorry about the kerfuffle, didn't realize this was set to use the M1 branch as base - that's merged now after some additional testing.

zkingston:
@kylc @wbthomason @mamoll are there any outstanding comments on this pull? If not, let's rebase and merge in.

kylc:
The build is currently broken on `main`, but is fixed by https://github.com/ompl/ompl/pull/1119. For now I've rebased onto that PR and we should merge it first. Otherwise, I think the only items are:
- Enable Github Actions on the repo and kick off a job to make sure everything is up and running
- Someone with Apple M1 hardware should take a look at the issue in the original PR description

I just pushed up a few changes to only upload wheels for builds on the main branch, fix some issues with the Brew steps on MacOS, and specify pygccxml==2.2.1.

kylc:
Great! I rebased and enabled builds for all the Python versions we discussed in the comments above. Ship it!

.github/workflows/wheels.yaml
kylc: Yep, I enabled for: `cp3{10,11,12}-macosx_{x86_64,arm64} cp3{7,8,9,10,11,12}-manylinux_x86_64`. Seems to only take about 2hrs on CI, so plenty of time to spare.

simonschmeisser:
All of them seem to fail since about a month, do you have any idea why?

kylc:
Hey @simonschmeisser. I've noticed that and have a fix sitting on my fork at https://github.com/kylc/ompl, but haven't had time to open a PR. I'll get around to it this weekend I think.

PR #1106, opened 2023-10-30
Adding RRT-Rope planner

lopetit:
@mamoll @zkingston could you please review this pull request? Thank you!

### DIFF
--- a/doc/markdown/planners.md
+++ b/doc/markdown/planners.md
@@ -32,6 +32,8 @@ Planners in this category only accounts for the geometric and kinematic constrai
     This is one of the first single query planners. The algorithm is easy to understand and easy to implement. Many, many variants of RRT have been proposed. OMPL contains several RRT variants:
     - [RRT Connect (RRTConnect)](\ref gRRTC)<br>
       This planner is a bidirectional version of RRT (i.e., it grows two trees). It usually outperforms the original RRT algorithm.
+    - [RRT Rope (RRTRope)](\ref gRRTRope)<br>
+      This planner uses an altered version of RRT-connect (without infinitesimal step size and with intermediate nodes), and a deterministic shortening technique once a feasible path is found. It usually outperforms other algorithms in large uncluttered environments in R3 state space.
     - [RRT*](\ref gRRTstar)<br>
       An asymptotically optimal version of RRT: the algorithm converges on the optimal path as a function of time. This was the first provably asymptotically planner (together with PRM). Since its publication, several other algorithms have appeared that improve on RRT*'s convergence rate, such as [RRT#](\ref gRRTsharp) and [RRTX](\ref gRRTXstatic).
     - [Lower Bound Tree RRT (LBTRRT)](\ref gLBTRRT)<br>
--- a/py-bindings/headers_geometric.txt
+++ b/py-bindings/headers_geometric.txt
@@ -18,6 +18,7 @@ src/ompl/geometric/planners/kpiece/LBKPIECE1.h
 src/ompl/geometric/planners/pdst/PDST.h
 src/ompl/geometric/planners/rrt/RRT.h
 src/ompl/geometric/planners/rrt/RRTConnect.h
+src/ompl/geometric/planners/rrt/RRTRope.h
 src/ompl/geometric/planners/rrt/LazyRRT.h
 src/ompl/geometric/planners/rrt/TRRT.h
 src/ompl/geometric/planners/rrt/RRTstar.h
--- a/src/ompl/geometric/PathGeometric.h
+++ b/src/ompl/geometric/PathGeometric.h
@@ -38,6 +38,7 @@
 #define OMPL_GEOMETRIC_PATH_GEOMETRIC_
 
 #include "ompl/base/SpaceInformation.h"
+#include <ompl/base/spaces/RealVectorStateSpace.h>
 #include "ompl/base/Path.h"
 #include <vector>
 #include <utility>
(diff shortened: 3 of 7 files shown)


lopetit:
Hi @gammell by any chance would you have the time to go through this pull request? Thank you!

zkingston:
As far as I can tell, this planner is `RRTConnect` but with the addition of a shortcutting heuristic after a path is found. This does not merit a new planner so I will close this PR, but may merit a new addition to the `PathSimplifier` class - if you show it can produce higher quality paths than the current heuristics.

lopetit:
Hi Zachary, thanks for taking the time to look at this PR. Indeed, RRT-Rope is based on an alternative of RRT-Connect with 2 differences: 
1. the generation of [q_{new}](https://msl.cs.uiuc.edu/~lavalle/cs497_2001/book/iplan/img83.gif) without epsilon (without the check performed [here](https://ompl.kavrakilab.org/RRTConnect_8cpp_source.html#l00128)), 
2. the addition of intermediate nodes on each branch, which is already implemented [here](https://ompl.kavrakilab.org/RRTConnect_8cpp_source.html#l00148) if `addIntermediateStates` is true. 
So I kept the planner separate because I was thinking of implementing feature 1 later (which I could do quickly now that I've found that part in the code). To be honest, I was hesitant to implement the algorithm in a separate class or in the `RRTConnect` class and add a `useRopeShortcut` variable, similar to `addIntermediateStates`. 

I took the time to look at the `PathSimplifier` class, and particularly the `shortcutPath` [function](https://ompl.kavrakilab.org/classompl_1_1geometric_1_1PathSimplifier.html#a911bdadf0c84acdae69d003b99f8d48e). From what I understand, the current heuristic is based on partial-shortcut, which iteratively shortcuts between random points on the path, and adds points on the newly created branches. As shown in our [paper](https://www.researchgate.net/publication/357636884_RRT-Rope_A_deterministic_shortening_approach_for_fast_near-optimal_path_planning_in_large-scale_uncluttered_3D_environments) (SMC’21 best student paper) - [Fig. 7](https://ieeexplore.ieee.org/document/9659071?figureId=fig7#fig7), RRT-Rope's deterministic shortening method performs better in computation time, and better or equal in path length, in targeted environments (large and sparse). So I get your point of adding it to the `PathSimplifier` class. **What would be the most appropriate way to do this**; a new `ropeShortcutPath` function, or replace the current `shortcutPath` function, or create 2 functions `ropeShortcutPath` and `partialShortcutPath`, with the `shortcutPath` function calling one or the other? 


We've also shown in [Fig. 6](https://ieeexplore.ieee.org/document/9659071?figureId=fig6#fig6) that, in the targeted environments, a shorter path before shortening doesn't lead to a shorter path after shortening. It's simply a matter of finding a feasible path as quickly as possible and shortening it. Any attempt at online optimization will not reduce the path length in those environments. Plus, the computational effort devoted to Informed RRT* or similar online optimizations slows down the overall path computation time. **Given this new information, do you agree that the full algorithm (including the modified RRT-connect part) is also of interest?** However, I could move the `shortcutPath` function to the `PathSimplifier` class and call it in the `RRTRope` or `RRTConnect` class, so that other algorithms could use just the shortening part.

zkingston:
I would prefer the addition of a `ropeShortcutPath` function in `PathSimplifier`, with the potential modification of adding `ropeShortcutPath` to the battery of heuristics used by default in the high-level `simplify` method - your heuristic seems to work better in large and sparse environments, which is not everyone's use-case. Having it be optional is essential.

Whatever the case, the shortcutting heuristic shouldn't just be inside of the planner. Also to your last point, yes, see [AnytimePathShortening](https://ompl.kavrakilab.org/classompl_1_1geometric_1_1AnytimePathShortening.html#gAPS) for an abstract wrapper around the concept of finding feasible paths and simplifying/hybridizing them together. You probably could get similar behavior by adding your rope heuristic to APS.

lopetit:
I've added the `ropeShortcutPath` function to `PathSimplifier`. I've compared it with the original heuristic (Partial-Shortcut), in a single scenario so far, as I'm not very familiar with OMPL test environments yet. The results are quite similar to those in our paper, in that 
- if the number of Partial-Shortcut steps is _high_, the Rope shortcut produces an _equal or shorter path_ in a _shorter computation time_.
- if the number of Partial-Shortcut steps is _low_, the Rope shortcut produces a much _shorter path_.

I'd like to clarify one point about the targeted environments (large and sparse). RRT-Rope, as a whole (including the RRT-Connect part), has been designed to perform well in these environments, compared to optimality. However, for the shortcut part only, Rope shortcut always performs (statistically) better than Partial Shortcut, whatever the environment. Here are some screenshots of the results with the raw path in white (generated with RRT-Connect):
Screenshot from 2023-11-29 10-22-15
Screenshot from 2023-11-29 10-19-01
Screenshot from 2023-11-29 10-17-09
Considering these results, I think it would be better to change the original `shortcutPath` function name to `partialShortcut`, and have the `shortcutPath` function allow you to choose between `partialShortcut` and `ropeShortcut`. What do you think?

zkingston:
I'm happy to accept `ropeShortcut` as a new component of `PathSimplifier` and renaming `shortcutPath` to `partialShortcut` but do not want to add it as default behavior. While it performs well on this problem, there are time and performance trade-offs with this approach that may affect simplification in other domains (e.g., manipulators, molecules, other higher-dimensional problems) where collision checking is more computationally expensive. After further testing, changing `ropeShortcut` to be the default may be warranted.

Issue #1105, opened 2023-10-30
[CMake] yaml-cpp error


esatgundogdu:
I have tried to compile ompl from latest commit source files, but it gives cmake error. Then I also tried version 1.6.0 source files and everything was fine. I think there is an issue with commits after 1.6.0 version. 

Here what I tried to do and the results:
```bash
git clone https://github.com/ompl/ompl.git
cd ompl
mkdir -p build/Release
cd build/Release
cmake ../..
```
After this, I got this error:
```
CMake Error at demos/CMakeLists.txt:67 (add_library):
  add_library cannot create ALIAS target "yaml-cpp::yaml-cpp" because target
  "yaml-cpp" is imported but not globally visible.
```

My system is Ubuntu 20.04 and as I said before, ompl version 1.6.0 was successfully compiled.

captain-yoshi:
Same problem with identical OS.

PR #1111, opened 2023-11-11
Fix yaml-cpp target to be compatible with CMake <3.18

kylc:
Modifies the yaml-cpp CMake target to be IMPORTED_GLOBAL, which is required in order to define it as an ALIAS in CMake <3.18. However, setting this property requires CMake >=3.11 so the minimum version is bumped.

Test it yourself under an Ubuntu 20.04 container. `main` currently fails with the error described in https://github.com/ompl/ompl/issues/1105. This branch successfully compiles. Seems to have no adverse effects on 22.04.

```
docker run --mount type=bind,source="$(pwd)",target=/ompl --rm -it ubuntu:20.04

export DEBIAN_FRONTEND=noninteractive

apt-get update
apt-get -y install g++ cmake pkg-config libboost-serialization-dev libboost-filesystem-dev libboost-system-dev libboost-program-options-dev libboost-test-dev libeigen3-dev libode-dev wget libyaml-cpp-dev
mkdir -p /ompl/build/Release && cd /ompl/build/Release/
cmake ../.. && cmake --build .
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.11)
 project(ompl VERSION 1.6.0 LANGUAGES CXX)
 set(OMPL_ABI_VERSION 17)
 
--- a/demos/CMakeLists.txt
+++ b/demos/CMakeLists.txt
@@ -64,6 +64,12 @@ if (OMPL_BUILD_DEMOS)
         # https://github.com/jbeder/yaml-cpp/issues/1025
         # https://github.com/jbeder/yaml-cpp/pull/1196
         if(NOT TARGET yaml-cpp::yaml-cpp AND TARGET yaml-cpp)
+            # Aliasing local imported targets is only supported in CMake >=3.18.
+            # Promoting yaml-cpp to the global scope here prevents errors on
+            # older CMake versions.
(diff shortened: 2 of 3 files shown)


zkingston:
Bumping CMake to 3.11 may be problematic for those compiling on Ubuntu 18.04 (for ROS Melodic). Is there anyway around this without bumping the requirement? Also - please edit the main README with the updated requirement.

kylc:
After some testing, the actual minimum version I can compile with (before the ALIAS issues were introduced and without this PR, basically https://github.com/ompl/ompl/commit/fca10b4bd4840856c7a9f50d1ee2688ba77e25aa) is CMake 3.12, due at least to some use of [add_compile_definitions](https://cmake.org/cmake/help/latest/command/add_compile_definitions.html).

So to summarize:
- Before https://github.com/ompl/ompl/pull/1067: CMake >=3.12
- After https://github.com/ompl/ompl/pull/1067: CMake >=3.18
- With https://github.com/ompl/ompl/pull/1111: CMake >=3.12

I've updated the README and `cmake_minimum_version` to reflect the 3.12 requirement. I'm not sure how much work would be required to go back to CMake 3.10. It seems it hasn't been supported since at least https://github.com/ompl/ompl/pull/1013.

zkingston (approved):
LGTM, thank you for investigating.

PR #1119, opened 2023-11-29
Fix for ompl::geometric Python bindings

zkingston:
On some platforms, internal classes such as `Motion`, etc. are leaked into global namespaces due to their use in vectors, NNs, etc. - this pull filters these declarations out.


kylc:
Tested in CI (via https://github.com/ompl/ompl/pull/1103) and this fixes the pygccxml errors in the `geometric` module.

mamoll (approved):

PR #1120, opened 2023-11-30
Added Rope Shortcut

lopetit:
Hi @zkingston, I've made the changes we talked about in this [closed PR](https://github.com/ompl/ompl/pull/1106), i.e., adding Rope shortening heuristic in `PathSimplifier`, and renaming `shortcutPath` to `partialShortcutPath`, and keeping this as default in other functions (e.g., `simplify`). 

Regarding your last comment, I'd still argue that the Rope shortcut would outperform the Partial shortcut even in other domains where the collision checks might be computationnaly expensive, as the performance comparison results measured in computation time still hold if measured in number of collision checks instead. This is due to irrelevant shortcuts performed by the random approach (for either a portion of path that is already optimal, or for a portion of path that is intended to be optimized in the future). Checking the farthest nodes first guarantee you to not perform those irrelevant operations. One remaining point is that our deterministic shortcut approach has the "advantage" that the optimization ends at a given point, unlike partial-shortcut, which requires an exit condition. For an extremely costly collision check, as you mentionned, one might not want to shortcut the path to the near-optimum, as it might be expensive in calculations. But this can be done with the Rope shortcut by adding a timer as an exit condition. The output path quality would be statistically better than partial-shortcut, for the same computation time, as the farthest shortcuts are more likely to drastically improve the trajectory. I could quickly implement that if you see fit.


lopetit:
Hi @zkingston, any update about this?

src/ompl/base/src/OptimizationObjective.cpp
```diff
@@ -98,6 +103,11 @@ ompl::base::Cost ompl::base::OptimizationObjective::combineCosts(Cost c1, Cost c
     return Cost(c1.value() + c2.value());
 }
 
+ompl::base::Cost ompl::base::OptimizationObjective::subtractCosts(Cost c1, Cost c2) const
+{
+    return Cost(c2.value() - c1.value());
+}
+
```
zkingston, line 110: I dislike how the operators are in reverse order for this call - intuition would say that it would return `c1 - c2`.

src/ompl/base/src/OptimizationObjective.cpp
```diff
@@ -78,6 +78,11 @@ bool ompl::base::OptimizationObjective::isCostEquivalentTo(Cost c1, Cost c2) con
     return !isCostBetterThan(c1, c2) && !isCostBetterThan(c2, c1);
 }
 
+bool ompl::base::OptimizationObjective::isCostEquivalentTo(Cost c1, Cost c2, double threshold) const
+{
+    return isCostEquivalentTo(c1, c2) || (std::fabs(c1.value() - c2.value()) < threshold);
+}
+
```
zkingston, line 85: I believe the second condition should be something like `isCostBetterThan(subtractCosts(c1, c2), threshold)` where `threshold` is a cost, similar to the existing `OptimizationObjective` threshold?

src/ompl/geometric/src/PathSimplifier.cpp
```diff
@@ -184,7 +184,112 @@ bool ompl::geometric::PathSimplifier::reduceVertices(PathGeometric &path, unsign
...
+    std::vector<base::State *> &states = path.getStates();
+
+    // Loops through the path segments to add intermediate nodes seperated by a distance delta apart.
+    // This is done by linearly interpolating between the two nodes
+    for (unsigned int i = 0; i < states.size() - 1; ++i)
+    {
+        double dist = si->distance(states[i], states[i + 1]);
+        if (dist > delta)
+        {
+            int numIntermediateStates = (int)(floor(dist / delta));
```
zkingston, line 203: Use `std::size_t` for all indexing.

src/ompl/geometric/src/PathSimplifier.cpp
```diff
@@ -184,7 +184,112 @@ bool ompl::geometric::PathSimplifier::reduceVertices(PathGeometric &path, unsign
...
+    std::vector<base::Cost> costs(states.size(), obj_->identityCost());
+    for (unsigned int i = 1; i < costs.size(); ++i)
+    {
+        costs[i] = obj_->combineCosts(costs[i - 1], obj_->motionCost(states[i - 1], states[i]));
+    }
+
+    bool result = false;
+
+    // Loops through the path nodes to perform shortcutting, considering the farthest nodes first.
+    for(unsigned int i = 0; i < states.size() - 2; ++i)
```
zkingston, line 226: Run formatting - there should be a space `for (`

zkingston: Also, use either `auto i = OU` or `std::size_t i = 0`

src/ompl/geometric/src/PathSimplifier.cpp
```diff
@@ -184,7 +184,112 @@ bool ompl::geometric::PathSimplifier::reduceVertices(PathGeometric &path, unsign
...
+        for (unsigned int j = states.size() - 1; j > i + 1; --j)
+        {
+            // Check if the shortcut is valid
+            if (si->checkMotion(states[i], states[j]))
+            {
+                base::Cost shortcutCost = obj_->motionCost(states[i], states[j]);
+                base::Cost alongPath = obj_->subtractCosts(costs[i], costs[j]);
+
+                // Check if the path segment i-j is already optimal and break out of j loop if it is.
+                if (obj_->isCostEquivalentTo(alongPath, shortcutCost, 0.1 * delta))
```
zkingston, line 238: `0.1` should be a parameter of the algorithm

zkingston (changes requested):
Some minor changes requested.

lopetit:
Thank you for the review @zkingston. I implemented the minor changes. Let me know if you have other comments!

src/ompl/geometric/src/PathSimplifier.cpp
```diff
@@ -195,14 +195,14 @@ bool ompl::geometric::PathSimplifier::ropeShortcutPath(PathGeometric &path, doub
...
     // Loops through the path segments to add intermediate nodes seperated by a distance delta apart.
     // This is done by linearly interpolating between the two nodes
-    for (unsigned int i = 0; i < states.size() - 1; ++i)
+    for (std::size_t i = 0; i < states.size() - 1; ++i)
     {
         double dist = si->distance(states[i], states[i + 1]);
         if (dist > delta)
         {
-            int numIntermediateStates = (int)(floor(dist / delta));
+            std::size_t numIntermediateStates = (int)(floor(dist / delta));
```
zkingston, line 203: should use `static_cast<std::size_t>()`

src/ompl/geometric/src/PathSimplifier.cpp
```diff
@@ -215,27 +215,29 @@ bool ompl::geometric::PathSimplifier::ropeShortcutPath(PathGeometric &path, doub
...
     std::vector<base::Cost> costs(states.size(), obj_->identityCost());
-    for (unsigned int i = 1; i < costs.size(); ++i)
+    for (std::size_t i = 1; i < costs.size(); ++i)
     {
         costs[i] = obj_->combineCosts(costs[i - 1], obj_->motionCost(states[i - 1], states[i]));
     }
 
     bool result = false;
+    ompl::base::Cost equivalenceCost(equivalenceTolerance * delta);
+    obj_->setCostThreshold(equivalenceCost);
```
zkingston, line 225: I like how you've done the equivalence cost now, but I don't think you should modify the optimization objective, as this could affect planner behavior outside of simplification.

zkingston (changes requested):
Two small changes and then we are ready to go.

zkingston (approved):
LGTM - please run `clang-format` and then this can be merged.

lopetit:
@zkingston Thanks for the approval. `clang-format` done.

PR #1125, opened 2023-12-12
PyPI upload for OMPL wheels

### LINKED ISSUE #1121: Segmentation Fault in Python using Development Build Wheels with existing OMPL installation
Hello developers,
first things first: thank you for the amazing work with OMPL and thank you for start publishing the python wheels without the need to compile all the source code. I am truly grateful.

I know that the wheels are still in early development, but I've tested them and think I should report the issue that I am encountering. Both using python 3.10 and 3.11, following the instructions in https://github.com/ompl/ompl/pull/1103, I always receive a segmentation fault when running https://ompl.kavrakilab.org/RigidBodyPlanning_8py_source.html and https://ompl.kavrakilab.org/StateSampling_8py_source.html .

Let me know if I can help with something on the python side!

wbthomason:
Thanks to @kylc's work in #1103, we now have consistently building OMPL Python wheels. The next step for `pip`-installable OMPL, I think, is to get these wheels automatically deploying to PyPI.

This PR adapts the basic upload action example from the `cibuildwheel` docs for this purpose. It seems pretty straightforward, but I'm not very experienced with GH Actions, so we should verify that I've not missed anything.

There are two (that I know of) outlying tasks:
- [ ] We need to set up an OMPL project on PyPI following: https://docs.pypi.org/trusted-publishers/creating-a-project-through-oidc/
- [ ] We need to resolve #1121, or at least add a warning that this package is known to conflict with system-level OMPL installations

Then, we should be able to cut a new release, and see the wheels upload.

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -33,6 +33,21 @@ jobs:
           name: wheels
           path: wheelhouse
 
+  upload_all:
+    needs: [build_wheels]
+    environment: pypi
+    permissions:
+      id-token: write
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release' && github.event.action == 'published'
+    steps:
+    - uses: actions/download-artifact@v3
+      with:
+        name: artifact
+        path: dist
+
+    - uses: pypa/gh-action-pypi-publish@release/v1
+
   prerelease:
     if: github.ref == 'refs/heads/main'
     runs-on: ubuntu-latest


kylc:
This is great, it will be very nice to `pip install ompl`!

I recently went through the PyPI process on another project, and I found a useful testing workflow is to set up an account on https://test.pypi.org/, configure the CI job to use the TestPyPI instance, and then make releases on a forked repo to see that the CI job runs successfully when it's supposed to and that wheels appear on TestPyPI.

It's useful to test everything out since there's a lot to juggle: PyPI secret keys, GitHub actions YAML, wheel files... Once everything is validated, change the workflow to point to the real PyPI instance and merge.

.github/workflows/wheels.yaml
```diff
@@ -33,6 +33,21 @@ jobs:
...
+    needs: [build_wheels]
+    environment: pypi
+    permissions:
+      id-token: write
+    runs-on: ubuntu-latest
+    if: github.event_name == 'release' && github.event.action == 'published'
+    steps:
+    - uses: actions/download-artifact@v3
+      with:
+        name: artifact
```
kylc, line 46: This should probably say `name: wheels` to match the artifact name on L33.

wbthomason: Good catch, thanks!

wbthomason:
@kylc Thanks! Starting with test.pypi.org seems prudent - @zkingston or @mamoll, do you think we could make an account there to start testing?

mamoll:
I think we can close this, since I already added this in separate commits.

PR #1134, opened 2024-01-24
Fix termination before planner termination condition is met.

taylanisleyici:
The current implementation of the multilevel planner's solve function terminates as soon as a valid path is found, rendering the user-defined termination parameter ineffective in many scenarios. For instance, when attempting to optimize a solution for a specific duration using the timedPlannerTerminationCondition, the planner terminates prematurely upon finding an initial solution.

To address this issue, the proposed pull request introduces a modification to the termination condition logic. The existing implementation uses an 'or' condition, causing the planner to terminate if either a valid path is found or the user-defined termination parameter is satisfied. The proposed fix adjusts this logic to use an 'and' condition, ensuring that the planner will only terminate when both conditions are met.

### DIFF
--- a/src/ompl/multilevel/datastructures/BundleSpaceSequenceImpl.h
+++ b/src/ompl/multilevel/datastructures/BundleSpaceSequenceImpl.h
@@ -190,10 +190,10 @@ ompl::multilevel::BundleSpaceSequence<T>::solve(const ompl::base::PlannerTermina
         if (priorityQueue_.size() <= currentBundleSpaceLevel_)
             priorityQueue_.push(kBundle);
 
-        ompl::base::PlannerTerminationCondition ptcOrSolutionFound(
-            [this, &ptc] { return ptc || foundKLevelSolution_; });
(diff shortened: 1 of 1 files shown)


aorthey:
Hi @taylanisleyici, thank you for your contribution. I remember changing this when working on QRRT*/QMP* to optimize, but this has not been added yet. 

The idea for the `ptcOrSolutionFound` function is that it terminates whenever you find a solution on one of the levels, such that we can progress to the next level in the sequence. The correct fix here should be that we terminate if either ptc is true OR if there is a solution on the k-th level EXCEPT if there is a solution on the last level (bundleSpaces_.size() - 1). Something like `return ptc || (foundKLevelSolution_ && k < bundleSpaces_.size() - 1);`. If we use the 'and' condition, we would only optimize on the lowest level and never progress to the next one. Also, if ptc is true, you want to exit immediately, not wait until a solution is found.

Hope this makes sense.

taylanisleyici:
Thanks, this fixed my case. I'm committing the solution in case you want to merge it.

aorthey (approved):
LGTM

PR #1136, opened 2024-01-26
Do not retry on invalid start/goal/unrecognized goal type

galou:
In AnytimePathShortening do not retry when the error is INVALID_START, INVALID_GOAL, or UNRECOGNIZED_GOAL_TYPE.


zkingston (approved):
LGTM, thanks for the contribution.

PR #1139, opened 2024-02-01
fix default value for goal bias fraction in EST planner

mamoll:
Fix a bug I introduced 7 years ago (!).


zkingston (approved):
Good catch!

PR #1147, opened 2024-03-04
More relevant return code

galou:
Some algorithms didn't make use of other return code than `SUCCESS` and `TIMEOUT` but we can return earlier if the start or the goal are invalid because this won't change.

I added some early-return code to some algorithms but have the impression that catching the `INVALID_GOAL` condition is more complicated than what I did.

### DIFF
--- a/src/ompl/control/planners/est/src/EST.cpp
+++ b/src/ompl/control/planners/est/src/EST.cpp
@@ -97,6 +97,12 @@ ompl::base::PlannerStatus ompl::control::EST::solve(const base::PlannerTerminati
     base::Goal *goal = pdef_->getGoal().get();
     auto *goal_s = dynamic_cast<base::GoalSampleableRegion *>(goal);
 
+    if (goal_s == nullptr)
+    {
+        OMPL_ERROR("%s: Unknown type of goal", getName().c_str());
+        return base::PlannerStatus::UNRECOGNIZED_GOAL_TYPE;
+    }
+
     // Initializing tree with start state(s)
     while (const base::State *st = pis_.nextStart())
     {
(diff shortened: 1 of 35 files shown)


zkingston (approved):
Wow, LGTM - thank you so much for this improvement!

joao-pm-santos96:
Hi @galou and @zkingston 
First of all, thank you for your amazing work!

I do have a question. Hasn't this pull request break some functionality? 

In earlier versions, I could use RRT (for example, KPIECE1 is another) to plan with a goal which inherited only from `Goal `or `GoalRegion`. After this update, I can no longer do it (I receive `base::PlannerStatus::UNRECOGNIZED_GOAL_TYPE`)... and I don't understand why... I mean, I don't understand why these two planners need it to have goals of type GoalSampleableRegion. I'll try to explain it as best as I can:

I do understand that the verification is done here https://github.com/ompl/ompl/blob/7f30de1871b09be2e08af0669a93809cefc1f602/src/ompl/control/planners/rrt/src/RRT.cpp#L97-L103

and also here https://github.com/ompl/ompl/blob/7f30de1871b09be2e08af0669a93809cefc1f602/src/ompl/control/planners/rrt/src/RRT.cpp#L119-L123

But, prior to these changes, we where allowed to not define a sampler and it would be automatically instantiated for us here 
https://github.com/ompl/ompl/blob/7f30de1871b09be2e08af0669a93809cefc1f602/src/ompl/geometric/planners/rrt/src/RRT.cpp#L128-L129

and the cycle would always step into the `else` branch of the if condition
https://github.com/ompl/ompl/blob/7f30de1871b09be2e08af0669a93809cefc1f602/src/ompl/control/planners/rrt/src/RRT.cpp#L144-L147

and, actually, this worked extremely well. 

Is this breaking change desired or is it just an unforeseen consequence? I really hope that it is the former. If so, I think I can create a pull request to help you, but would love to have your feedback first.


Thank you.

galou:
I guess that replacing l. 99 by`if ((goal_s == nullptr) && !sampler)` would do the trick. Sorry about breaking things.

joao-pm-santos96:
Well, actually, as far as I understand, at that point `sampler_` is still `nullptr` so I think that will not change anything.
Sorry to say this but as far as I've reviewed, these early exit clauses that where added, are breaking the behavior that seemingly is desired, across  all affected planners... I do think we should rever this... but maybe let's wait for @zkingston and see he's opinion on this ...

PR #1154, opened 2024-04-22
Fix syntax error to avoid unnecessary error message in bash script

Freyja923:
### Background
Noticed a minor typo in the bash script that triggers an unnecessary error message when installing OMPL.app due to missing space in the if statement.

### Changes Made
``` diff
- if [ -z $APP]; then
+ if [ -z $APP ]; then
```

### Impact Assessment
This change is minimal and does not affect the logic of the script. It purely enhances the script cleanliness and prevents misleading error messages during execution.


zkingston (approved):
LGTM, thanks!

PR #1159, opened 2024-05-03
Fix Python wheels missing constrained planning components

wbthomason:
I found a bug wherein, due to `numpy` not being installed correctly before building Boost in the Python wheel CI, `libboost_numpy` would not be built, and so part of the binding generator would silently fail, leading to some components - notably, the constrained planning modules - being missing from the pre-built wheels.

Note that the pip install seems to be necessary; the system-package install (for Linux, at least) places `numpy` in a location that Boost's build does not find.

This PR also adds some output to prevent silent failures like this from going unnoticed in the future and updates the versions of other CI actions used in building Python wheels, in compliance with the warning output visible on the last few runs of the wheel action.

- **Install numpy before building Boost in CI**
- **Make binding generation exceptions less silent**
- **Update versions of actions for Python wheel CI**


zkingston (approved):
LGTM - Thank you for investigating this really hairy bug.

PR #1164, opened 2024-05-18
BITstar: use atomic variable instead of lock in IdGenerator

Freyja923:
In the current implementation of `BITstar`, the `IdGenerator` class uses a lock to prevent data races. The same task can be accomplished more effectively with atomic variables, leading to cleaner code and a slight performance boost.

Please review my changes and provide any feedback. I am willing to make further adjustments or conduct additional testing if required.

Thank you for your time and consideration.

reference: https://en.cppreference.com/w/cpp/atomic/atomic/operator_arith

### DIFF
--- a/src/ompl/geometric/planners/informedtrees/bitstar/IdGenerator.h
+++ b/src/ompl/geometric/planners/informedtrees/bitstar/IdGenerator.h
@@ -42,7 +42,7 @@
 #include "ompl/geometric/planners/informedtrees/BITstar.h"
 
 #include <thread>
-#include <mutex>
+#include <atomic>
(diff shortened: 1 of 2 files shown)


saahu27 (approved):
This is a great optimization! The change from mutex-based locking to atomic variables is clean and appropriate for this use case.

The IdGenerator class now uses std::atomic<VertexId> which provides lock-free thread-safe incrementation. This simplifies the code and improves performance by:
- Eliminating mutex overhead
- Removing lock contention
- Using hardware-level atomic operations

The changes are minimal, focused, and well-implemented. This aligns with how AIT* already handles ID generation.

LGTM! ✅

PR #1171, opened 2024-07-02
chore: Add Ubuntu 24.04 to the list of OS versions tested in CI

wbthomason:
Ubuntu 24.04 is now offered as a runner in beta (see https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). As such, we should test the OMPL build against it. Once it's out of beta, there's another chore to update the wheel building workflow to use Ubuntu 24 instead of 22.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-22.04, ubuntu-20.04, macos-12, windows-2022]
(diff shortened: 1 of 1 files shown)


mamoll:
Looks good to me!

mamoll (approved):

mamoll:
@wbthomason the build of python wheels failed on 24.04. Have you tried to build them locally on 24.04?

wbthomason:
@mamoll I haven't, but can - this PR shouldn't have affected the wheels workflow, though. Did it, somehow?

PR #1177, opened 2024-09-05
fix strrt distance function: return inf when t1 > t2, and change the order of the arguments in distance function for goal tree.

kaizer-nurik:
in original paper of STRRT* distance function is defined as
image

As we see, if t1 > t2, than distance function returns infinity, but in SpaceTimeStateSpace if t1>t2, it just returns the absolute value of |t1-t2|. This leads to incorrect work of nearest neighbours, that returns invalid states. In this pull request i add a check of (t1>t2) in SpaceTimeStateSpace, so it can be identical to  the original distance function, that is defined in paper. 

Also, i changed order of arguments in distanceFunction for goal tree, because it grows backwards.

### DIFF
--- a/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
+++ b/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
@@ -67,7 +67,7 @@ void ompl::geometric::STRRTstar::setup()
     if (!tGoal_)
         tGoal_.reset(new ompl::NearestNeighborsLinear<Motion *>());
     tStart_->setDistanceFunction([this](const Motion *a, const Motion *b) { return distanceFunction(a, b); });
-    tGoal_->setDistanceFunction([this](const Motion *a, const Motion *b) { return distanceFunction(a, b); });
+    tGoal_->setDistanceFunction([this](const Motion *a, const Motion *b) { return distanceFunction(b, a); });
 
     if (si_->getStateSpace()->as<ompl::base::SpaceTimeStateSpace>()->getTimeComponent()->isBounded())
     {
--- a/src/ompl/base/spaces/src/SpaceTimeStateSpace.cpp
+++ b/src/ompl/base/spaces/src/SpaceTimeStateSpace.cpp
@@ -53,7 +53,7 @@ double ompl::base::SpaceTimeStateSpace::distance(const ompl::base::State *state1
     double deltaSpace = distanceSpace(state1, state2);
     double deltaTime = distanceTime(state1, state2);
 
-    if (deltaSpace / vMax_ > deltaTime + eps_) return std::numeric_limits<double>::infinity();
+    if (std::isinf(deltaTime) || (deltaSpace / vMax_ > deltaTime + eps_)) return std::numeric_limits<double>::infinity();
 
     return weights_[0] * deltaSpace + weights_[1] * deltaTime;
 
@@ -77,6 +77,8 @@ double ompl::base::SpaceTimeStateSpace::distanceSpace(const ompl::base::State *s
 double ompl::base::SpaceTimeStateSpace::distanceTime(const ompl::base::State *state1,
                                                      const ompl::base::State *state2) const
 {
+    if(getStateTime(state1)>getStateTime(state2)) return std::numeric_limits<double>::infinity();
+    
     const auto *cstate1 = static_cast<const CompoundState *>(state1);
     const auto *cstate2 = static_cast<const CompoundState *>(state2);
 


src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
```diff
@@ -67,7 +67,7 @@ void ompl::geometric::STRRTstar::setup()
     if (!tGoal_)
         tGoal_.reset(new ompl::NearestNeighborsLinear<Motion *>());
     tStart_->setDistanceFunction([this](const Motion *a, const Motion *b) { return distanceFunction(a, b); });
-    tGoal_->setDistanceFunction([this](const Motion *a, const Motion *b) { return distanceFunction(a, b); });
+    tGoal_->setDistanceFunction([this](const Motion *a, const Motion *b) { return distanceFunction(b, a); });
```
kaizer-nurik, line 70: I made a mistake, in NearestNeighborsLinear argument b is for the state, for which we are finding neighbours, 

https://github.com/ompl/ompl/blob/af4d6d625e15a4ed6d85255c718ce26bcd79e4cc/src/ompl/datastructures/NearestNeighborsLinear.h#L105

I fixed it locally, but strrt* does not work, with or without this fix. This means there is another problem.

aorthey: Hi @kaizer-nurik. Did you check if you can run the demo scripts of ST-RRT* in the demo folder? It is important to define a `MotionValidator` as described in the `SpaceTimePlanning.cpp` demo file, so that no negative time edges are added. If you still have issues, please don't hesitate to contact the author @frangrothe.

kaizer-nurik:
Hi @aorthey , I found the issue. I forgot to modify distance in rewireGoalTree and growTreeSingle. Now, my application and demo script SpaceTimePlanning.cpp works fine. I didn't notice significant improvements in single tests, but theoretically, this fixes can improve performance in hard tests by speeding up nearestR (because there won't be states, that are close enough, but t1>t2) or by helping nearestK (states with t1>t2 won't take up space of valid states in array size of K). 

Here is explanation of my changes:
1) SpaceTimeStateSpace::distanceTime(state1,state2) now returns inf if state1.time>state2.time; SpaceTimeStateSpace::distance(state1,state2) returns  inf if state1.time>state2.time. This changes make distance function defined as in paper;

2) I noticed, that in NearestNeighborsLinear.h in nearest function query state is passed as second argument

https://github.com/ompl/ompl/blob/af4d6d625e15a4ed6d85255c718ce26bcd79e4cc/src/ompl/datastructures/NearestNeighborsLinear.h#L105

also, query state is passed as second argument in nearestR and nearestK functions.

in function ompl::geometric::STRRTstar::growTree, if start tree is growing, that means, that query state in tree->nearest() is in future, while we want to find a state in start tree in the past (so that t1 < t2). Thats why, query state must be passed as second argument to distance function, and thats why line  

https://github.com/ompl/ompl/blob/a1f6347aac29ad0b9228ff129a967e55e82de876/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp#L69

is correct

Similarly, in function ompl::geometric::STRRTstar::growTree, if goal tree is growing, query state in tree->nearest() must be passed in distance function as first argument, because is is in the "past", and goal tree states are in the "future" (so t1<t2). Thats why line 

https://github.com/ompl/ompl/blob/a1f6347aac29ad0b9228ff129a967e55e82de876/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp#L70

is correct;

src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
```diff
@@ -518,7 +518,14 @@ ompl::geometric::STRRTstar::GrowState ompl::geometric::STRRTstar::growTreeSingle
...
     ompl::base::State *dstate = rmotion->state;
-    double d = si_->distance(nmotion->state, rmotion->state);
+
+    double d = 0;
+    if(tgi.start){
+        d = si_->distance(nmotion->state, rmotion->state);
+    }
+    else{
+        d = si_->distance( rmotion->state,nmotion->state);
+    }
```
kaizer-nurik, line 528: 3) in function ompl::geometric::STRRTstar::growTreeSingle, when we are growing from start tree, nmotion->state is "past", and rmotion->state is "future", so to satisfy t1<t2, we must pass this arguments as si_->distance(nmotion->state, rmotion->state);

When we are growing from goal tree, nmotion->state is "future", and rmotion->state is "past", so to satisfy t1<t2, we must pass this arguments as si_->distance( rmotion->state,nmotion->state);

src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
```diff
@@ -993,7 +1000,10 @@ bool ompl::geometric::STRRTstar::rewireGoalTree(Motion *addedMotion)
 {
     bool solved = false;
     std::vector<Motion *> nbh;
+    tGoal_->setDistanceFunction([this](const Motion *a, const Motion *b) { return distanceFunction(a, b); });
     getNeighbors(tGoal_, addedMotion, nbh);
+    tGoal_->setDistanceFunction([this](const Motion *a, const Motion *b) { return distanceFunction(b, a); });
```
kaizer-nurik, line 1005: 4) In ompl::geometric::STRRTstar::rewireGoalTree, when we are finding candidates for rewiring, we expect them to be in the "past", so our query state "addedMotion" is in the future. So, ti satisfy t1<t2, we must change distance function, and pass query state as second argument in distance function. (Inside nearestK and nearestR the query state is passed as second argument "b" )

PR #1181, opened 2024-09-09
fix signedness compare warning (for -Werror build)

v4hn:
Fixes these clang warnings:

```
ompl/geometric/planners/xxl/src/XXL.cpp:356:26: error: comparison of integers of different signs: 'int' and 'size_type' (aka 'unsigned long') [-Werror,-Wsign-compare]

ompl/demos/multilevel/MultiLevelPlanningKinematicChain.cpp:68:22: error: comparison of integers of different signs: 'value_type' (aka 'int') and 'const unsigned int' [-Werror,-Wsign-compare]
```


zkingston (approved):
LGTM, thanks!

Issue #1174, opened 2024-07-19
Compile ompl in cmake external project through ExternalProject or FetchContent


icosac:
Hi, 

I was trying to see if it was possible to compile and link ompl to an external cmake project I am working on. My idea is to use either FetchContent or ExternalProject from cmake to download and compile ompl and link it against my code. 

The problem is that if I add:

```cmake 
FetchContent_Declare(
  ompl
  GIT_REPOSITORY https://github.com/ompl/ompl.git
  GIT_TAG 1.6.0
)
FetchContent_MakeAvailable(ompl)
```

Then I cannot find the headers when compiling my code. So I added

```cmake
target_include_directories(points PRIVATE ${CMAKE_BINARY_DIR}/_deps/ompl-src/src)
```

Which fixes the missing headers for my files, but it raises the following errors when compiling ompl

```
In file included from /home/user/Projects/PD-line/build/_deps/ompl-src/src/ompl/base/StateSampler.h:41,
                 from /home/user/Projects/PD-line/build/_deps/ompl-src/src/ompl/base/StateSpace.h:42,
                 from /home/user/Projects/PD-line/build/_deps/ompl-src/src/ompl/base/spaces/SE2StateSpace.h:40,
                 from /home/user/Projects/PD-line/build/_deps/ompl-src/src/ompl/base/spaces/DubinsStateSpace.h:40,
                 from /home/user/Projects/PD-line/src/main_read_points.cpp:37:
/home/user/Projects/PD-line/build/_deps/ompl-src/src/ompl/util/RandomNumbers.h:46:10: fatal error: ompl/config.h: No such file or directory
   46 | #include "ompl/config.h"
      |          ^~~~~~~~~~~~~~~
compilation terminated.
gmake[2]: *** [CMakeFiles/points.dir/build.make:76: CMakeFiles/points.dir/src/main_read_points.cpp.o] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:357: CMakeFiles/points.dir/all] Error 2
gmake[1]: *** Waiting for unfinished jobs....
```

I'm not that good with ExternalProject as I'm not sure how to correctly link ompl headers and compiled files to my project. 

Has anyone had any success in linking ompl with FetchContent or ExternalProject? 

I successfully managed to use git modules and `add_subdirectory`, but I would much prefer to have something integrated directly with cmake. Do you reckon it would be possible? 

Thanks!

Ryanf55:
[...]

PR #1182, opened 2024-09-09
Use target include directories for FetchContent support

Ryanf55:
# Purpose

* Carry includes through target properties
* Attempt to preserve backwards compatability with OMPL_INCLUDE_DIRS
* Document how to use the library through FetchContent (platform-portable)
* Ensure the build interface can include ompl/config.h

This should also make it work in the same way with `add_subdirectory` for those who use git submodules and similar.

# TIcket

* Solves #1174

# Risks

I'm haven't tested if I handled `OMPL_INCLUDE_DIRS`. Can we set up a test for this somehow? Make sure `config.h` is found!  Of note, the docs say this: `      To support the previous CMake module behavior, the following variables are defined. These are for backwards compatability, but not recommended anymore.` I don't think this is actually true when we install the package. Let's discuss whether we should fix it. 

```bash
$ grep OMPL_INCLUDE_DIRS -rn install
install/share/ompl/cmake/Findcastxml.cmake:37:        "${OMPL_INCLUDE_DIRS}"
$ grep OMPL_LIBRARIES -rn install
$ grep OMPL_VERSION -rn install
```

# Testing

Use this code example in another folder. Place it outside of your ompl folder. 

```cmake
cmake_minimum_required(VERSION 3.22)
project(ompl_consumer)

include(FetchContent)
FetchContent_Declare(
  ompl
)

FetchContent_MakeAvailable(ompl)
add_executable(main main.cpp)
target_link_libraries(main PRIVATE ompl::ompl)
```

Build it like so. Make sure to check out this branch of ompl, and set the path to it correctly so FetchContent uses the local version.
`cmake -B build -DFETCHCONTENT_SOURCE_DIR_OMPL=/your/path/to/omplapp/ompl`

**main.cpp**
```
#include <ompl/base/ScopedState.h>

int main() {
    return 0;
}
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,11 +25,8 @@ include(OMPLUtils)
 
 set(OMPL_CMAKE_UTIL_DIR "${CMAKE_CURRENT_SOURCE_DIR}/CMakeModules"
     CACHE FILEPATH "Path to directory with auxiliary CMake scripts for OMPL")
-set(OMPL_INCLUDE_DIRS "${CMAKE_CURRENT_SOURCE_DIR}/src;${CMAKE_CURRENT_BINARY_DIR}/src")
 set(OMPL_DEMO_INSTALL_DIR "${CMAKE_INSTALL_DATAROOTDIR}/ompl/demos"
     CACHE STRING "Relative path to directory where demos will be installed")
-
-include_directories("${OMPL_INCLUDE_DIRS}")
 
 if(MSVC)
     add_definitions(-DBOOST_ALL_NO_LIB)
--- a/doc/markdown/installation.md
+++ b/doc/markdown/installation.md
@@ -10,6 +10,7 @@
     <li class="nav-item"><a class="nav-link" id="linux-tab" data-toggle="pill" href="#linux" role="tab" aria-controls="linux" aria-selected="false">Linux (generic)</a></li>
     <li class="nav-item"><a class="nav-link" id="osx-tab" data-toggle="pill" href="#osx" role="tab" aria-controls="macos" aria-selected="false">macOS</a></li>
     <li class="nav-item"><a class="nav-link" id="windows-tab" data-toggle="pill" href="#windows" role="tab" aria-controls="windows" aria-selected="false">MS Windows</a></li>
+    <li class="nav-item"><a class="nav-link" id="cmake-fetch-tab" data-toggle="pill" href="#cmakefetch" role="tab" aria-controls="cmakefetch" aria-selected="false">CMake Fetchcontent</a></li>
   </ul>
 </div>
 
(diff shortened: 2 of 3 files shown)


zkingston (approved):
LGTM - was able to compile the example and also another project that uses OMPL as a submodule.

MarqRazz:
With this change I now get build errors when using Colcon. 
```
--- stderr: moveit_planners_ompl                                
CMake Error in ompl_interface/CMakeLists.txt:
  Imported target "ompl::ompl" includes non-existent path

    "/root/moveit2_ws/install/ompl/include/ompl"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

  * The path was deleted, renamed, or moved to another location.

  * An install or uninstall procedure did not complete successfully.

  * The installation package was faulty and references files it does not
  provide.

CMake Generate step failed.  Build files cannot be regenerated correctly.
gmake: *** [Makefile:380: cmake_check_build_system] Error
```

The actual path to the headers in this container is 
`/root/moveit2_ws/install/ompl/include/ompl-1.6/ompl`

Ryanf55:
> With this change I now get build errors when using Colcon.
> 
> ```
> --- stderr: moveit_planners_ompl                                
> CMake Error in ompl_interface/CMakeLists.txt:
>   Imported target "ompl::ompl" includes non-existent path
> 
>     "/root/moveit2_ws/install/ompl/include/ompl"
> 
>   in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:
> 
>   * The path was deleted, renamed, or moved to another location.
> 
>   * An install or uninstall procedure did not complete successfully.
> 
>   * The installation package was faulty and references files it does not
>   provide.
> 
> CMake Generate step failed.  Build files cannot be regenerated correctly.
> gmake: *** [Makefile:380: cmake_check_build_system] Error
> ```
> 
> The actual path to the installed headers in my container is `/root/moveit2_ws/install/ompl/include/ompl-1.6/ompl`

Thanks for the report and sorry for breaking your build. I used this in a ROS workspace with a fresh clone and it built fine like so.
`ros2 pkg create --build-type ament_cmake --dependencies ompl --node-name foo  bar`

I couldn't reproduce with a standard test package generated with `ros2 pkg create`; I realized this is because this issue ONLY surfaces if you link to the new `ompl::ompl` target; the current package generation on `humble` uses `ament_target_dependencies` on the project name instead. 


Digging into this, it seems OMPL is setting the header install path in two places, so I suspect I should drop the one from `target_link_libraries`.
https://stackoverflow.com/questions/59743205/do-i-really-need-to-specify-install-interface-includes-dependencies



The super odd thing is that I already have a project for checking the target works here:
https://github.com/ompl/ompl/blob/f8813eccd7e96487231b5813e69daa65d130d65b/tests/cmake_export/CMakeLists.txt#L6
Then, it's used in the dockerfile here:
https://github.com/ompl/ompl/blob/f8813eccd7e96487231b5813e69daa65d130d65b/scripts/docker/ompl.Dockerfile#L46
`OMPL_VERSIONED_INSTALL` is on by default. 
You can build the docker image in ompl with `cmake --build build --target docker`.

At this point, it might be obvious. The dockerfile is not built in CI, so this is not tested in CI. Oops! 

I'll ensure with the patch that I have test to check this can't regress.

Apologies for the code thrash, hopefully it doesn't impact too many users.

Ryanf55:
Please check this resolves your issue: https://github.com/Ryanf55/ompl/commit/40dc2cce3e58313c7652164384f3659242885ec2

I am going to need some time to iterate on the CI - it's using an action I am not familiar with.

Ryanf55:
https://github.com/ompl/ompl/pull/1184 
^ This fixes the regression and adds CI for it.

PR #1184, opened 2024-09-11
Hotfix: Cmake installation with version folder

Ryanf55:
* The installation path was incorrectly changed and broke users who link to ompl::ompl after find_package. It was changed to support FetchContent.
* This was caused by the accidental change from https://github.com/ompl/ompl/pull/1182/files#diff-148715d6ea0c0ea0a346af3f6bd610d010d490eca35ac6a9b408748f7ca9e3f4L15  to https://github.com/ompl/ompl/pull/1182/files#diff-148715d6ea0c0ea0a346af3f6bd610d010d490eca35ac6a9b408748f7ca9e3f4R6
* Add CI to enforce the existing export test because the dockerfile is not build in CI

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -55,7 +55,12 @@ jobs:
         run-test: true
         ctest-options: --output-on-failure
         install-build: true
-
+    - name: Test CMake target linkage to ompl::ompl
+      if: runner.os != 'Windows'
+      run: |
+        cd tests/cmake_export
+        cmake -B build -DCMAKE_INSTALL_PREFIX=${{ runner.workspace }}/install
+        cmake --build build
     - name: Build & Test (Windows)
       if: runner.os == 'Windows'
       uses: ashutoshvarma/action-cmake-build@master
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -3,7 +3,9 @@ target_include_directories(ompl
   PUBLIC
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>"
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>"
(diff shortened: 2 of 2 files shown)


Ryanf55:
The Ubuntu 24 failure is unrelated:
https://github.com/ompl/ompl/actions/runs/10804144479/job/29969133090?pr=1184#step:6:1256
The MacOS failure is caused by incorrect exception specifiers, and is only caught by doing this kind of test
https://github.com/ompl/ompl/actions/runs/10804144479/job/29969133511?pr=1184#step:7:32
We might have to fix that in order to add CI.

zkingston (approved):
@Ryanf55 thank you so much for handling the CMake fixes.

MarqRazz (commented):
Just wanted to say thanks for patching this and adding tests! I ran this locally in a standard Colcon (Jazzy) workspace with MoveIt2 and it fixes the regression I was seeing.

MarqRazz:
I think the test is good (its the same process I'm following with my proposed changes [here](https://github.com/moveit/moveit2/pull/2994) to MoveIt2)

If OMPL would like to properly support Ament/Colcon I would suggest we look at how [BehaviorTree.CPP selectively uses it](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/CMakeLists.txt#L52-L60) when available.

Ryanf55:
> I think the test is good (its the same process I'm following with my proposed changes [here](https://github.com/moveit/moveit2/pull/2994) to MoveIt2)
> 
> If OMPL would like to properly support Ament/Colcon I would suggest we look at how [BehaviorTree.CPP selectively uses it](https://github.com/BehaviorTree/BehaviorTree.CPP/blob/master/CMakeLists.txt#L52-L60) when available.

I was just curious if it would buy us anything than we already have. Yea BTCPP's implementation is a good approach, I use that on a lot of repos that aren't fully bought into the ROS build system.

PR #1188, opened 2024-09-23
Fix various Python wheel generation issues

kylc:
- Fix numpy at <2.0 because Boost-Python does not support numpy>=2.0 as of Boost-1.86. See this issue; the fix will likely be included in 1.87: https://github.com/boostorg/python/issues/431
- Use the new Github Mac M1 runner when building the arm64 wheel. This makes that target a native build instead of a cross-compile.
  - This requires some additional logic to find the system clang++ executable. I don't have a Mac so I can't test a better solution, but it could probably be cleaned up.
- Work around a `error: no member named 'piecewise_construct' in namespace 'std';` error on MacOS by bumping Boost dependency version. See: https://github.com/NREL/OpenStudio/issues/5229
- Fix a CastXML build error complaining about missing the clang resource directory.
- Silence a CMake warning when using CMake>=3.30 caused by the removal of `FindBoost` (https://cmake.org/cmake/help/latest/policy/CMP0167.html)

@zkingston I also pulled in your changes to fix an error in the Python binding generation due to the recent changes in include paths. These affected the CI wheel generation too, but I can wait and rebase if you want to merge those separately.

Follow here to see a successful CI build: https://github.com/kylc/ompl/actions/runs/10998657273

### DIFF
--- a/.github/workflows/wheels.yaml
+++ b/.github/workflows/wheels.yaml
@@ -14,7 +14,7 @@ jobs:
         os: [ubuntu-22.04, macos-13]
         arch: [x86_64]
         include:
-          - os: macos-13
+          - os: macos-14
             arch: arm64
     steps:
       - name: Check out repository code
--- a/py-bindings/setup.py
+++ b/py-bindings/setup.py
@@ -49,6 +49,7 @@ class CMakeBuild(build_ext):
         # EXAMPLE_VERSION_INFO shows you how to pass a value into the C++ code
         # from Python.
         cmake_args = [
+            "-DCMAKE_CXX_COMPILER=/usr/bin/clang++",  # Force Clang for castxml
             f"-DCMAKE_LIBRARY_OUTPUT_DIRECTORY={extdir}{os.sep}",
             f"-DPYTHON_EXEC={sys.executable}",
             f"-DPYTHON_INCLUDE_DIRS={get_paths()['include']}",
--- a/py-bindings/setup.py
+++ b/py-bindings/setup.py
@@ -108,8 +108,8 @@ class CMakeBuild(build_ext):
         if sys.platform.startswith("darwin"):
             # Search these paths to try to find the clang++ binary.
             potential_clang_paths = [
-                "/usr/local/opt/llvm@16/bin/clang++",
-                "/opt/homebrew/opt/llvm@16/bin/clang++",
+                "/usr/local/opt/llvm@18/bin/clang++",
+                "/opt/homebrew/opt/llvm@18/bin/clang++",
             ]
             clang_bin_path = next(
                 filter(lambda p: Path(p).exists(), potential_clang_paths), None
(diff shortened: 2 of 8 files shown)


.github/workflows/wheels.yaml
```diff
@@ -2,8 +2,6 @@ name: Build Wheels
 on:
   workflow_dispatch:
   push:
-    branches:
-      - main
```
kylc, line 6: TODO: Just for testing, remove before merging

zkingston:
Thanks @kylc! Happy to have those changes I made put here. Something we might also want to consider is dropping the pinned version of pygccxml, and forcing only clang to be used as the compiler for the bindings (since castxml is not playing nice with modern gcc). See: 
  - https://github.com/ompl/ompl/commit/8caf5711f5f6e082521a2b9087f2a358194ac89f#diff-dd108539255863968fda1b52f8d1908bde57908f32f93d1beb58652aa06b05bdR6
  - https://github.com/ompl/ompl/commit/8caf5711f5f6e082521a2b9087f2a358194ac89f#diff-60f7d754a0c159e4789a1ac359dc6552c6187ec903af6e6c3af43232be7449a8R55

py-bindings/setup.py
```diff
@@ -49,6 +49,7 @@ def build_extension(self, ext: CMakeExtension) -> None:
         # EXAMPLE_VERSION_INFO shows you how to pass a value into the C++ code
         # from Python.
         cmake_args = [
+            "-DCMAKE_CXX_COMPILER=/usr/bin/clang++",  # Force Clang for castxml
```
zkingston, line 52: Probably want to have this be a variable with the found clang executable, not explicitly hardcoding this path.

zkingston: Maybe move the below clang search to above?

zkingston (approved):
LGTM. Not sure what is going on with the failed Mac build - cannot replicate that on my local machine.

py-bindings/setup.py
kylc: Now that we build the macos/arm64 target natively instead of cross-compiling, it seem like we can use the system clang installation instead of the one from homebrew. The clang binary search logic is now gone in https://github.com/ompl/ompl/pull/1188/commits/c817f3668f38cd36a7184a63ee1520453b990740.

I've replaced this bit with a `shutil.where("clang++")` in https://github.com/ompl/ompl/pull/1188/commits/8740aa897b2d0943c9a3e4458d69077aa46e5f2d.

kylc:
Pulled in your changes mostly as-is. I also bumped the CastXML version in case that separately resolves the GCC issues; I see some GCC-related commits recently in the CastXML repo.

Will remove the WIP commit after the wheel job succeeds on my fork and then we can get this merged.

If you can take a look at https://github.com/ompl/pyplusplus/pull/2, it fixes some errors with the Python bindings on MacOS that have been reported as issues here.

zkingston:
Mark got there before me, but looks good to go!

kylc:
Thanks, ready to merge!

PR #1195, opened 2024-10-31
Fix AIT*/EIT* not respecting termination condition when there are no valid samples

yurirocha15:
Currently, both the AIT* and EIT* are initializing their Informed State Samplers' maxNumberCalls with `std::numeric_limits<unsigned int>::max()`. The maxNumberCalls will [determine the maximum number of tries the sampler will do to find a valid sample](https://github.com/ompl/ompl/blob/main/src/ompl/base/samplers/informed/src/RejectionInfSampler.cpp#L77-L96). So currently, if for some reason the sampler cannot find a valid sample (maybe because a PHS is way smaller than the full sampling space), it will try at most `std::numeric_limits<unsigned int>::max()` (~4.3 billion) times before returning to the planner. (In my experiments, I had a moveit plan taking 1000s when the maximum planning time was set to 50s)

As this value was already hard coded, I simply changed the value to 100, mainly because this is the default value used by [RRT*](https://github.com/ompl/ompl/blob/main/src/ompl/geometric/planners/rrt/RRTstar.h#L491). (Should this be exposed with a getter and a setter like RRT*?)

It would be better if the samplers also respected the termination conditions, but this change would require changing the ABI in a lot of places.

### DIFF
--- a/src/ompl/geometric/planners/informedtrees/aitstar/src/ImplicitGraph.cpp
+++ b/src/ompl/geometric/planners/informedtrees/aitstar/src/ImplicitGraph.cpp
@@ -267,8 +267,7 @@ namespace ompl
                     // Allocate a state sampler if we have at least one start and one goal.
                     if (!startVertices_.empty() && !goalVertices_.empty())
                     {
-                        sampler_ = objective_->allocInformedStateSampler(problemDefinition_,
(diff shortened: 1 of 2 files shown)


marlinstrub (approved):
Apologies for the super late review and thanks for the improvement.

PR #1199, opened 2024-11-27
Save test planner data to different files

JafarAbdi:
Fix: https://github.com/conda-forge/ompl-feedstock/issues/43#issuecomment-2502656357

I ran into this issue as well while debugging https://github.com/conda-forge/ompl-feedstock/issues/43, these two tests are storing/loading from the same file when ran in parallel.


mamoll (approved):
Finally! This bug has been present for so many years.

Thanks, Jafar!

PR #1185, opened 2024-09-12
Restore legacy CMAKE variables 

Ryanf55:
# Purpose

The legacy variables that historically used with ompl have been broken on `main` for a few months as a result of #1067. 
These variables were not tested for validity in CI, so there was nothing to enforce regressions.
These should not be broken because OMPL did not bump its major version.
This has caused projects such as MoveIt that are consuming the `main` branch of `ompl` to be forced to linking `ompl::ompl`. We want to recommend the new targets, but do our best not to break older workflows.

This PR does a best-effort attempt to bring them back. 

# Details

The legacy variables have some known issues which is why we want to recommend target-based linkage to OMPL consumers. 
One concern is the installation is not portable to systems with dependencies installed in other locations, as seen [here](https://github.com/ompl/ompl/pull/1067#discussion_r1753001282). The PR does not attempt to solve those pre-existing problems. 

# Behavior

Here's what the old OMPL would set up for the "legacy" variables.
```
--  OMPL_FOUND: TRUE
--  OMPL_INCLUDE_DIRS: /home/ryan/Dev/moveit_ws/install/ompl/include/ompl-1.6;/usr/include;/usr/include/eigen3
--  OMPL_LIBRARY_DIRS: /home/ryan/Dev/moveit_ws/install/ompl/lib;/usr/lib/x86_64-linux-gnu;/usr/lib
--  OMPL_LIBRARIES: /home/ryan/Dev/moveit_ws/install/ompl/lib/libompl.so;optimized;/usr/lib/x86_64-linux-gnu/libboost_serialization.so.1.74.0;debug;/usr/lib/x86_64-linux-gnu/libboost_serialization.so;optimized;/usr/lib/x86_64-linux-gnu/libboost_filesystem.so.1.74.0;debug;/usr/lib/x86_64-linux-gnu/libboost_filesystem.so;optimized;/usr/lib/x86_64-linux-gnu/libboost_system.so.1.74.0;debug;/usr/lib/x86_64-linux-gnu/libboost_system.so
--  ompl_FOUND: TRUE
```
Here's what you get on ompl `main`:
```
-- OMPL_FOUND: 
--  OMPL_INCLUDE_DIRS: /home/ryan/Dev/moveit_ws/install/ompl/include/ompl-1.6;/usr/include
--  OMPL_LIBRARY_DIRS: /home/ryan/Dev/moveit_ws/install/ompl/lib
[...]

PR #1206, opened 2024-12-20
Create a new release 1.7.0

JafarAbdi:
@MarqRazz and I are working on cutting a new release to Jazzy/Rolling, this is our first attempt to take over the release process for ROS 2.

@mamoll We have a few questions:

- How do you organize the release notes? I just copied commits' since the last release for now
- How do you want to handle tagging the release? Pushing the git tag
- Should we bump the version number after we tag? To make it so building from source has a different version from the binaries

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.12)
-project(ompl VERSION 1.6.0 LANGUAGES CXX)
-set(OMPL_ABI_VERSION 17)
+project(ompl VERSION 1.7.0 LANGUAGES CXX)
+set(OMPL_ABI_VERSION 18)
 
 # Use the FindBoost provided by CMake, rather than the one provided by Boost
 # (for CMake >=3.30).
--- a/doc/markdown/releaseNotes.md
+++ b/doc/markdown/releaseNotes.md
@@ -1,5 +1,53 @@
 # Release Notes {#releaseNotes}
 
+## OMPL 1.7.0 (December 20, 2024)
+
+- Fix cmake installation with version folder 
+- Use target properties for include properties 
+- Fix off-by-one error and deprecated call in documentation 
+- Fix signedness compare (for -Werror build) 
+- Ensure that `webbrowser` pkg will not use terminal browser
+- Repair library install destination
+- Correctly overload virtual function
+- Fix Python wheels missing constrained planning components
+- fix(script): add missing space in if statement
+- More relevant return code
+- Fix default value for goal bias fraction in EST planner
+- Do not retry on invalid start/goal/unrecognized goal type
+- Add missing build command in dockerfile
+- Fix termination before planner termination condition is met in multilevel planning.
+- Added Rope Shortcut
+- remove vestigial bits of code now that ODE extension has been removed
+- remove unused parameter in getDubinsClass
+- Add support for building Python wheels for Linux & MacOS
+- Fix for ompl::geometric Python bindings
+- Fix python binding header include order
+- Python bindings compiling on M1/2 Mac
+- fix #1117: out-of-bounds query in 2D environmen
+- Fix yaml-cpp target to be compatible with CMake <3.18
+- Fix Bug in Loading Planner Data
+- Allow allocating custom state sampler in rejection sampler
+- removed unnecessary config import in RandomNumbers
+- Use full install paths for pkg-config
+- Add a CMake export step for config packages.
+- wrap bias function in constrained planning demo
+- fix issue: EIGEN3_INCLUDE_DIR is not defined when Eigen3 is installed from source
+- Use of Dubins Set Classification for faster distance computations in DubinsStateSpace
+- Only prune goal in BIT* if it is not part of the current best solution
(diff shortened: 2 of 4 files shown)


doc/markdown/releaseNotes.md
```diff
@@ -1,5 +1,53 @@
 # Release Notes {#releaseNotes}
 
+## OMPL 1.7.0 (December 20, 2024)
+
+- Fix cmake installation with version folder
```
JafarAbdi, line 5: I removed the CI related ones

mamoll:
@JafarAbdi / @MarqRazz, thanks for getting the ball rolling for a 1.7.0 release. An OMPL release is separate from a release for ROS2. There are several downstream maintainers (like yourself) who release OMPL for: ROS2, the "official" Debian/Ubuntu package, vcpkg, Fedora, Homebrew, Macports, etc. It seems like enough has changed that perhaps it's a good time to cut a new release. @zkingston @wbthomason @kylc : are there any outstanding issues that need to be fixed first? I am thinking of Python binding issues, but there could be others.

To answer you question re. release process, this is the checklist we put together many years ago that roughly describes this:
https://github.com/ompl/omplapp/blob/main/doc/releaseChecklist.txt

MarqRazz:
Would it be possible for us to make some changes to the release process? I think it would be clearer if we were to bump the version number after each release. That way from source or binary it is obvious which version you are working with and know if the feature/patch you need has been released.

mamoll:
> Would it be possible for us to make some changes to the release process? I think it would be clearer if we were to bump the version number after each release. That way from source or binary it is obvious which version you are working with and know if the feature/patch you need has been released.

I think we already bump the version number for every release. Is this not clear in the release check list? Or have we not followed the check list in the past? I agree that stealth updates should be avoided.

MarqRazz:
Sorry I'm suggesting we bump the source code after each release so that it does not match the last tag. I found it confusing that `1.6.0` was released in January 2023 and source still had the same version number but had breaking changes merged into it. If we were to move bumping the number to after we tag then it would be obvious which version you are working with.

simonschmeisser:
There's at least one open MoveIt PR that could have been merged if there were a new version number already, so +1 for early bumping from me

MarqRazz:
Is there anything I can do to help move this along?

kylc:
As far as I know there isn't any blocker on the Python wheel generation. The wheel CI jobs on the latest main branch commit were successful.

One nice-to-have before a release would be to upgrade to boost-python 1.87 to enable numpy 2.0 support (see https://github.com/boostorg/python/issues/431). Currently, if you use the CI wheels, you have to install `numpy<2.0`.

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.12)
-project(ompl VERSION 1.6.0 LANGUAGES CXX)
-set(OMPL_ABI_VERSION 17)
+project(ompl VERSION 1.7.0 LANGUAGES CXX)
```
kylc, line 2: Also bump the version in [py-bindings/setup.py](https://github.com/ompl/ompl/blob/d3f2cc76de0dfc33d1e6322adb2376b90e1c1241/py-bindings/setup.py#L153)

JafarAbdi:
> I'd also like [my own recent PR](https://github.com/ompl/ompl/pull/1210) merged in. That shouldn't affect much else in OMPL (or MoveIt).

Anything I can help with? I fixed an issue with Python wheel generation and added python-binding for the new state spaces [here](https://github.com/ompl/ompl/pull/1215)

MarqRazz:
Bumping this to see how it's coming along. What do you think about making a list of blocking PR's here so that we can start to track progress towards releasing?

lepalom:
@mamoll If you release a new version I would try to put it on Debian ASAP. I know that most of you use the ROS2 version, but the original Debian package doesn't depend on ROS and can be interesting for some people.

MarqRazz:
Any movement here? Do we have a list of what is blocking us from releasing?

mamoll:
Hi @MarqRazz, sorry it's taking so long. We are very close. I have a PR (PR #1242) with some cleanup in preparation of release. I am handing over the actual release process to @saahu27, a new maintainer and software engineer in Lydia Kavraki's lab. @saahu27, do you have an estimate of how much more time you need?

saahu27:
Hi @MarqRazz  @mamoll I'll need about 1-2 weeks to complete the release process. I'm currently getting up to speed with all the necessary procedures and working through some parallel tasks. I should have a more specific timeline by the end of next week.

Please let me know if you need any additional information from me in the meantime or if there are any specific priorities I should focus on for this release.

MarqRazz:
> Please let me know if you need any additional information from me in the meantime or if there are any specific priorities I should focus on for this release.

[This](https://github.com/ompl/ompl/pull/1185) is the only PR that I can think of. If @Ryanf55 is able to build with the legacy CMAKE variables then I think we should get it in and not worry about MoveIt (we don't need legacy any more). If we get that fixed down the road it should be an easy patch to extend legacy support.

mamoll:
release is out, so closing this issue. Please re-open if there are issues with it.

PR #1215, opened 2025-01-05
Add python binding to 3D Dubins models

JafarAbdi:


py-bindings/headers_base.txt
```diff
@@ -53,6 +53,9 @@ src/ompl/base/spaces/DiscreteStateSpace.h
 src/ompl/base/spaces/DubinsStateSpace.h
 src/ompl/base/spaces/ReedsSheppStateSpace.h
 src/ompl/base/spaces/WrapperStateSpace.h
+src/ompl/base/spaces/VanaStateSpace.h
```
JafarAbdi, line 56: Should I add entries to py-bindings/ompl_py_base.h?

JafarAbdi: Fixed!

mamoll (approved):

mamoll:
Seems to work! Thanks a lot. The Python bindings are always a pain. I am writing a simple python version of the DubinsAirplane.cpp demo and get a `RuntimeError: unidentifiable C++ exception` sometimes. I need to track that down as well as fix an unrelated assertion failure in OwenStateSpace.

PR #1216, opened 2025-01-05
Update boost version to support numpy>=2.0

JafarAbdi:


JafarAbdi:
@kylc Is this what I need to do? I tested it locally with numpy>=2.0 and numpy 1.26.4 without any issues

kylc (approved):

mamoll (approved):

Issue #1061, opened 2023-06-14
global-buffer-overflow on getControl()


Rrooach:
Hello, I found a global-buffer-overflow on ompl/src/ompl/control/PlannerData.h:getControl()

### Steps to reproduce 
- Download the latest source code, compiled with Clang and ASAN, by exporting CFLAGS="-g -fsanitize=address" LDFLAGS="-fsanitize=address" before compiling.
- run the example code: test_planner_data_control

### Expected behavior
The program finished without error.

### Actual behavior 
When running the first example, meet the following output:
``` bash
Running 5 test cases...
=================================================================
==672751==ERROR: AddressSanitizer: global-buffer-overflow on address 0x7fa6adc771a8 at pc 0x7fa6acb37939 bp 0x7ffe17a714d0 sp 0x7ffe17a714c0
READ of size 8 at 0x7fa6adc771a8 thread T0
    #0 0x7fa6acb37938 in ompl::control::PlannerDataEdgeControl::getControl() const ompl/src/ompl/control/PlannerData.h:83
    #1 0x7fa6acb37938 in ompl::control::PlannerData::removeEdge(ompl::base::PlannerDataVertex const&, ompl::base::PlannerDataVertex const&) ompl/src/ompl/control/src/PlannerData.cpp:96
    #2 0x55c23048833d in AddRemoveVerticesAndEdges::test_method() ompl/tests/control/planner_data.cpp:464
    #3 0x55c23048de38 in AddRemoveVerticesAndEdges_invoker ompl/tests/control/planner_data.cpp:414
    #4 0x7fa6add08531  (/lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0+0x27531)
    #5 0x7fa6add0e5b0 in boost::execution_monitor::catch_signals(boost::function<int ()> const&) (/lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0+0x2d5b0)
    #6 0x7fa6add0eaa0 in boost::execution_monitor::execute(boost::function<int ()> const&) (/lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0+0x2daa0)
    #7 0x7fa6add0eb5b in boost::execution_monitor::vexecute(boost::function<void ()> const&) (/lib/x86_64-linux-gnu/libboost_unit_test_framework.so.1.74.0+0x2db5b)
[...]

PR #1235, opened 2025-02-17
Fix buffer overflow when edge is not a PlannerDataEdgeControl

clunietp:
Fixes ompl/ompl#1061


clunietp:
@mamoll I have not contributed to this repo before.  I'd like to contribute more bug/security-related fixes but I'm unsure if I'm following the correct process.  What is the typical process here?  Thanks

Issue #1183, opened 2024-09-11
ompl.Dockerfile unmet dependencies


Ryanf55:
# Purpose

I'm reporting an issue I found trying to build the OMPL dockerfile. I don't need it working, but thought it could be tracked here in case someone else runs into it.

# Steps to reproduce
Check out `main` on hash f8813eccd7e96487231b5813e69daa65d130d65b.

Command
```
cmake -B build
cmake --build build --target docker
```

Logs
```
106.9 Setting up libhdf5-openmpi-dev (1.10.7+repack-4ubuntu2) ...
106.9 update-alternatives: using /usr/bin/h5pcc.openmpi to provide /usr/bin/h5pcc (h5pcc) in auto mode
106.9 update-alternatives: warning: skip creation of /usr/share/man/man1/h5pcc.1.gz because associated file /usr/share/man/man1/h5pcc.openmpi.1.gz (of link group h5pcc) doesn't exist
106.9 update-alternatives: warning: skip creation of /usr/share/man/man1/h5pfc.1.gz because associated file /usr/share/man/man1/h5pfc.openmpi.1.gz (of link group h5pcc) doesn't exist
106.9 update-alternatives: using /usr/lib/x86_64-linux-gnu/pkgconfig/hdf5-openmpi.pc to provide /usr/lib/x86_64-linux-gnu/pkgconfig/hdf5.pc (hdf5.pc) in auto mode
106.9 update-alternatives: updating alternative /usr/bin/mpicc.openmpi because link group mpi has changed slave links
106.9 update-alternatives: warning: skip creation of /usr/share/man/man1/mpiCC.1.gz because associated file /usr/share/man/man1/mpiCC.openmpi.1.gz (of link group mpi) doesn't exist
106.9 update-alternatives: warning: skip creation of /usr/share/man/man1/mpic++.1.gz because associated file /usr/share/man/man1/mpic++.openmpi.1.gz (of link group mpi) doesn't exist
106.9 update-alternatives: warning: skip creation of /usr/share/man/man1/mpicc.1.gz because associated file /usr/share/man/man1/mpicc.openmpi.1.gz (of link group mpi) doesn't exist
106.9 update-alternatives: warning: skip creation of /usr/share/man/man1/mpicxx.1.gz because associated file /usr/share/man/man1/mpicxx.openmpi.1.gz (of link group mpi) doesn't exist
[...]

PR #1242, opened 2025-03-08
pre1.7release cleanup

mamoll:
minor clean-up
update Dockerfiles for ompl and omplapp to ubuntu 24.04 (noble)

PR #1243, opened 2025-03-13
dockerfile updated to ubuntu 24.04

saahu27:
Fixes Issue #1183


mamoll:
Does this successfully build the python bindings? I already have a Dockerfile update pending in PR #1242. See https://github.com/ompl/ompl/pull/1242/files#diff-41d3d8136a81f2b33ed5b0a32a943796dccf5e6bd83da816237194ecfebe6b04. I had to use clang++ to get the python bindings to compile.

PR #1244, opened 2025-03-15
Add trochoid state space for dubins vehicles under wind

Jaeyoung-Lim:
**Problem Description**
This PR adds an implementation of a trochoid state space, which computes the curvature constrained shortest path in the influence of flow (wind).

### DIFF
--- a/src/ompl/base/spaces/TrochoidStateSpace.h
+++ b/src/ompl/base/spaces/TrochoidStateSpace.h
@@ -1,0 +1,198 @@
+/*********************************************************************
+* Software License Agreement (BSD License)
+*
+*  Copyright (c) 2010, Rice University
+*  All rights reserved.
+*
+*  Redistribution and use in source and binary forms, with or without
+*  modification, are permitted provided that the following conditions
+*  are met:
+*
+*   * Redistributions of source code must retain the above copyright
+*     notice, this list of conditions and the following disclaimer.
+*   * Redistributions in binary form must reproduce the above
+*     copyright notice, this list of conditions and the following
+*     disclaimer in the documentation and/or other materials provided
+*     with the distribution.
+*   * Neither the name of the Rice University nor the names of its
+*     contributors may be used to endorse or promote products derived
+*     from this software without specific prior written permission.
+*
+*  THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+*  "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+*  LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS
+*  FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE
+*  COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT,
+*  INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING,
+*  BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+*  LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+*  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+*  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+*  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+*  POSSIBILITY OF SUCH DAMAGE.
+*********************************************************************/
+
+/* Author: Mark Moll */
+
+#ifndef OMPL_BASE_SPACES_TROCHOID_STATE_SPACE_
+#define OMPL_BASE_SPACES_TROCHOID_STATE_SPACE_
+
+#include "ompl/base/spaces/SE2StateSpace.h"
+#include "ompl/base/MotionValidator.h"
+#include <boost/math/constants/constants.hpp>
+
+namespace ompl
+{
+    namespace base
+    {
+        /** \brief An SE(2) state space where distance is measured by the
+            length of Trochoid shortest paths curves.
+
+            The notation and solutions in the code are taken from:<br>
+            Techy, Laszlo, and Craig A. Woolsey. "Minimum-time path planning for unmanned aerial vehicles
+             in steady uniform winds." Journal of guidance, control, and dynamics 32.6 (2009): 1736-1746.
+
+            The implementation is from adapted from https://github.com/robotics-uncc/ConvectedDubins
+            written by Artur Wolek.
+            */
+        class TrochoidStateSpace : public SE2StateSpace
+        {
+        public:
+            /** \brief The Dubins path segment type */
+            enum TrochoidPathSegmentType
+            {
+                TROCHOID_LEFT = 0,
+                TROCHOID_STRAIGHT = 1,
+                TROCHOID_RIGHT = 2
+            };
+
+            /** \brief Dubins path types */
+            static const TrochoidPathSegmentType dubinsPathType[6][3];
+            /** \brief Complete description of a Dubins path */
+            class TrochoidPath
+            {
+            public:
+                TrochoidPath(const TrochoidPathSegmentType *type = dubinsPathType[0], double tA = 0.,
+                           double p = std::numeric_limits<double>::max(), double tB = 0., double t_2pi = 0.)
+                  : type_(type)
+                {
+                    length_[0] = tA;
+                    length_[1] = p;
+                    length_[2] = t_2pi - tB;
+                    assert(tA >= 0.);
+                    assert(p >= 0.);
+                    assert(t_2pi - tB >= 0.);
+                }
+                double length() const
+                {
+                    ///TODO: Fix for trochoid paths
+                    return length_[0] + length_[1] + length_[2];
+                }
+
+                friend std::ostream& operator<<(std::ostream& os, const TrochoidPath& path);
+                /** Path segment types */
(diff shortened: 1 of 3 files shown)


mamoll (changes requested):
Thanks for your contribution! This seems very useful. I have a couple high-level comments/questions before doing a detailed review:
* Would it be possible to avoid some duplication of code w.r.t. the DubinsStateSpace by using inheritance? It seems like you probably don't need a new class for the motion validator. Look at what I did for the 3D Dubins spaces as an example.
* You'll probably also want to make a change in SpaceInformation.cpp s.t. your motion validator is configured automatically as the default for TrochoidStateSpace
* At a minimum, a simple demo that shows how this space can be used should be part of this PR. (Ideally, also some unit tests, but a demo is sufficient.) You should consider expanding the demo in `ompl/demos/GeometricCarPlanning.cpp`. You probably also have some code that you used to convince yourself that things worked correctly. Is it possible to get this included in this PR?

Jaeyoung-Lim:
@mamoll Would you have any clue on why the code would be failing on macOS 13 with the owen statespace?
```
  unknown location:0: fatal error: in "Trochoid_Simple": signal: SIGABRT (application abort requested)
  /Users/runner/work/ompl/ompl/tests/base/state_spaces.cpp:106: last checkpoint: "Trochoid_Simple" test entry
  unknown location:0: fatal error: in "Owen_Simple": signal: illegal opcode; address of failing instruction: 0x7ff80186337b
  /Users/runner/work/ompl/ompl/tests/base/state_spaces.cpp:119: last checkpoint: "Owen_Simple" test entry
  
  *** 2 failures are detected in the test module "StateSpaces"
```

Jaeyoung-Lim:
> - Would it be possible to avoid some duplication of code w.r.t. the DubinsStateSpace by using inheritance? It seems like you probably don't need a new class for the motion validator. Look at what I did for the 3D Dubins spaces as an example.

@mamoll I am attempting to address this comment, but seems like I need to make quite some changes to the `DubinsStateSpace` and `ReedSheppStateSpace`. Would it be possible to address this comment on a followup PR?

Jaeyoung-Lim:
@mamoll I believe I have addressed all comments!

mamoll:
> @mamoll Would you have any clue on why the code would be failing on macOS 13 with the owen statespace?
> 
> ```
>   unknown location:0: fatal error: in "Trochoid_Simple": signal: SIGABRT (application abort requested)
>   /Users/runner/work/ompl/ompl/tests/base/state_spaces.cpp:106: last checkpoint: "Trochoid_Simple" test entry
>   unknown location:0: fatal error: in "Owen_Simple": signal: illegal opcode; address of failing instruction: 0x7ff80186337b
>   /Users/runner/work/ompl/ompl/tests/base/state_spaces.cpp:119: last checkpoint: "Owen_Simple" test entry
>   
>   *** 2 failures are detected in the test module "StateSpaces"
> ```

Is this still an issue? Maybe it is/was caused by very aggressive optimization flags? Running `valgrind` on this test might be helpful?

src/ompl/base/spaces/TrochoidStateSpace.h
```diff
@@ -0,0 +1,199 @@
...
+            bool isSymmetric_;
+        };
+
+        /** \brief A Dubins motion validator that only uses the state validity checker.
+            Motions are checked for validity at a specified resolution.
+
+            This motion validator is almost identical to the DiscreteMotionValidator
+            except that it remembers the optimal TrochoidPath between different calls to
+            interpolate. */
+        class TrochoidMotionValidator : public MotionValidator
```
mamoll: I think you don't need this class anymore.

src/ompl/base/spaces/src/TrochoidStateSpace.cpp
```diff
@@ -0,0 +1,730 @@
...
+
+    double trochoid_dely(double omega, double dir, double delt, double hInitRad ){
+    return 1.0/(dir*omega)*(-std::cos(dir*omega*delt+hInitRad) + std::cos(hInitRad));
+    }
+
+    double trochoid_delh(double omega, double dir, double delt ) {
+        return dir*omega*delt;
+    }
+
+    void computeEndpointBSB(double delta_1, double delta_2, double tA, double tBeta, double T, double phi0, double radius, double wind_ratio, double &xf, double &yf, double &phif) {
```
mamoll, line 85: some high-level documentation for computeEndpoints*, checkConditions*, trochoid* would be helpful (references to equations in papers would also be helpful)

src/ompl/base/spaces/src/TrochoidStateSpace.cpp
```diff
@@ -0,0 +1,730 @@
...
+
+        double phit1 = fmod(phi0, 2. * onepi);
+        double xt10 = x0 - delta_1 * radius * std::sin(phit1);
+        double yt10 = y0 + delta_1 * radius * std::cos(phit1);
+
+        double phit2 = fmod(phif - delta_2 * omega * t2pi, 2. * onepi);
+        double xt20 = xf - delta_2 * radius * std::sin(delta_2 * omega * t2pi + phit2) - V_w * t2pi;
+        double yt20 = yf + delta_2 * radius * cos(delta_2 * omega * t2pi + phit2);
+
+        bool solnFound = false;
```
mamoll: unused

src/ompl/base/spaces/src/TrochoidStateSpace.cpp
```diff
@@ -0,0 +1,730 @@
...
+
+        double xt10 = x0 - 1.0/(delta_1*omega)*sin(phit1);
+        double yt10 = y0 + 1.0/(delta_1*omega)*cos(phit1);
+
+        double delta_2 = - delta_1;
+
+        // test a grid of initial conditions, grid resolution
+        int numTestPts = 10;
+        std::vector< std::vector<double> > rootsComputed(numTestPts*numTestPts);
+        for (int i = 0; i < rootsComputed.size(); i++){
```
mamoll:
```suggestion
        for (std::size_t i = 0; i < rootsComputed.size(); i++){
```

src/ompl/base/spaces/src/TrochoidStateSpace.cpp
```diff
@@ -0,0 +1,730 @@
...
+            rootsComputed[i].resize(3);
+        }
+        double sameRootEpsilon = 0.001;
+        bool solnFound{false};
```
mamoll: unused

mamoll (commented):
The Trochoid space looks like it's doing some numerical root finding like the 3D Dubins classes. However, in the Throchoid space it is assumed that getPath() will always succeed. Is that true? You always know you have bracketed a root and you will converge to a solution in a fixed number of iterations? I started running `./build/demos/demo_GeometricCarPlanning --trochoid --distance` to see if this will trigger some failures, but this is very slow.

Have you been able to successfully solve any problems with `./build/demos/demo_GeometricCarPlanning --trochoid --easyplan`?
I couldn't, even after bumping the planning time up to 2 minutes.

src/ompl/base/spaces/src/TrochoidStateSpace.cpp
Jaeyoung-Lim: Removed

src/ompl/base/spaces/src/TrochoidStateSpace.cpp
Jaeyoung-Lim: Thanks! Done

src/ompl/base/spaces/src/TrochoidStateSpace.cpp
Jaeyoung-Lim: Removed

src/ompl/base/spaces/TrochoidStateSpace.h
Jaeyoung-Lim: Removed!

Jaeyoung-Lim:
@mamoll Thank you for spotting the anomaly in the distance computation.

There were some numerical issues in the solution checks which I fixed in the last commit.
Figure_1

Jaeyoung-Lim:
@mamoll There was a small kink in the above contour plot, which was caused by another problem that I fixed.
Figure_1

src/ompl/base/spaces/src/TrochoidStateSpace.cpp
Jaeyoung-Lim: Done!

src/ompl/base/spaces/TrochoidStateSpace.h
```diff
@@ -0,0 +1,173 @@
...
+                    length_[0] = tA;
+                    length_[1] = p;
+                    length_[2] = t_2pi - tB;
+                    assert(tA >= 0.);
+                    assert(p >= 0.);
+                    assert(t_2pi - tB >= 0.);
+                }
+                double length() const
+                {
+                    ///TODO: Fix for trochoid paths
```
mamoll: Do you plan to address this in this PR or in a later PR?

Jaeyoung-Lim: Ah, yes it should be resolved and therefore I have removed the comment.

The point that might be a little sticky is that the length returns the air-relative path length, not the geometric path length in the inertial space. However, I think this is the correct length value we want to have when doing trochoid path computations.

Jaeyoung-Lim:
> Can you check that these changes look good to you, too, and add my commits to your PR? There were two serious bugs (while (i<=N) should be while (i<N) in two places).

@mamoll Thank you for thoroughly reviewing the changes, and fixing the serious problem in the while loop.

> Note that the solution path you posted above is an approximate solution. I realize this is a bit confusing, so I changed the demo to only print the solution if an exact solution is found (and increased the planning time).

Sorry for missing this. I have just realized that now.

I have found out that the reason why trochoid path computation is so slow is due to the CCC path type computation taking much longer, compared to CSC paths. I am looking into whether I can do some classification to speed up the computation, but I think this needs more work(and a proof) to make it into the implementation. Therefore will probably be a follow-up PR.

mamoll:
I was able to get a complete solution. I am ok with merging this. Are these two items the remaining to-do you want to do in subsequent PRs?:
* Report correct path length for Trochoid paths.
* Improve performance

mamoll (approved):

PR #1245, opened 2025-03-17
Update README.md

saahu27:


mamoll (changes requested):
* The formatting should be done in a separate PR. Right now, the code is not all clang-formatted. I suggest you do this after the 1.7 release and then turn on the formatting CI workflow.
* The log directory seems like an accidental commit.

saahu27:
> * The formatting should be done in a separate PR. Right now, the code is not all clang-formatted. I suggest you do this after the 1.7 release and then turn on the formatting CI workflow.

Understood. will circle back to this after the release.

.github/workflows/format.yml
mosfet80, line 1: checkout@2 use deprecated node.js (version <20).

PR #1246, opened 2025-03-19
Restore legacy variables 

mamoll:
Rebased version of #1185.
Manually resolved a merge conflict.
See #1185 for details

### DIFF
--- a/omplConfig.cmake.in
+++ b/omplConfig.cmake.in
@@ -1,14 +1,4 @@
-# CMake OMPL config module
-#
-# Usage:
-# find_package(ompl @OMPL_VERSION_MAJOR@ REQUIRED)
-# add_library(your_app main.cpp)
-# target_link_libraries(your_app PRIVATE ompl::ompl)
-#
-#
-
 @PACKAGE_INIT@
-
 
 # These must be set BEFORE any call to find_package because PACKAGE_PREFIX_DIR gets overwritten by Eigen3.
 # See https://cmake.org/cmake/help/latest/module/CMakePackageConfigHelpers.html#command:configure_package_config_file.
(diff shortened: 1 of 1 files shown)


saahu27 (commented):
@Ryanf55  @mamoll I can confirm, The changes do not break the existing build. But as these changes are made to support legacy users, i am trying to understand how to verify these changes for legacy users. Please advice. Thank you.

saahu27:
Hi @Ryanf55  @mamoll  @MarqRazz  the changes i pushed to this branch works with moveit humble and jazzy without breaking any legacy code and also integrating it into latest distributions. i guess the changes made to moveit under ompl_interface_update is not necessary anymore. 

Can anybody verify this changes, so we can go ahead with the release? Thank you.

saahu27:
@mamoll  @Ryanf55 @MarqRazz  Any build system compiling a package looks for a dependency in the packages config.cmake.in. Because of the changes made to OMPL CMakeLists to modernize the package management, the look up in other packages should be like ompl::ompl. Since, older versions use ompl to look up, this broke the legacy systems. 

The above commit enables both the look ups to point to the right libraries. this way both the legacy systems and the new modern approach both work. 

I have tested the build and plan, both in humble and also jazzy versions of moveit. I believe this solves the issue we are trying to solve. 

Could anyone help verify this changes?

MarqRazz (approved):
Tested and it works with both MoveIt2 methods, `main` and [my PR](https://github.com/moveit/moveit2/pull/2994) :tada:.

Do we want to add any deprecation information here about the old method? Any notes you want to add to the [ReleaseNotes](https://github.com/ompl/ompl/blob/main/doc/markdown/releaseNotes.md)?

saahu27:
@MarqRazz Thank you for verifying the changes.  

This pull request gives access to deprecated methods. (find_package(ompl)).
I am working on the release, which include updates to website, release notes. i will make a mention of this in the release notes.

saahu27 (approved):

Issue #1253, opened 2025-03-26
Missing import in prerelease build of Python bindings


srmainwaring:
The latest prerelease build of the Python bindings is failing to import because of a missing import

- https://github.com/ompl/ompl/releases/tag/prerelease

Systems:
- Ubuntu 22.04 and Ubuntu 24.04

```bash
>>> from ompl import base as ob 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/venv/lib/python3.10/site-packages/ompl/base/__init__.py", line 1, in <module>
    from ompl import util
  File "/venv/lib/python3.10/site-packages/ompl/util/__init__.py", line 3, in <module>
    dll_loader('ompl', dirname(abspath(__file__)))
  File "/venv/lib/python3.10/site-packages/ompl/__init__.py", line 17, in dll_loader
    fpath = os.path.join(path, fname)
NameError: name 'os' is not defined
``` 

This was working with the previous version.

Suspect this commit: https://github.com/ompl/ompl/commit/381b27934e88cb91c82f3999fb407c6566a46b3c

https://github.com/ompl/ompl/blob/381b27934e88cb91c82f3999fb407c6566a46b3c/py-bindings/ompl/__init__.py#L16-L26

There are possibly other issues with this commit. Editing the `__init__.py` locally to `import os` then yields the following error:

```bash
>>> from ompl import base as ob
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/venv/lib/python3.10/site-packages/ompl/base/__init__.py", line 1, in <module>
    from ompl import util
  File "/venv/lib/python3.10/site-packages/ompl/util/__init__.py", line 3, in <module>
    dll_loader('ompl', dirname(abspath(__file__)))
  File "/venv/lib/python3.10/site-packages/ompl/__init__.py", line 23, in dll_loader
    for _fname in os.listdir(sitepackagedir):
FileNotFoundError: [Errno 2] No such file or directory: '/venv/local/lib/python3.10/dist-packages'
```

Suggested fix: seems that in a virtual environment `site.getsitepackages()` may return paths that do not exist:

```diff
def dll_loader(lib, path):
    from platform import system
    from os.path import isfile
    import ctypes
    from ctypes.util import find_library
+    import os
    import site

    # First, try the user-specified path
    sys = system()
    if sys == "Windows":
        ext = ".dll"
    elif sys == "Darwin":
        ext = ".dylib"
[...]

PR #1254, opened 2025-03-26
Fix dll_loader in py-bindings

srmainwaring:
Fix library import issue when using virtual environments

- Add missing import.
- Add check site package dir exists.

Fixes: https://github.com/ompl/ompl/issues/1253


saahu27 (approved):
In virtual environment, it does have a problem. thank you @srmainwaring for the fix.

PR #1264, opened 2025-04-04
Python bindings for OwenStateSpace::PathType member functions

srmainwaring:
This change forces the `pyplusplus` binding generator to expose the `OwenStateSpace::PathType::length()` member function, which is not otherwise generated.


mamoll:
I fixed the issue a slightly different (but more generic) way in PR #1265. Can you test if that works for you?

srmainwaring:
Closed in favour of https://github.com/ompl/ompl/pull/1265

PR #1265, opened 2025-04-05
python bindings for 3D Dubins PathType length method

mamoll:


mamoll:
@srmainwaring please check this out

srmainwaring:
@mamoll - works well thanks. Tested on the example I'd been using to reconstruct Dubins path segments and all working as expected.  

owenstatespace_segments

saahu27 (approved):

Issue #1270, opened 2025-04-08
Possible memory corruption in ompl-1.7 bindings


srmainwaring:
There may be an issue causing memory corruption in the latest binding library. Still investigating, but the issue occurs when using `pytest` to run Python test cases.

- Does occur on an Ubuntu 22.04 docker image using the published wheels (Python 3.10)
- Does not occur on macOS when the bindings are built from source against Python 3.13

## Details

Minimal example to replicate. Create a test case that imports the library:

```python
# test_ompl_import.py
def test_ompl_import()
    from ompl import base as ob

```

Run the test:

```bash
$ pytest ./test_ompl_import.py 
=============================================================== test session starts ===============================================================
platform linux -- Python 3.10.12, pytest-8.3.5, pluggy-1.5.0
rootdir: /terrain_nav_py
configfile: pyproject.toml
collected 1 item                                                                                                                                  

test_ompl_import.py .                                                                                                                       [100%]

================================================================ 1 passed in 0.14s ================================================================
corrupted size vs. prev_size while consolidating
Aborted
```

kylc:
I'm running into this as well. gdb complains of some illegal memory access of type `std::vector<DubinsPathSegmentType>`, which pointed me to this commit: https://github.com/ompl/ompl/commit/2e953fe8c620b72cbb87d3b90ee635c236371215. Dropping this commit only (everything else stays the same) causes the test to succeed, while adding it back in causes the test to fail.

Perhaps something about the RAII free ordering of the static vectors vs. the pointers into them?

Relevant stack trace:
```
#7  0x00007ffff66e831c in std::__new_allocator<ompl::base::DubinsStateSpace::DubinsPathSegmentType>::deallocate (this=0x49b970, __n=<optimized out>, 
    __p=<optimized out>) at /nix/store/yg4ahy7gahx91nq80achmzilrjyv0scj-gcc-13.3.0/include/c++/13.3.0/bits/new_allocator.h:172
#8  std::allocator_traits<std::allocator<ompl::base::DubinsStateSpace::DubinsPathSegmentType> >::deallocate (__n=<optimized out>, __p=<optimized out>, 
    __a=...) at /nix/store/yg4ahy7gahx91nq80achmzilrjyv0scj-gcc-13.3.0/include/c++/13.3.0/bits/alloc_traits.h:517
[...]

PR #1271, opened 2025-04-09
Fix static initialisation / destruction issue for Dubins path types

srmainwaring:
Provide safe initialisation and no-destruct for `dubinsPath` object with static storage.

- Closes: https://github.com/ompl/ompl/issues/1270

**Edit** Convert to draft as getting a seg fault on release build on Ubuntu 22.04 in docker. No issue on macOS.

### DIFF
--- a/src/ompl/base/spaces/src/DubinsStateSpace.cpp
+++ b/src/ompl/base/spaces/src/DubinsStateSpace.cpp
@@ -260,7 +260,7 @@ namespace
             assert(fabs(p * cos(alpha + t) - sa + sb - d) < (1 + p) * DUBINS_EPS);
             assert(fabs(p * sin(alpha + t) + ca - cb) < (1 + p) * DUBINS_EPS);
             assert(mod2pi(alpha + t + q - beta + .5 * DUBINS_EPS) < DUBINS_EPS);
-            return DubinsStateSpace::DubinsPath(&DubinsStateSpace::dubinsPathType()[0], t, p, q);
+            return DubinsStateSpace::DubinsPath(DubinsStateSpace::dubinsPathType()[0], t, p, q);
         }
         return {};
     }
@@ -278,7 +278,7 @@ namespace
             assert(fabs(p * cos(alpha - t) + sa - sb - d) < (1 + p) * DUBINS_EPS);
             assert(fabs(p * sin(alpha - t) - ca + cb) < (1 + p) * DUBINS_EPS);
             assert(mod2pi(alpha - t - q - beta + .5 * DUBINS_EPS) < DUBINS_EPS);
-            return DubinsStateSpace::DubinsPath(&DubinsStateSpace::dubinsPathType()[1], t, p, q);
+            return DubinsStateSpace::DubinsPath(DubinsStateSpace::dubinsPathType()[1], t, p, q);
         }
         return {};
     }
@@ -296,7 +296,7 @@ namespace
             assert(fabs(p * cos(alpha - t) - 2. * sin(alpha - t) + sa + sb - d) < 2 * DUBINS_EPS);
             assert(fabs(p * sin(alpha - t) + 2. * cos(alpha - t) - ca - cb) < 2 * DUBINS_EPS);
             assert(mod2pi(alpha - t + q - beta + .5 * DUBINS_EPS) < DUBINS_EPS);
-            return DubinsStateSpace::DubinsPath(&DubinsStateSpace::dubinsPathType()[2], t, p, q);
+            return DubinsStateSpace::DubinsPath(DubinsStateSpace::dubinsPathType()[2], t, p, q);
         }
         return {};
     }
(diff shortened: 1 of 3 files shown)


src/ompl/base/spaces/src/DubinsStateSpace.cpp
```diff
@@ -712,21 +712,26 @@ namespace ompl::base
...
-const std::vector<std::vector<DubinsStateSpace::DubinsPathSegmentType> > DubinsStateSpace::dubinsPathType {{
-    {{DUBINS_LEFT, DUBINS_STRAIGHT, DUBINS_LEFT}},
-    {{DUBINS_RIGHT, DUBINS_STRAIGHT, DUBINS_RIGHT}},
-    {{DUBINS_RIGHT, DUBINS_STRAIGHT, DUBINS_LEFT}},
-    {{DUBINS_LEFT, DUBINS_STRAIGHT, DUBINS_RIGHT}},
-    {{DUBINS_RIGHT, DUBINS_LEFT, DUBINS_RIGHT}},
-    {{DUBINS_LEFT, DUBINS_RIGHT, DUBINS_LEFT}}
-  }};
+const std::vector<std::vector<DubinsStateSpace::DubinsPathSegmentType> >& DubinsStateSpace::dubinsPathType() {
+  static std::vector<std::vector<DubinsStateSpace::DubinsPathSegmentType> >* pathType
```
mamoll, line 722: Can't this be `std::vector<std::vector<DubinsStateSpace::DubinsPathSegmentType> >` (to avoid heap allocation) and return `pathType`?

srmainwaring: The idea discussed in this article (https://ppwwyyxx.com/blog/2023/Safe-Static-Initialization-No-Destruction), is that the destructor is never called, to ensure that the static storage object does not go out of scope for any other objects referencing it at library unload. Not using a pointer will result in the same issue we currently have, unless we use the [`NoDestructor<>`](https://chromium.googlesource.com/chromium/src/base/+/40343e3fbc1e4835145f125adedcf15a2eb4542f/no_destructor.h) wrapper referred to, which we could do at the cost of introducing the extra template class. I though the pointer to be the simpler of the two, but if heap is a concern I can add the template.

srmainwaring: This is what the `NoDestructor<>` version would look like:
https://github.com/srmainwaring/ompl/tree/prs/pr-dubins-path-static-init-v2

The `NoDestructor<>` is reproduced directly from the link above, I've included the original copyright notice - may need updating to comply with the copyright / license notice for OMPL.

mamoll: ok, I only skimmed the article you linked to the first time. I read it more carefully now. This solution makes sense. There is not a big additional benefit to the `NoDestructor<>` solution, so I'm fine with merging this.

mamoll (approved):
I'll merge if you remove the "draft" status.

mamoll:
One possible change to consider: change this signature:
```
DubinsPath(const std::vector<DubinsPathSegmentType> *type = &dubinsPathType()[0],
                  double t = 0., double p = std::numeric_limits<double>::max(), double q = 0.);
```
to this:
```
DubinsPath(const std::vector<DubinsPathSegmentType>& type = dubinsPathType()[0],
                  double t = 0., double p = std::numeric_limits<double>::max(), double q = 0.);
```
(I.e., use reference rather than a raw C-style pointer.)

srmainwaring:
> One possible change to consider: change this signature:

Good suggestion. Updated in https://github.com/ompl/ompl/pull/1271/commits/be319bc6250716764d3c2b1826244d4e257a4cdf.

> I'll merge if you remove the "draft" status.

I moved to draft as I am seeing a segmentation fault in my Ubuntu 22.04 docker build. This may be because I am using `clang++` to build on `x86_64`, so I'll test with `gcc`. 

In the meanwhile, would it be possible to have a very simple `pytest` included in the CI to catch the type of error this PR is trying to resolve. I'm checking locally on macOS, but am concerned that my docker build is not exactly reproducing the CI configuration.

srmainwaring:
> This is concerning, because I had to use clang++ on Linux

I've built and run the Dockerfile `./scripts/docker/ompl.Dockerfile`, and the resulting bindings are not exhibiting any issue, so not quite sure what is going on with my image. But this change seems ok.

I had to do the following to get the image to build and work (using Docker Desktop on macOS)

```bash
# build the image
cd  ~/ompl
cp ./scripts/docker/ompl.Dockerfile ./Dockerfile
docker build -t ompl:latest . 

# run the image as root to be able to install python3-pytest
docker run -it --rm -u 0 ompl:latest 
cat > test_import.py <<EOL
def test_import():
    from ompl import base as ob
EOL
apt install python3-pytest
pytest ./test_import.py
```

PR #1273, opened 2025-04-15
first attempt to clean up build of python wheels, update docs now that wheels are on PyPI

mamoll:


mamoll:
@kylc I am trying to get OMPL wheels to have a bit more metadata set in the wheels, so that the [OMPL page on PyPI](https://pypi.org/project/ompl/) doesn't look so bare. Is the pyproject.toml file the right place to fill  out? Right now the version is hard-coded in setup.py. Should I use cmake's `configure_file` on setup.py.in to do this automatically or is there a better way?

For now I'm uploading the wheels manually, but I have a twine access token that I can specify as a secret in Github so that Gihub CI can publish directly to PyPI. That's a separate issue for later, though.

kylc:
> Is the pyproject.toml file the right place to fill out?

I think so. My memory is that PyPI pulls the metadata from the pyproject.toml file to display. Mainly, it follows the README path to fill out the project description.

> Right now the version is hard-coded in setup.py. Should I use cmake's `configure_file` on setup.py.in to do this automatically or is there a better way?

That seems like a reasonable approach. Maybe also interesting: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#static-vs-dynamic-metadata. With this method it seems like you can write custom code in `setup.py` to determine the version.

> For now I'm uploading the wheels manually, but I have a twine access token that I can specify as a secret in Github so that Gihub CI can publish directly to PyPI. That's a separate issue for later, though.

As a reference, I have an automated pipeline for doing this using Rust's wheel build tool `maturin` here: https://github.com/kylc/optik/blob/master/.github/workflows/python.yaml#L101-L116. It should be mostly the same with `twine`.

---

BTW to run the wheel job on this branch you should be able to manually dispatch it. The wheel job has the `workflow_dispatch` marker needed for https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow

py-bindings/pyproject.toml
```diff
@@ -1,13 +1,31 @@
+[project]
+name = "ompl"
+version = "1.7.0"
+description = "The Open Motion Planning Library"
+authors=[
+    {name = "Ioan A. Șucan", email = "ioan@sucan.ro"},
+    {name = "Mark Moll", email = "mark@moll.ai"},
+    {name = "Zachary Kingston", email = "zkingston@purdue.edu"},
+    {name = "Lydia E. Kavraki", email = "kavraki@rice.edu"}
+]
```
kylc, line 10: Consider adding something like:
```
classifiers = [
  "Topic :: Scientific/Engineering"
]
```

There is a full list of classifiers somewhere.

.github/workflows/wheels.yaml
```diff
@@ -79,7 +79,12 @@ jobs:
     name: Publish to PyPI
     runs-on: ubuntu-latest
     needs: build_wheels
-    if: startsWith(github.ref, 'refs/tags/')
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
+    environment:
+      name: pypi
+      url: https://pypi.org/p/ompl
+    permissions:
+      id-token: write
```
kylc, line 87: nit: does this do anything within this PR, or is it just setting up for future work?

kylc (approved):
LGTM! `pip install ompl` is a nice upgrade.

.github/workflows/wheels.yaml
mamoll: It's set-up for the next release. When we tag 1.7.1, I'd like to get the auto-push to PyPI working.

PR #1278, opened 2025-05-08
more cleanup, wheel fixes

mamoll:
Wheels build now. See https://github.com/ompl/ompl/actions/runs/14904214635/job/41862792999

You can download them and try it out:
Screenshot 2025-05-08 at 6 10 07 AM

### DIFF
--- a/bindings/setup.py
+++ b/bindings/setup.py
@@ -48,6 +48,9 @@ class CMakeBuild(build_ext):
         # - Set CMAKE_LIBRARY_OUTPUT_DIRECTORY to a location expected by the
         #   wheel builder so that libraries are included in the output artifact.
         # - Disable unnecessary targets.
+        # - Disable generation of symlinked shared library versioning like
+        #   "libompl.so" -> "libompl.1.8.0.so" -> etc. These will result in
+        #   duplicated files in the wheel artifact and aren't necessary.
         #
         # [1]: https://cmake.org/cmake/help/latest/module/FindPython.html#hints
         cmake_args = [
@@ -61,6 +64,7 @@ class CMakeBuild(build_ext):
             "-DOMPL_BUILD_DEMOS=OFF",
             "-DOMPL_BUILD_PYTESTS=OFF",
             "-DOMPL_BUILD_TESTS=OFF",
+            "-DCMAKE_PLATFORM_NO_VERSIONED_SONAME=ON"
         ]
         build_args = []
         # Adding CMake arguments set as environment variable


mamoll:
@WeihangGuo have you had a chance to look into this?

mamoll:
@kylc we are working on new Python bindings using nanobind, manually created (by @WeihangGuo). I am trying (unsuccessfully) to update your wheel build magic for this code. Everything builds and runs locally, but somehow the wheel-building code does not pick up the `_ompl.xxxx.so`. Could you have a quick look to see if I am missing something obvious?

kylc:
FYI I have a few WIP fixes based on this branch here: https://github.com/ompl/ompl/compare/pybinding-wheels...kylc:ompl:pybinding-wheels

- Add hints for CMake to pick up the right Python target
- Add missing `typing_extensions` dependency for older Python versions
- Locate Python with `Development.SABIModule` so that nanobind automatically builds `abi3`-compatible wheels for recent Python versions
 - Use Boost from the repos on Linux rather than building locally. This speeds up CI wheel builds a bunch since Boost had to be built for N different Python versions. Boost 1.66 is the latest available in the repos, which is <1.68 required right now. However, it seems to work. Maybe the current 1.68 version requirement is too conservative?
 - Build all Python target wheels in parallel, which again speeds up the overall build a bunch. See an example build here: https://github.com/kylc/ompl/actions/runs/15655319501

mamoll:
Thanks for doing this!

> Boost 1.66 is the latest available in the repos, which is <1.68 required right now. However, it seems to work. Maybe the current 1.68 version requirement is too conservative?

I see in your example build that the Boost version it pulls in is indeed 1.66, which seems very old for Ubuntu 24.04. It must be something about the cibuildwheel/manylinux/yum build/install process I don't fully understand.

I looked at the git log for when this Boost version was last bumped, which was through this PR:
https://github.com/ompl/ompl/pull/1219
The author claims this was necessary to support compilation in C++17 mode on MacOS. However, everything seems to compile in your build. You'd have to go out of your way to install boost <1.68 on MacOS as all MacOS package managers default to something newer. So I think it's ok to lower min. req. version, even if it's not 100% accurate.

I downloaded the wheel for macos13 / python3.13. I can pip install it and import it, but there may be some redundant copies of libompl:
```
> find ~/Library/Python/3.13/lib/python/site-packages -name \*ompl\*dylib
/Users/mmoll/Library/Python/3.13/lib/python/site-packages/libompl.18.dylib
/Users/mmoll/Library/Python/3.13/lib/python/site-packages/libompl.dylib
/Users/mmoll/Library/Python/3.13/lib/python/site-packages/libompl.1.8.0.dylib
/Users/mmoll/Library/Python/3.13/lib/python/site-packages/ompl/.dylibs/libompl.1.8.0.dylib
```

kylc:
> I see in your example build that the Boost version it pulls in is indeed 1.66, which seems very old for Ubuntu 24.04. It must be something about the cibuildwheel/manylinux/yum build/install process I don't fully understand.
> 
> I looked at the git log for when this Boost version was last bumped, which was through this PR: #1219 The author claims this was necessary to support compilation in C++17 mode on MacOS. However, everything seems to compile in your build. You'd have to go out of your way to install boost <1.68 on MacOS as all MacOS package managers default to something newer. So I think it's ok to lower min. req. version, even if it's not 100% accurate.

Maybe a conditional boost requirement, >=1.66 for linux and >=1.68 for MacOS?

The reason for 1.66 is because the `manylinux_2_28` image we use is based on AlmaLinux 8 which is quite old. There is a new `manylinux_2_34` image in alpha now which would give us to more up-to-date packages, but would also raise the glibc requirement for OMPL users (those are the numbers in the image name). Not worth it, I think.

The official images come from here: https://github.com/pypa/manylinux. There is also a common distro compatibility list for wheels built from a particular image which makes it easier to choose one.

> I downloaded the wheel for macos13 / python3.13. I can pip install it and import it, but there may be some redundant copies of libompl:
> 
> ```
> > find ~/Library/Python/3.13/lib/python/site-packages -name \*ompl\*dylib
> /Users/mmoll/Library/Python/3.13/lib/python/site-packages/libompl.18.dylib
> /Users/mmoll/Library/Python/3.13/lib/python/site-packages/libompl.dylib
> /Users/mmoll/Library/Python/3.13/lib/python/site-packages/libompl.1.8.0.dylib
> /Users/mmoll/Library/Python/3.13/lib/python/site-packages/ompl/.dylibs/libompl.1.8.0.dylib
> ```

It looks like these are placed as symlinks in the regular build process, and then perhaps symlinks aren't supported in the ZIP archive so they are just copied as files. They don't seem to add any overall size to the archive -- it's still just ~10MB where each .dylib is ~8MB on its own. I don't know enough about MacOS to say if these can be removed or not, seems safer just to leave them.

```
 [247/463] Creating library symlink /Users/runner/work/ompl/ompl/bindings/build/lib.macosx-10.9-x86_64-cpython-313/libompl.18.dylib /Users/runner/work/ompl/ompl/bindings/build/lib.macosx-10.9-x86_64-cpython-313/libompl.dylib
```

kylc:
FYI https://github.com/ompl/ompl/pull/1293. This includes a few more changes than those which you have already reviewed @mamoll.

Note that I've only done very minimal testing on these output wheel artifacts.

> It's fine to have some redundant copies, I guess.

Found a solution to this by setting `CMAKE_PLATFORM_NO_VERSIONED_SONAME` when building wheels.

PR #1302, opened 2025-08-02
Fix bug in ST-RRT* and add unit tests

aorthey:
This MR fixes a bug in ST-RRT*, where a non-zero start time lead to an increase in planning time. This was found by https://github.com/NiklasHargus. 

Some unit tests have been added by grok to verify the correct behavior.


mamoll (approved):

mamoll:
Looks good to me. Thanks @aorthey and @NiklasHargus!

PR #1304, opened 2025-08-07
correct maxextent in sphere statespace

mamoll:
basic bug where default max distance implementation is not correct for a sphere.


saahu27 (approved):
Looks fine to me.

PR #1311, opened 2025-08-24
vamp arm aarch flags

saahu27:
## Problem
The `test_vamp_integration` test was failing with "Not Run" status on Ubuntu 24.04-arm-aarch64 GitHub Actions builds due to conflicting compiler flags between OMPL's VAMP configuration and the VAMP submodule's own compiler settings.


Thodoris1999 (approved):
lgtm!

PR #1325, opened 2025-09-30
Vamp state space integration

saahu27:


src/ompl/base/spaces/VAMPStateSpace.h
```diff
@@ -0,0 +1,264 @@
...
+#include <vamp/robots/ur5.hh>
+#include <vamp/robots/fetch.hh>
+#include <vamp/robots/baxter.hh>
+
+// Standard library
+#include <memory>
+#include <array>
+#include <string>
+
+namespace ompl { namespace geometric {
```
mamoll:
```suggestion
namespace ompl::base {
```
Need to make code changes elsewhere, too, if you accept this change.

demos/Vamp.cpp
```diff
@@ -0,0 +1,330 @@
...
+    std::cout << "\n2. Creating VAMP state space..." << std::endl;
+    auto space = std::make_shared<geometric::VAMPStateSpace<vamp::robots::Panda>>(environment);
+    std::cout << "   State space: " << space->getName() << " (dimension: " << space->getDimension() << ")" << std::endl;
+    
+    // 3. Create SimpleSetup and configure validators
+    std::cout << "\n3. Creating SimpleSetup..." << std::endl;
+    geometric::SimpleSetup ss(space);
+    
+    auto si = ss.getSpaceInformation();
+    ss.setStateValidityChecker(space->allocDefaultStateValidityChecker(si));
```
mamoll:
```suggestion
    si->setStateValidityChecker(space->allocDefaultStateValidityChecker(si));
```
(to make it analogous to the next line)

mamoll (commented):
Looks much better!

src/ompl/base/spaces/VAMPStateSpace.h
```diff
@@ -0,0 +1,264 @@
...
+        
+        // Set robot joint limits
+        base::RealVectorBounds bounds(dimension);
+        for (std::size_t i = 0; i < dimension; ++i) {
+            bounds.setLow(i, Robot::s_a[i]);
+            bounds.setHigh(i, Robot::s_a[i] + Robot::s_m[i]);
+        }
+        setBounds(bounds);
+        
+        setName("VAMP_" + std::string(Robot::name));
```
mamoll, line 235: add:
```
type_ = STATE_SPACE_VAMP;
```

and add this:
```diff
diff --git a/src/ompl/base/StateSpaceTypes.h b/src/ompl/base/StateSpaceTypes.h
index 2cf952eea..e69456a19 100644
--- a/src/ompl/base/StateSpaceTypes.h
+++ b/src/ompl/base/StateSpaceTypes.h
@@ -98,7 +98,10 @@ namespace ompl
 
             /** \brief ompl::base::TrochoidStateSpace */
             STATE_SPACE_TROCHOID = 17,

+            /** \brief ompl::base::VAMPStateSpace */
+            STATE_SPACE_VAMP = 18;
+
             /** \brief Number of state space types; To add new types,
                 use values that are larger than the count*/
             STATE_SPACE_TYPE_COUNT
```

demos/Vamp.cpp
zkingston, line 1: The name should probably be lowercase.

demos/Vamp.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+    2. Benchmarking Mode (--benchmark):
+       - Compares performance across multiple planners (RRTConnect, RRTstar, BITstar, PRM)
+       - Generates benchmark logs compatible with Planner Arena
+       - Shows how to use VAMP with OMPL's benchmarking infrastructure
+
+    The VAMPStateSpace provides automatic configuration of:
+    - Joint limits from robot definitions
+    - SIMD-accelerated state validity checking
+    - Vectorized continuous collision detection
```
zkingston: These are basically the same thing, would just say the first. It's also not fully continuous CC.

demos/Vamp.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+    All OMPL geometric planners work seamlessly with VAMPStateSpace without modification.
+
+    For more information on VAMP, see:
+    Wil Thomason, Zachary Kingston, and Lydia E. Kavraki,
+    "VAMP: Motions in Microseconds via Vectorized Sampling-Based Planning,"
```
zkingston: Paper doesn't actually have the acronym in the front.

demos/Vamp.cpp
```diff
@@ -0,0 +1,330 @@
...
+ * 4. Simplifies the solution path
+ *
+ * This serves as a minimal working example for integrating VAMP with OMPL.
+ */
+void runSimpleDemo() {
+    std::cout << "VAMP-OMPL Simple Planning Demo" << std::endl;
+    std::cout << "===============================" << std::endl;
+    
+    // 1. Create VAMP collision environment
+    std::cout << "\n1. Creating collision environment..." << std::endl;
```
zkingston: Very minor gripe, buy why mix `\n` and std::endl?

src/ompl/base/spaces/VAMPStateSpace.h
zkingston: Agree!

src/ompl/base/spaces/VAMPStateSpace.h
```diff
@@ -0,0 +1,264 @@
...
+       si->setMotionValidator(space->allocDefaultMotionValidator(si));
+
+       // Set start/goal and plan using any OMPL geometric planner
+       @endcode
+
+       @par External Documentation
+       The following paper describes VAMP's vectorization methodology and performance characteristics:
+
+       Wil Thomason, Zachary Kingston, and Lydia E. Kavraki,
+       "VAMP: Motions in Microseconds via Vectorized Sampling-Based Planning,"
```
zkingston: Same comment as in other file, no acronym at front

demos/Vamp.cpp
```diff
@@ -0,0 +1,330 @@
...
+ *  CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT
+ *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
+ *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ *  POSSIBILITY OF SUCH DAMAGE.
+ *********************************************************************/
+
+/* Author: Sahruday Patti */
+
+/** @file Vamp.cpp
+    @brief Demonstration of VAMP-OMPL integration for SIMD-accelerated motion planning.
```
claytonwramsey, line 38: Maybe also link to the VAMP repository, since that seems to be the de facto landing page for the project.

demos/Vamp.cpp
```diff
@@ -0,0 +1,330 @@
...
+ *
+ * This function demonstrates basic usage of VAMPStateSpace for motion planning:
+ * 1. Creates a sphere cage environment
+ * 2. Sets up a VAMPStateSpace for the Panda robot
+ * 3. Plans paths using RRTConnect
+ * 4. Simplifies the solution path
+ *
+ * This serves as a minimal working example for integrating VAMP with OMPL.
+ */
+void runSimpleDemo() {
```
claytonwramsey, line 127: Do we have any visualization or other way to check that this is all correct? I want to make sure we don't have the robot accidentally phasing through obstacles.

doc/markdown/demos.md
```diff
@@ -2,7 +2,12 @@
...
+- **[VAMP Integration: SIMD-Accelerated Motion Planning](Vamp_8cpp_source.html)**. This demo showcases how to use ompl::geometric::VAMPStateSpace for high-performance manipulator planning using SIMD-accelerated collision detection. VAMP (Vector Accelerated Motion Planning) achieves planning speeds orders of magnitude faster than traditional methods by processing multiple robot configurations simultaneously using vectorized instructions (AVX2/NEON). The demo includes both simple planning and benchmarking modes, supporting popular robots including Panda (Franka Emika), UR5, Fetch, and Baxter. Command-line usage:
```
claytonwramsey: maybe want to make sure the formatting matches the other entries

doc/markdown/demos.md
```diff
@@ -2,7 +2,12 @@
...
+
+        > ./demo_Vamp --simple      # Run simple planning demo
```
claytonwramsey: probably use

````md
```sh
```
````

instead of `>`

File: doc/markdown/releaseNotes.md [hunk 12]
```diff
@@ -1,18 +1,19 @@
...
-  - Includes comprehensive demo (`vampdemo`) showcasing integration with popular robots (Panda, UR5, Fetch).
-  - Build option `OMPL_BUILD_VAMP=ON` enables VAMP integration as a git submodule and is enabled by default.
-  - Supports both portable builds (`VAMP_PORTABLE_BUILD=ON`) for distribution and native builds for maximum performance.
-  - Compatible with existing OMPL planners as a drop-in acceleration layer for motion validation
-- New kinodynamic planner: HySST, and adaptation of the SST planner for hybrid systems.
-- New state space: ThrochoidStateSpace, an SE(2)-like state space where distance and interpolation is defined for Dubins vehicles subject to constant drift. This is useful in planning for aerial/underwater drones subject to constant wind/current.
-- Bug fixes.
+  - Added ompl::geometric::VAMPStateSpace, a new state space with SIMD-accelerated collision detection for manipulator planning
+  - Integrates [VAMP (Vector-Accelerated Motion Planning)](https://github.com/KavrakiLab/vamp) library for vectorized collision checking and motion validation
+  - Achieves planning speeds orders of magnitude faster than traditional methods using AVX2/NEON SIMD instructions
```
claytonwramsey: I recommend removing some of these bullets. Changelogs aren't press releases, so they should be as concise as possible as far as which breaking changes occur and which things are added.

doc/markdown/releaseNotes.md
```diff
@@ -1,18 +1,19 @@
 # Release Notes {#releaseNotes}
 
-## OMPL 2.0.0beta (September 15th, 2025)
+## OMPL 2.0.0beta (October 9th, 2025)
```
claytonwramsey, line 3: again, should this be 1.8.0? it's a non-breaking release (I hope).

src/ompl/base/spaces/VAMPStateSpace.h
```diff
@@ -0,0 +1,264 @@
...
+       @anchor gVAMP
+       @par Short Description
+       VAMPStateSpace provides SIMD-accelerated collision detection for manipulator motion planning by
+       integrating the VAMP (Vector Accelerated Motion Planning) library into OMPL. VAMP leverages
+       modern CPU vectorization (AVX2, NEON) to perform collision checking orders of
+       magnitude faster than traditional methods. The state space automatically configures joint limits
+       from robot definitions and provides optimized state validity checking and motion validation.
+
+       VAMPStateSpace supports both primitive collision objects (spheres, boxes, capsules) and point
+       cloud environments through VAMP's Collision-Affording Point Tree (CAPT) data structure, enabling
```
claytonwramsey, line 77: rephrase, makes it seem like CAPT supports primitives

src/ompl/base/spaces/VAMPStateSpace.h
```diff
@@ -0,0 +1,264 @@
...
+
+       // Set start/goal and plan using any OMPL geometric planner
+       @endcode
+
+       @par External Documentation
+       The following paper describes VAMP's vectorization methodology and performance characteristics:
+
+       Wil Thomason, Zachary Kingston, and Lydia E. Kavraki,
+       "VAMP: Motions in Microseconds via Vectorized Sampling-Based Planning,"
+       arXiv preprint arXiv:2309.14545, 2024.
```
claytonwramsey, line 106: not an arXiv preprint, is an ICRA paper

src/ompl/base/spaces/VAMPStateSpace.h
```diff
@@ -0,0 +1,264 @@
...
+
+    public:
+        StateValidityChecker(const base::SpaceInformationPtr& si,
+                           const VectorizedEnvironment& env,
+                           const VAMPStateSpace* space)
+            : base::StateValidityChecker(si), environment_(env), space_(space) {}
+
+        bool isValid(const base::State* state) const override {
+            auto config = space_->convertOMPLToVAMP(state);
+            return vamp::planning::validate_motion<Robot, simd_width, 1>(
```
claytonwramsey, line 149: why `validate_motion` instead of `validate_vector`?

src/ompl/base/spaces/VAMPStateSpace.h
```diff
@@ -0,0 +1,264 @@
...
+                config, config, environment_);
+        }
+    };
+
+    /** \brief Motion validator using VAMP's vectorized continuous collision detection.
+     *
+     * This validator uses VAMP's SIMD-optimized swept-sphere collision detection to verify
```
claytonwramsey: collision checking is not using swept spheres (more precisely, it uses intermediate state probing)

src/ompl/base/spaces/VAMPStateSpace.h
```diff
@@ -0,0 +1,264 @@
...
+     *
+     * This state space extends RealVectorStateSpace to provide high-performance collision
+     * detection for manipulator robots using SIMD vectorization. Joint limits are automatically
+     * configured from the robot definition, and custom validators leverage VAMP's optimized
+     * collision checking algorithms.
+     *
+     * @tparam Robot VAMP robot type (e.g., vamp::robots::Panda, vamp::robots::UR5)
+     */
+template<typename Robot>
+class VAMPStateSpace : public base::RealVectorStateSpace {
```
claytonwramsey: using `RealVectorStateSpace` isn't the worst choice in the world, but it means that you're pinned to a lot of slightly-incompatible decisions. Maybe avoid inheriting everything so you can have floats instead of constantly converting between doubles and floats (among other things)?

src/ompl/base/spaces/VAMPStateSpace.h
```diff
@@ -0,0 +1,264 @@
...
+     */
+    base::StateValidityCheckerPtr allocDefaultStateValidityChecker(
+        const base::SpaceInformationPtr& si) const {
+        return std::make_shared<StateValidityChecker>(si, vectorized_environment_, this);
+    }
+
+    /** \brief Allocate the VAMP motion validator.
+     *
+     * This method provides VAMP's vectorized continuous collision checker instead of the
+     * standard motion validator. The returned validator uses swept-sphere collision detection
```
claytonwramsey: does not use swept spheres

src/ompl/base/spaces/VAMPStateSpace.h
```diff
@@ -0,0 +1,264 @@
...
+#define OMPL_BASE_SPACES_VAMP_STATE_SPACE_
+
+// OMPL includes
+#include <ompl/base/spaces/RealVectorStateSpace.h>
+#include <ompl/base/StateValidityChecker.h>
+#include <ompl/base/MotionValidator.h>
+#include <ompl/base/SpaceInformation.h>
+
+// VAMP includes
+#include <vamp/collision/environment.hh>
```
claytonwramsey, line 52: maybe add some `#ifdef` gating to give a friendly error in case people accidentally include this header without the optional VAMP dependency?

tests/base/state_spaces.cpp
```diff
@@ -691,3 +700,139 @@ BOOST_AUTO_TEST_CASE(KleinBottle_Simple)
...
+    
+    auto si = std::make_shared<base::SpaceInformation>(space);
+    si->setStateValidityChecker(space->allocDefaultStateValidityChecker(si));
+    si->setMotionValidator(space->allocDefaultMotionValidator(si));
+    si->setup();
+    
+    // Verify CAPT collision checking works
+    base::ScopedState<> state(space);
+    state.random();
+    si->isValid(state.get());  // Should execute without errors
```
claytonwramsey, line 805: why should we believe that a random state should always be valid?

src/ompl/base/spaces/VAMPStateSpace.h
saahu27: you’re right that our demos/tests only include this header under OMPL_HAVE_VAMP. However, the header itself is installed and can be included by downstream users even when OMPL was built without VAMP. In that case they’d hit errors.

I added a simple guard so users get a clear, actionable error if they include this header without the optional dependency:
```cpp
// at the top of VAMPStateSpace.h, right after the include guard
#include "ompl/config.h"
#if !OMPL_HAVE_VAMP
#  error "ompl::geometric::VampStateSpace requires OMPL built with VAMP (OMPL_HAVE_VAMP=ON)."
#endif
```

tests/base/state_spaces.cpp
saahu27: We don’t assume random states are valid in these tests. We only verify that the validity pipeline executes without crashing.

If preferred, I can adjust the test to:
Assert true only for empty environments: BOOST_CHECK(si->isValid(state.get()));

PR #1331, opened 2025-10-24
add build option to control shared/static build

mamoll:
this doesn't change any of the defaults, but makes it possible to create a static build of libompl with a cmake option.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -153,6 +153,12 @@ endif()
 # R is needed for running Planner Arena locally
 find_program(R_EXEC R)
 
+# default to building shared libs except Windows
+if (MSVC)
+    option(OMPL_BUILD_SHARED "Build OMPL as a shared library" OFF)
+else()
+    option(OMPL_BUILD_SHARED "Build OMPL as a shared library" ON)
+endif()
 add_subdirectory(src)
 add_subdirectory(py-bindings)
 
--- a/src/ompl/CMakeLists.txt
+++ b/src/ompl/CMakeLists.txt
@@ -50,18 +50,24 @@ target_link_libraries(ompl
     PRIVATE
         "$<$<BOOL:${OMPL_HAVE_SPOT}>:Spot::Spot>")
 
-if (MSVC)
-    set_target_properties(ompl PROPERTIES VERSION "${PROJECT_VERSION}" STATIC_LIBRARY_FLAGS "psapi.lib ws2_32.lib")
-else (MSVC)
+set_target_properties(ompl PROPERTIES VERSION "${PROJECT_VERSION}")
+
+if (OMPL_BUILD_SHARED)
+    set_target_properties(ompl PROPERTIES SOVERSION "${OMPL_ABI_VERSION}")
     if (MINGW)
         target_link_libraries(ompl psapi ws2_32)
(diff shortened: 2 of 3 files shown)


src/ompl/CMakeLists.txt
saahu27, line 64: When building as STATIC, this copies a .a file but names it with CMAKE_SHARED_LIBRARY_SUFFIX (.so), which is incorrect and will break Python bindings.

But i think we never supported python bindings on windows?

May be ? 
```cpp
if (NOT MSVC AND OMPL_BUILD_SHARED)
    add_custom_command(TARGET ompl POST_BUILD
        COMMAND "${CMAKE_COMMAND}" -E copy "$<TARGET_FILE:ompl>"
        "${CMAKE_CURRENT_SOURCE_DIR}/../../py-bindings/ompl/util/libompl${CMAKE_SHARED_LIBRARY_SUFFIX}")
endif()
```

saahu27, line 53: Problem:
STATIC_LIBRARY_FLAGS should only be set for static libraries
SOVERSION should only be set for shared libraries
The current logic bases this on MSVC, not on library type

May be ?

```cpp
set_target_properties(ompl PROPERTIES VERSION "${PROJECT_VERSION}")

if (OMPL_BUILD_SHARED)
    set_target_properties(ompl PROPERTIES SOVERSION "${OMPL_ABI_VERSION}")
    if (MINGW)
        target_link_libraries(ompl psapi ws2_32)
        set_target_properties(ompl PROPERTIES LINK_FLAGS "-Wl,--export-all-symbols")
    endif()
else()
    if (MSVC)
        set_target_properties(ompl PROPERTIES STATIC_LIBRARY_FLAGS "psapi.lib ws2_32.lib")
    elseif (MINGW)
        target_link_libraries(ompl psapi ws2_32)
    endif()
endif()

```

CMakeLists.txt
```diff
@@ -153,6 +153,12 @@ endif()
...
 find_program(R_EXEC R)
 
+# default to building shared libs except Windows
+if (MSVC)
+    option(OMPL_BUILD_SHARED "Build OMPL as a shared library" OFF)
+else()
+    option(OMPL_BUILD_SHARED "Build OMPL as a shared library" ON)
+endif()
 add_subdirectory(src)
 add_subdirectory(py-bindings)
```
saahu27, line 163: The omplConfig.cmake.in file may need adjustments to properly handle both shared and static library installations?

saahu27 (commented):
I was never able to build python bindings on windows.  ( I still cant). are we addressing that in this pull?

I also added some comments about whether change needs to be propagated to different files.

src/ompl/CMakeLists.txt
mamoll: Good catch. I just replaced it with a `if(OMPL_BUILD_SHARED)` check. Python bindings never worked on Windows and there are no plans to make the current bindings work on Windows.

mamoll: Yes. Fixed.

CMakeLists.txt
mamoll: I don't think any changes are required. I did an installation of a static build and the generated cmake config files look fine to me.

saahu27 (approved):

PR #1336, opened 2025-11-25
Add Greedy RRT* (G-RRT*) - bidirectional anytime planner

mlsdpk:
I would like to contribute an implementation of our planner, Greedy RRT* (G-RRT*), to OMPL, based on the following preprint: [https://arxiv.org/abs/2405.03411](https://arxiv.org/abs/2405.03411)

This paper has been out for a while, and it took me some time to clean things up and prepare a proper PR, but I’m happy to finally open it now.

In short, G-RRT* is a bidirectional planner, similar to RRTConnect, but with additional RRT*-style rewiring and greedy heuristic properties that make it anytime and allow it to converge faster than using informed sampling alone on most planning problems. I have also tested it on high-dimensional planning problems and MotionBenchMaker tasks; further details and experimental results are available in the manuscript.

Screenshot 2025-11-25 at 12 02 31 AM

To my knowledge, OMPL currently does not really have a bidirectional, RRT*-like anytime planner (for example, RRT*-Connect) so I believe this would be a useful addition for the community to experiment with and build upon. G-RRT* works very similar to RRT*-Connect but it has better heuristics and some additional properties which will make it find solutions very fast and converge faster.

### DIFF
--- a/demos/OptimalPlanning.cpp
+++ b/demos/OptimalPlanning.cpp
@@ -53,6 +53,7 @@
 #include <ompl/geometric/planners/fmt/FMT.h>
 #include <ompl/geometric/planners/fmt/BFMT.h>
 #include <ompl/geometric/planners/prm/PRMstar.h>
+#include <ompl/geometric/planners/rrt/GreedyRRTstar.h>
 #include <ompl/geometric/planners/rrt/InformedRRTstar.h>
 #include <ompl/geometric/planners/rrt/RRTstar.h>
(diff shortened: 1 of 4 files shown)


mlsdpk:
@zkingston Hi Zac, not sure if you're the right person here to review this PR, but since grrtstar is already in vamp, it would be great to merge the OMPL version too. The planner behaves exactly the same as vamp's implementation, and a new revision of the paper will be out soon as well.

cc @mamoll

PR #1339, opened 2025-12-02
Comment out BLITstar::costCombine declaration

yi213-robotic:
This PR comments out the unused "BLITstar::costCombine" declaration in "BLITstar.h". 
In some build configurations, the Python bindings fail to import because the symbol does not exist.


mamoll:
Can you just remove the commented-out code? I'll merge when you make that change.

Issue #1341, opened 2025-12-12
Desired changes for the future of OMPL


WeihangGuo:
- [ ] Rewrite Python bindings using `nanobind` (#1342)
   - [ ] Support Benchmark class in Python bindings?
- [ ] Clean up CMake and remove old dependencies
- [ ] Benchmark against MotionBenchMaker dataset
- [ ] Clean up old VAMP demo
- [ ] Add VAMP demonstrations (base motion planning and CAPT)
- [ ] More documentation and tutorials for using VAMP
- [ ] Remove old tutorials if unneccessary
- [ ] Remove OMPLapp
- [ ] Remove old Python bindings

mamoll:
- [ ] Update Planner Arena server
   - [ ] Run new benchmarks with OMPL 2.0.0 release candidate and generate sample benchmark.db file for the Planner Arena web site
   - [ ] Update old OMPL.app visualizations of benchmark in Planner Arena help page with images/movies of new benchmarks
   - [ ] Update Planner Arena version after changes are completed
   - [x] Deploy latest version via docker on plannerarena.kavrakilab.org @Thodoris1999

PR #1342, opened 2025-12-12
Add nanobind-based Python bindings

WeihangGuo:
Advances #1341.


result_demo.ppm
Thodoris1999, line 1: I assume this was added by mistake?

nanobinds/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+endif()
+
+target_link_libraries(_ompl PRIVATE ompl::ompl)
+target_compile_options(_ompl PRIVATE -Wno-unused-parameter)
+
+# Set RPATH so the module can find libompl.so at runtime during development
+set_target_properties(_ompl PROPERTIES
+    BUILD_RPATH "$<TARGET_FILE_DIR:ompl>"
+    INSTALL_RPATH "$ORIGIN;$ORIGIN/../lib"
+)
```
Thodoris1999, line 24: todo: reexamine if this is needed, ideally we don't interfere with user's RPATHs. Same with the library output dir

CMakeModules/OMPLDependencies.cmake
```diff
@@ -0,0 +1,100 @@
...
+
+# Nanobind
+set_package_properties(Nanobind PROPERTIES
+    URL "https://github.com/wjakob/nanobind"
+    PURPOSE "Used for building Python bindings.")
+
+if(OMPL_HAVE_PYTHON AND EXISTS "${CMAKE_CURRENT_SOURCE_DIR}/external/nanobind/CMakeLists.txt")
+    add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/external/nanobind)
+    set(OMPL_HAVE_NANOBIND 1)
+    find_package_handle_standard_args(Nanobind
```
Thodoris1999: what does find_package_handle_standard_args do here?

CMakeModules/OMPLDependencies.cmake
```diff
@@ -0,0 +1,100 @@
...
+        REQUIRED_VARS OMPL_HAVE_NANOBIND
+        VERSION_VAR nanobind_VERSION
+    )
+else()
+    set(OMPL_HAVE_NANOBIND 0)
+endif()
```
Thodoris1999, line 96: I would personally prefer that this logic that finds nanobind gets moved to inside `if(OMPL_BUILD_PYTHON_BINDINGS)` so we don't check the path at the if statement and cmake fails early with its internal message.
In general I have a preference that dependencies are found either inline at the top level `CMakeLists.txt` or in their own `.cmake` file (not all in XXXDependencies.cmake). It makes it explicitly obvious that they are optional, and cmake can not bother finding them if they are not turned on. We also not have to maintain find methods that certainly do not error out

nanobinds/base/spaces/RealVectorStateSpace.cpp
```diff
@@ -47,7 +47,13 @@ void ompl::binding::base::initSpaces_RealVectorStateSpace(nb::module_ &m)
                      }
                  }
                  return result;
-             }, nb::rv_policy::reference_internal)
```
Thodoris1999: You can instead throw a runtime error when the end of the slice is not given and support slices otherwise. Something along the lines of (cr: chatGPT)
```cpp
.def("__getitem__",
     [](const ob::RealVectorStateSpace::StateType *s,
        const nb::slice &slice) {

         // Reject open-ended slices: arr[:], arr[a:], arr[::k]
         if (slice.stop().is_none()) {
             throw nb::type_error(
                 "Open-ended slices are not supported (unknown state dim)");
         }

         // Extract stop (required)
         ssize_t stop = nb::cast<ssize_t>(slice.stop());

         // Extract start (default 0)
         ssize_t start = 0;
         if (!slice.start().is_none())
             start = nb::cast<ssize_t>(slice.start());

         // Extract step (default 1)
         ssize_t step = 1;
         if (!slice.step().is_none())
             step = nb::cast<ssize_t>(slice.step());

         // Reject negative or zero step
         if (step <= 0) {
             throw nb::type_error("Slice step must be a positive integer");
         }

         // Reject negative indices (need length to resolve them)
         if (start < 0 || stop < 0) {
             throw nb::type_error(
                 "Negative slice indices require a known dim count");
         }

         // Empty slice
         if (stop <= start) {
             return std::vector<double>{};
         }

         // Compute slice length safely
         size_t slicelength =
             static_cast<size_t>((stop - start + step - 1) / step);

         std::vector<double> result(slicelength);

         if (step == 1) {
             // Contiguous fast path
             std::memcpy(result.data(),
                         &s->values[start],
                         slicelength * sizeof(double));
         } else {
             // Strided copy
             for (size_t i = 0; i < slicelength; ++i) {
                 result[i] = s->values[start + i * step];
             }
         }

         return result;
     })
```

nanobinds/base/spaces/RealVectorStateSpace.cpp
```diff
@@ -62,7 +68,12 @@ void ompl::binding::base::initSpaces_RealVectorStateSpace(nb::module_ &m)
                          s->values[start + i * step] = values[i];
                      }
                  }
-             }, nb::rv_policy::reference_internal);
+             }, nb::rv_policy::reference_internal,
```
Thodoris1999, line 149: same here

nanobinds/CMakeLists.txt
Thodoris1999: Update: this is okay for dynamic library dependencies

WeihangGuo:
https://github.com/ompl/ompl/pull/1342#discussion_r2632994801

nanobinds/base/ProblemDefinition.cpp
```diff
@@ -0,0 +1,151 @@
...
+        // isTrivial: Bind a version that takes pointer parameters.
+        .def("isTrivial", nb::overload_cast<unsigned int*, double*>
+             (&ompl::base::ProblemDefinition::isTrivial, nb::const_))
+        // Also provide a simpler overload that calls isTrivial with nullptr for outputs.
+        .def("isTrivial", [](const ompl::base::ProblemDefinition &pd) {
+            return pd.isTrivial(nullptr, nullptr);
+        })
+
+        // isStraightLinePathValid returns a path pointer.
+        .def("isStraightLinePathValid", &ompl::base::ProblemDefinition::isStraightLinePathValid,
```
Thodoris1999: Since this is a shared_ptr, need to remove reference_internal and leave it default?

nanobinds/base/ProblemDefinition.cpp
```diff
@@ -0,0 +1,151 @@
...
+        // fixInvalidInputStates
+        .def("fixInvalidInputStates", &ompl::base::ProblemDefinition::fixInvalidInputStates)
+
+        // Solution properties.
+        .def("hasSolution", &ompl::base::ProblemDefinition::hasSolution)
+        .def("hasExactSolution", &ompl::base::ProblemDefinition::hasExactSolution)
+        .def("hasApproximateSolution", &ompl::base::ProblemDefinition::hasApproximateSolution)
+        .def("getSolutionDifference", &ompl::base::ProblemDefinition::getSolutionDifference)
+        .def("hasOptimizedSolution", &ompl::base::ProblemDefinition::hasOptimizedSolution)
+        .def("getSolutionPath", &ompl::base::ProblemDefinition::getSolutionPath,
```
Thodoris1999: shared_ptr

nanobinds/base/ProblemDefinition.cpp
```diff
@@ -0,0 +1,151 @@
...
+             nb::rv_policy::reference_internal)
+
+        // getSolution: because the original function fills a PlannerSolution passed by reference,
+        // we wrap it so that it returns the solution if one is found.
+        .def("getSolution", &ompl::base::ProblemDefinition::getSolution, 
+             nb::rv_policy::reference_internal)
```
Thodoris1999: perhaps rv_policy not needed for bool?

nanobinds/base/ProblemDefinition.cpp
```diff
@@ -0,0 +1,151 @@
...
+        .def("getSolutions", &ompl::base::ProblemDefinition::getSolutions)
+
+        // Clear solution paths.
+        .def("clearSolutionPaths", &ompl::base::ProblemDefinition::clearSolutionPaths)
+
+        // Proof of non-existence functions.
+        .def("hasSolutionNonExistenceProof", &ompl::base::ProblemDefinition::hasSolutionNonExistenceProof)
+        .def("clearSolutionNonExistenceProof", &ompl::base::ProblemDefinition::clearSolutionNonExistenceProof)
+        .def("getSolutionNonExistenceProof", &ompl::base::ProblemDefinition::getSolutionNonExistenceProof,
+             nb::rv_policy::reference_internal)
```
Thodoris1999: shared_ptr

nanobinds/base/StateSpace.cpp
```diff
@@ -0,0 +1,128 @@
...
+        .def("cloneState", &ob::StateSpace::cloneState, nb::arg("source"), nb::rv_policy::take_ownership)
+        .def("getSerializationLength", &ob::StateSpace::getSerializationLength)
+        .def("getValueAddressAtIndex",
+             nb::overload_cast<ob::State *, unsigned int>(&ob::StateSpace::getValueAddressAtIndex, nb::const_),
+             nb::arg("state"), nb::arg("index"), nb::rv_policy::reference_internal)
+
+        .def("registerProjection", &ob::StateSpace::registerProjection, nb::arg("name"), nb::arg("projection"))
+        .def("registerDefaultProjection", &ob::StateSpace::registerDefaultProjection, nb::arg("projection"))
+        .def("registerProjections", &ob::StateSpace::registerProjections)
+        .def("getProjection", &ob::StateSpace::getProjection, nb::arg("name"), nb::rv_policy::reference_internal)
```
Thodoris1999: shared_ptr

nanobinds/base/StateSpace.cpp
```diff
@@ -0,0 +1,128 @@
...
+        .def("getDefaultProjection", &ob::StateSpace::getDefaultProjection, nb::rv_policy::reference_internal)
```
Thodoris1999: shared_ptr

nanobinds/base/StateSpace.cpp
```diff
@@ -0,0 +1,128 @@
...
+            nb::arg("state"))
+        .def("printSettings", [](const ob::StateSpace &ss) { ss.printSettings(std::cout); })
+        .def("printProjections", [](const ob::StateSpace &ss) { ss.printProjections(std::cout); })
+        .def("sanityChecks", nb::overload_cast<double, double, unsigned int>(&ob::StateSpace::sanityChecks, nb::const_),
+             nb::arg("zero"), nb::arg("eps"), nb::arg("flags"))
+        .def("Diagram", &ob::StateSpace::Diagram)
+        .def("List", &ob::StateSpace::List)
+        .def("allocSubspaceStateSampler",
+             nb::overload_cast<const ob::StateSpacePtr &>(&ob::StateSpace::allocSubspaceStateSampler, nb::const_),
+             nb::arg("subspace"), nb::rv_policy::reference_internal)
```
Thodoris1999: shared_ptr

nanobinds/control/SimpleSetup.cpp
```diff
@@ -0,0 +1,152 @@
...
+        .def("getPlannerAllocator", &oc::SimpleSetup::getPlannerAllocator)
+
+        // solution path checks
+        .def("haveExactSolutionPath", &oc::SimpleSetup::haveExactSolutionPath)
+        .def("haveSolutionPath", &oc::SimpleSetup::haveSolutionPath)
+
+        // getSolutionPath: returns a PathControl& => use reference_internal
+        .def(
+            "getSolutionPath", [](oc::SimpleSetup &self) -> oc::PathControl & { return self.getSolutionPath(); },
+            nb::rv_policy::reference_internal)
```
Thodoris1999, line 169: shared_ptr

nanobinds/control/SimpleSetup.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+        // getPlannerData
+        .def(
+            "getPlannerData", [](const oc::SimpleSetup &ss, ob::PlannerData &pd) { ss.getPlannerData(pd); },
+            nb::arg("plannerData"), nb::rv_policy::reference_internal)
```
Thodoris1999: reference_internal, but function is void?

CMakeModules/OMPLDependencies.cmake
```diff
@@ -0,0 +1,96 @@
...
+    set(OMPL_HAVE_YAMLCPP 0)
+endif()
+
+# Doxygen
+set_package_properties(Doxygen PROPERTIES
+    URL "http://doxygen.org"
+    PURPOSE "Used to create the OMPL documentation (i.e., https://ompl.kavrakilab.org).")
+find_package(Doxygen)
+
+# R
```
mamoll, line 66: We can remove R as dependency if we remove the R version of planner arena and switch to the (separate) Python plannerarena package. Let's discuss.

demos/ConstrainedPlanningCommon.py
mamoll, line 1: I'd prefer it if corresponding C++ and Python demos lived in the same directory so that it's easy to see that you can do mostly the same things in Python as in C++. In other words, I'd suggest moving this file and all other files out of the `nanobinds` dir

demos/nanobinds/InteractiveArmPlanning.py
```diff
@@ -0,0 +1,235 @@
...
+"""
+
+import time
+import os
+import numpy as np
+
+from ompl import base as ob
+from ompl import geometric as og
+
+import pinocchio
```
mamoll: Make sure to document in doc/markdown/ somewhere which optional requirements are needed for what (and how they can be installed). Same comment applies to ViserGUI below.

demos/nanobinds/InteractiveArmPlanning.py
```diff
@@ -0,0 +1,235 @@
...
+                sphere.parentJoint = 0
+                geom_id = self.collision_model.addGeometryObject(sphere)
+                
+                # Add collision pairs with all robot geometries
+                for i in range(num_robot_geoms):
+                    self.collision_model.addCollisionPair(pinocchio.CollisionPair(i, geom_id))
+        
+        self.collision_data = self.collision_model.createData()
+        
+    def is_state_valid(self, q: np.ndarray) -> bool:
```
mamoll: I really like the type hints. If it's not too much trouble, try to use that in any new code.

demos/nanobinds/PlannerData.py
```diff
@@ -0,0 +1,347 @@
...
+    print('Failed to import viser.  PlannerData will not be visualized')
+    viser_available = False
+
+try:
+    from ompl import base as ob
+    from ompl import geometric as og
+except ImportError:
+    # if the ompl module is not in the PYTHONPATH assume it is installed in a
+    # subdirectory of the parent directory called "py-bindings."
+    from os.path import abspath, dirname, join
```
mamoll: You can delete this chunk of error handling code from all python demos

demos/nanobinds/ViserGUI.py
mamoll, line 1: A high-level beginner tutorial with screenshots of this GUI would go a long way to fill the gap left by removing OMPL.app

demos/resources/README.md
```diff
@@ -0,0 +1,2 @@
+# Robots
+The UR5 models are sourced from [VAMP](https://github.com/KavrakiLab/vamp/tree/main/resources).
```
mamoll: why copy the file since you already have the vamp repo as a git submodule? would a symlink suffice?

nanobinds/base/goals/GoalRegion.cpp
```diff
@@ -0,0 +1,44 @@
...
+        // both overloads of isSatisfied
+        .def("isSatisfied", nb::overload_cast<const ob::State *>(&ob::GoalRegion::isSatisfied, nb::const_),
+             nb::arg("state"), "Check if a state satisfies the goal region")
+        .def("isSatisfied", nb::overload_cast<const ob::State *, double *>(&ob::GoalRegion::isSatisfied, nb::const_),
+             nb::arg("state"), nb::arg("distance"), "Check if a state satisfies the goal region, and return distance")
+        .def("setThreshold", &ob::GoalRegion::setThreshold, nb::arg("threshold"), "Set the goal‐region threshold")
+        .def("getThreshold", &ob::GoalRegion::getThreshold, "Get the goal‐region threshold")
+        // friendly print() returning a string
+        .def(
+            "print",
```
mamoll: the more pythonic way would be to call this method `__str__`. Thay way you can you just use `print(goalRegion)` (instead of `print(goalRefion.print())`)

nanobinds/control/planners/est/EST.cpp
```diff
@@ -0,0 +1,18 @@
...
+
+namespace nb = nanobind;
+namespace oc = ompl::control;
+namespace ob = ompl::base;
+
+void ompl::binding::control::initPlannersEst_EST(nb::module_& m)
+{
+    // TODO [oc::EST][TEST]
+    // TODO [oc::EST][MISSING]
+    // TAG [oc::EST][Planner]
```
mamoll: What does `TAG` mean? Also, are all the `TODO` comments blockers for release?

nanobinds/control/spaces/DiscreteControlSpace.cpp
mamoll, line 1: This looks like it wouldn't be very functional. I'd prefer if we left code that didn't do anything useful out of this PR

README.md
```diff
@@ -39,6 +39,11 @@ following commands:
...
     cmake ../..
-    # next step is optional
-    make -j 4 update_bindings # if you want Python bindings
     make -j 4 # replace "4" with the number of cores on your machine
+
+To install the Python binding. Go to the top-level directory of OMPL and type the following commands:
+```
+git submodule update --init --recursive  # for VAMP integration
+cd nanobinds
+pip install .
```
mamoll, line 48: I think these instructions are out of date and the python bindings are compiled as part of the normal build process, right? If not, they should be.

Thodoris1999: We are currently using the packaging procedure suggested in nanobind's documentation, which involves pyproject.toml and scikit-build [https://nanobind.readthedocs.io/en/latest/packaging.html](https://nanobind.readthedocs.io/en/latest/packaging.html). This makes the build system be designed to be called through scikit-build-core, hence the `pip install .`. This can also incorporate cibuildwheel, but we haven't figured out the distribution details yet (after this gets merged?).

The normal build process does build the nanobind module as well though (afaik the difference is that you get just the .so module and not the full wheel the scikit-build creates). By the way I wanted to bring up again if the python bindings should be opt out instead of opt in, since myself for example like to integrate with ompl as a subdirectory, so it may be surprising that the build takes the extra time to build the bindings by default.

mamoll: Ok, we will resolve packaging in a later PR.

demos/nanobinds/ViserGUI.py
mamoll: captured in issue #1341

CMakeModules/OMPLDependencies.cmake
mamoll: Will be done in a separate PR. I'll take care of this.

demos/ConstrainedPlanningCommon.py
WeihangGuo: Shall we move the files to e.g., `src/ompl/base` or `src/ompl/base/src`?  Also shall we rename the file to `OriginalNamePy.cpp`?

mamoll: The Python binding code should be in a separate directory, but the _Python demos_ should be in the `demos` directory alongside their C++ equivalents.

WeihangGuo: Oh, I misread. Yes, I will do that.

demos/nanobinds/InteractiveArmPlanning.py
WeihangGuo: I have temporarily deleted these files along with the viser utils. Will add a cleaner version later.

PR #1343, opened 2025-12-12
fix: remove macos-13 from github actions workflows

claytonwramsey:
[MacOS 13 is deprecated on GitHub Actions](https://github.blog/changelog/2025-09-19-github-actions-macos-13-runner-image-is-closing-down/) and so this will keep CI from breaking.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
           pybuilds: cp3{10,11,12,13}-manylinux_aarch64
           arch: aarch64
           id: linux_arm64
-        - os: macos-13
+        - os: macos-15-intel
           pybuilds: cp3{10,11,12,13}-macosx_x86_64
           arch: x86_64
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -21,10 +21,6 @@ jobs:
           pybuilds: cp3{10,11,12,13}-manylinux_aarch64
           arch: aarch64
           id: linux_arm64
-        - os: macos-13
```
mamoll, line 24: I'd just change this line to:
```
     -os: macos-15-intel 
```

.github/workflows/wheels.yaml
```diff
@@ -21,10 +21,6 @@ jobs:
           pybuilds: cp3{10,11,12,13}-manylinux_aarch64
           arch: aarch64
           id: linux_arm64
-        - os: macos-13
```
mamoll, line 24: same here

PR #1344, opened 2025-12-12
Clean up the old bindings, demo the vamps, and update the CMake files

WeihangGuo:
Advances #1341 
- [x] Fully remove the old bindings
- [x] Fully remove the old vamp demos
- [x] Clean up the CMakeLists.txt
  - [x] Move all dependencies into `OMPLDependencies.cmake`
  - [x] Clean up legacy setups
  - [x] Rearrange lines for readability


doc/markdown/buildOptions.md
```diff
@@ -12,7 +12,6 @@ If you are building OMPL from source, there are several options that you can use
 | OMPL_BUILD_VAMP               | ON            | Build VAMP (Vector-Accelerated Motion Planning) submodule for enhanced collision checking performance. |
 | OMPL_REGISTRATION             | ON            | Whether the registration page is shown. (Disabling it might be useful for build bots.) |
 | OMPL_VERSIONED_INSTALL        | ON            | Install header files in include/ompl-X.Y/ompl, where X and Y are the major and minor version numbers. |
-| VAMP_PORTABLE_BUILD           | OFF           | Build VAMP with portable SIMD settings for distribution (package maintainers). When OFF, uses native SIMD for best performance. |
```
mamoll, line 15: This seems like a feature we'd want to preserve. For *.deb and *.rpm releases maintained by others, we can't always default to `-mtune=native` or something like that.

py-bindings/pyproject.toml
mamoll, line 1: We'll want to have a similar file later on for the new Python bindings, but it can probably be much simpler. Once the nanobind-based bindings are merged, we can worry about setting up the GitHub CI build of binary ompl python modules and automatically uploading them to PyPI.

scripts/CMakeLists.txt
```diff
@@ -1,8 +1,3 @@
-install_python(PROGRAMS ompl_benchmark_statistics.py
-    DESTINATION ${CMAKE_INSTALL_BINDIR}
-    COMPONENT ompl
-    RENAME ompl_benchmark_statistics.py)
-
```
mamoll, line 5: Why do you no longer want to install this? This is an essential component of OMPL benchmarking.

tests/util/test_vamp_integration.cpp
```diff
@@ -28,44 +24,6 @@ BOOST_AUTO_TEST_CASE(VampTargetLinkingTest)
     BOOST_CHECK_MESSAGE(true, "vamp_cpp target successfully linked");
 }
 
-// Test OMPL to VAMP state conversion for Panda robot
```
mamoll, line 31: are you planning to have some basic test code for your own ompl-vamp integration?

CMakeLists.txt
```diff
@@ -2,6 +2,23 @@ cmake_minimum_required(VERSION 3.12)
 project(ompl VERSION 2.0.0 LANGUAGES CXX)
 set(PROJECT_VERSION 2.0.0-beta)
 set(OMPL_ABI_VERSION 19)
+include(FeatureSummary)
+
+option(OMPL_REGISTRATION "Enable one-time registration of OMPL" ON)
```
mamoll, line 7: The time may have come to get rid of this. Check with Lydia if the registration data is actually used at all.

mamoll (changes requested):
I'm mostly fine with these changes. There are some things that will have to be solved later (like PyPI packaging). I think there's also one accidental change (not installing ompl_benchmark_statistics.py) that needs to be undone.

doc/markdown/buildOptions.md
WeihangGuo: I have restored the `VAMP_PORTABLE_BUILD` flag.

WeihangGuo:
I have restored the VAMP_PORTABLE_BUILD flag.

CMakeLists.txt
WeihangGuo: I have added this to the todo list. #1341

PR #1282, opened 2025-05-30
Add Asymptotically Optimal RRT-Connect (AORRTC)

twill777:
This PR adds the implementation of the Asymptotically Optimal RRT-Connect (AORRTC) planner from https://arxiv.org/abs/2505.10542.
This is an asymptotically optimal motion planner that extends the AO-x meta algorithm and RRT-Connect to provide state-of-the-art performance on the MotionBenchMaker dataset.

PR #1345, opened 2025-12-17
Handle compound state space for AORRTC

Jaeyoung-Lim:
**Problem Description**
When using compound statespace(e.g. Dubins) with AORRTC https://github.com/ompl/ompl/pull/1282 the `euclideanDistanceFunction` was using the distance metric of the state space rather than the euclidean distance function. (Probably assuming all spaces are metric and euclidean?)

This PR checks whether the statespace is a compound statespace, and tries to extract the realvector statespace to compute the euclideandistance. 

FYI @twill777

### DIFF
--- a/src/ompl/geometric/planners/rrt/AOXRRTConnect.h
+++ b/src/ompl/geometric/planners/rrt/AOXRRTConnect.h
@@ -39,6 +39,7 @@
 
 #include "ompl/datastructures/NearestNeighbors.h"
 #include "ompl/base/OptimizationObjective.h"
+#include "ompl/base/spaces/RealVectorStateSpace.h"
 #include "ompl/geometric/planners/PlannerIncludes.h"
 #include <ompl/geometric/planners/rrt/RRTConnect.h>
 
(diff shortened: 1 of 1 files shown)


src/ompl/geometric/planners/rrt/AOXRRTConnect.h
```diff
@@ -192,9 +193,39 @@ namespace ompl
...
+             * distance() which may return path-based non-metric distances.
+             */
             double euclideanDistanceFunction(const Motion *a, const Motion *b) const
             {
+                // Try to get actual Euclidean distance for compound state spaces
+                auto *stateSpace = si_->getStateSpace().get();
+                if (stateSpace->isCompound())
+                {
+                    auto *compoundSpace = stateSpace->as<base::CompoundStateSpace>();
+                    // Get the first subspace which should be the position component (RealVectorStateSpace)
```
mamoll, line 209: This is not guaranteed. Rephrase as:
```suggestion
                    // Get the first subspace which should be the position component (RealVectorStateSpace) for state spaces like SE(2) and SE(3).
```

src/ompl/geometric/planners/rrt/AOXRRTConnect.h
```diff
@@ -192,9 +193,39 @@ namespace ompl
...
+                    auto *subspace = compoundSpace->getSubspace(0).get();
+                    if (subspace->getType() == base::STATE_SPACE_REAL_VECTOR)
+                    {
+                        auto *rvSpace = subspace->as<base::RealVectorStateSpace>();
```
mamoll, line 213: No need to reimplement RealVectorStateSpace::distance. I think you can just call `rvspace->distance(aCompound->as<base::RealVectorStateSpace::StateType>(0), bCompound->as<base::RealVectorStateSpace::StateType>(0))`

PR #1350, opened 2026-01-16
Benchmark in Python

WeihangGuo:


nanobinds/CMakeLists.txt
```diff
@@ -64,5 +64,11 @@ install(FILES
         ${CMAKE_CURRENT_SOURCE_DIR}/ompl/geometric.py
         ${CMAKE_CURRENT_SOURCE_DIR}/ompl/control.py
         ${CMAKE_CURRENT_SOURCE_DIR}/ompl/util.py
+        ${CMAKE_CURRENT_SOURCE_DIR}/ompl/tools.py
         DESTINATION ompl)
+
+# Install benchmark statistics as a private module
+install(FILES ${CMAKE_SOURCE_DIR}/scripts/ompl_benchmark_statistics.py
```
mamoll: I don't think this will work. People that use the C++ Benchmark class also need this script to be available as a command line tool that can turn logs into sqlite3 databases.

In a separate PR I could perhaps split this script into a part that becomes part of the ompl python module and a part that does just the command line parsing. We should discuss whether we should make the python bindings enabled by default. It'd just add python as a build dependency, right? (since nanobind is already include in ompl itself).

mamoll (changes requested):
One small change request. Looks good otherwise

nanobinds/CMakeLists.txt
WeihangGuo: Shall I keep this line for now or remove it? 
If the user clones the submodules, enabling the Python bindings only requires a Python dependency. However, it will increase compile time.

mamoll: Don't install it in the ompl dir, but install (without renaming) in the `bin/` directory

WeihangGuo: We install the `ompl_benchmark_statistics` in `bin` in the `scripts/CMakeLists.txt`.
```
install(PROGRAMS ompl_benchmark_statistics.py
    DESTINATION ${CMAKE_INSTALL_BINDIR}
    COMPONENT ompl
    RENAME ompl_benchmark_statistics.py)
```

nanobinds/ompl/tools.py
```diff
@@ -0,0 +1,17 @@
+from ._ompl.tools import *
+
+# Import benchmark statistics functions
+from ._benchmark_statistics import (
```
mamoll, line 4: this is not going to work unless we have split up the benchmark_statisics script into two parts: one part that is part of this module and contains almost all the functionality and a minimal command line tool that does the command line parsing and calls the functions in the ompl module

mamoll (changes requested):

nanobinds/ompl/tools.py
WeihangGuo: My bad, I forgot this. I will split it into two.

mamoll (approved):

PR #1355, opened 2026-02-06
remove registration

mamoll:
remove registration step in first-time build from source


mamoll:
also added some basic fixes for compiler warnings

PR #1361, opened 2026-02-20
Fix GitHub actions

Thodoris1999:
This hopefully makes the github CI pass and also automatically builds and uploads the python wheels as artifacts

Noteworthy info:
- For macos, I wasn't able to get the wheels to build with dynamically linked ompl so I switch the default to static library. Let me know if that's undesirable or you want to only fall back when OMPL_BUILD_PYTHON_BINDINGS is on.
- I tried to get windows CI to work, but [it doesn't](https://github.com/Thodoris1999/ompl/actions/runs/22209580572/job/64240751033#step:3:1977). I kept some of the infrastructure in case anyone wants to fix it in the future, but the workflow is disabled.
- On static builds some of the tests were crashing due to to `g_RNGSeedGenerator` and `g_allocatedSpaces` being invalid. I added some changes that fix this
- One [test](https://github.com/Thodoris1999/ompl/actions/runs/22243487593/job/64352340577) was failing on MacOS, I am not sure if this was a spontaneous fail or something to be addressed

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,7 +21,7 @@ add_feature_info(OMPL_BUILD_PYTHON_BINDINGS "${OMPL_BUILD_PYTHON_BINDINGS}" "Whe
 if (MSVC)
     option(OMPL_BUILD_SHARED "Build OMPL as a shared library" OFF)
 else()
-    option(OMPL_BUILD_SHARED "Build OMPL as a shared library" OFF)
+    option(OMPL_BUILD_SHARED "Build OMPL as a shared library" ON)
 endif()
 add_feature_info(OMPL_BUILD_SHARED "${OMPL_BUILD_SHARED}" "Whether to build OMPL as a shared library (default is ON for non-Windows, OFF for Windows).")
 
--- a/nanobinds/CMakeLists.txt
+++ b/nanobinds/CMakeLists.txt
@@ -18,7 +18,7 @@ target_link_libraries(_ompl PRIVATE ompl::ompl)
 target_compile_options(_ompl PRIVATE -Wno-unused-parameter)
 
 # Set RPATH so the module can find libompl.so
-if(UNIX)
+if(UNIX AND NOT APPLE)
     set_target_properties(_ompl PROPERTIES
         BUILD_RPATH "$ORIGIN"
         INSTALL_RPATH "$ORIGIN"


src/ompl/base/src/StateSpace.cpp
```diff
@@ -63,18 +62,10 @@ namespace ompl
                 unsigned int counter_{0};
             };
 
-            static boost::scoped_ptr<AllocatedSpaces> g_allocatedSpaces;
```
mamoll, line 66: Why was this change needed? This seems unrelated to fixing Github actions. Please undo this change

src/ompl/util/src/RandomNumbers.cpp
```diff
@@ -113,18 +113,10 @@ namespace
         std::uniform_int_distribution<> sDist_;
     };
 
-    std::once_flag g_once;
```
mamoll, line 116: same comment here as in StateSpace.cpp

CMakeLists.txt
```diff
@@ -17,12 +17,7 @@ add_feature_info(VAMP_PORTABLE_BUILD "${VAMP_PORTABLE_BUILD}" "Whether to build
 option(OMPL_BUILD_PYTHON_BINDINGS "Build OMPL Python bindings" OFF)
 add_feature_info(OMPL_BUILD_PYTHON_BINDINGS "${OMPL_BUILD_PYTHON_BINDINGS}" "Whether to build OMPL Python bindings.")
 
-# default to building shared libs except Windows
-if (MSVC)
-    option(OMPL_BUILD_SHARED "Build OMPL as a shared library" OFF)
-else()
-    option(OMPL_BUILD_SHARED "Build OMPL as a shared library" ON)
-endif()
+option(OMPL_BUILD_SHARED "Build OMPL as a shared library" OFF)
```
mamoll: Why default to a static library? This is atypical.

Thodoris1999: This is due to the first bullet point and that we build the bindings by default. If we manage to get the macos bindings to work with dynamic libs this is not needed.
The issue was that the RPATH trick is linux only and something different is needed for MACos, so the ompl library is not found at runtime. I am not very familiar with MACos and don't have a mac for testing, but this can be reproduced from the pipeline or locally by turning BUILD_SHARED on.

PR #1367, opened 2026-03-13
Clean up and deprecate legacy Python bindings

WeihangGuo:


src/ompl/base/goals/GoalLazySamples.h
```diff
@@ -59,13 +59,6 @@ namespace ompl
          sampling to happen in a separate thread, and the number of
          goals may increase, as the planner is running, in a
          thread-safe manner.
-
-         \todo The Python bindings for GoalLazySamples class are still broken.
-         The OMPL C++ code creates a new thread from which you should be able
-         to call a python Goal sampling function. Acquiring the right threads
-         and locks and messing around with the Python Global Interpreter Lock
-         (GIL) is very tricky. See ompl/py-bindings/generate_bindings.py for
```
Thodoris1999, line 67: How was this addressed in the new bindings?

mamoll: good question. Same question applies to PRM, which is also multithreaded. In the old bindings we had a special single-threaded implementation of PRM that was used for the Python bindings.

tests/CMakeLists.txt
```diff
@@ -84,29 +84,5 @@ if (OMPL_BUILD_TESTS)
             message(WARNING "OMPL_HAVE_VAMP is TRUE but vamp_cpp target not found")
         endif()
     endif()
-
```
mamoll, line 87: can we move the pytests dir into the tests dir?

doc/markdown/deprecated/pybindingsPlanner.md
mamoll, line 1: rename the dir `deprecate` to be `deprecated` (to be consistent with the other places where you have created `deprecated` dirs)

tests/pytests/test_geo_planners.py
```diff
@@ -71,7 +71,7 @@ def test_prm_star():
     assert solution_path is not None
 
 if __name__ == "__main__":
-    test_rrt()
-    test_rrt_connect()
+    # test_rrt()
```
mamoll: why are these tests now commented out?

mamoll (changes requested):
I'd like `ninja test` to run the python tests (as well as the c++ tests). Can you look at the `add_ompl_python_test` cmake macro we had for the old bindings and update it for the new tests?

Thodoris1999 (changes requested):
The wheel build pipeline is failing because the pytest path referenced there needs to be updated. Besides that, the commit history needs to be cleaned up

doc/markdown/deprecated/pybindingsPlanner.md
mamoll: you have to set DOXYGEN_EXCLUDE_PATTERNS in doc/CMakeLists.txt before the call to `doxygen_add_docs` to make sure they are not part of the web site when we regenerate it.

tests/CMakeLists.txt
mamoll: I also meant: can we make it so that running `ctest` or `make test` also runs the python tests? Maybe this old macro from OMPL 1.7 is helpful?
```cmake
macro(add_ompl_python_test test_file)
  get_filename_component(test_name "${test_file}" NAME)
  add_test(${test_name} "${PYTHON_EXEC}" "${CMAKE_CURRENT_SOURCE_DIR}/${test_file}" "-v")
endmacro(add_ompl_python_test)
```
It's nice to be able to run individual tests during development

mamoll: never mind, you already said you'd do this in a separate PR.

PR #1407, opened 2026-04-03
make motion_benchmaker create a massive parameterized benchmark db

mamoll:


WeihangGuo (approved):
Looks good to me. My only comment is maybe we can remind the user of the usage of `ompl_benchmark_statistics.py`, i.e., `ompl_benchmark_statistics.py -d mbm.db benchmark*/*.log` again somewhere in the script or in the README.

PR #1413, opened 2026-04-07
AORRTC: fix memory leak and findNeighbour padding for non-Euclidean costs

mlsdpk:
Hello! While testing AORRTC with non-Euclidean cost spaces, I ran into a couple of leaks and robustness issues:

- Memory leak in `AOXRRTConnect::reset()`: trees were cleared without freeing the Motion objects they hold by raw pointer. Now calls `freeMemory()` first. Also frees `cmotion->state` in `solve()` cleanup.
- `findNeighbour()` padding: absolute pad (`rootDist += 1e-5`) becomes negligible for large `rootDist` (e.g. SE2, SE3, non-Euclidean metrics), causing empty `nearestR` results. Replaced with relative pad `rootDist *= (1 + 1e-6)`.

Minor fixes bundled in:
- bounds check in `findNeighbour()` cost-iteration loop
- initialise `GrowState gs = TRAPPED`
- fix `tgi.xmotion` bookkeeping in `solve()`'s straight-line check
- inner connect loop now respects `ptc`

### DIFF
--- a/src/ompl/geometric/planners/rrt/AOXRRTConnect.h
+++ b/src/ompl/geometric/planners/rrt/AOXRRTConnect.h
@@ -223,11 +223,10 @@ namespace ompl
 
             std::size_t sampleAttempts{0};
 
-            /* Pad rootDist to account for floating point error
-               Needed to make sure the root is included in nearest list
-               TODO: Should use some relative epsilon for padding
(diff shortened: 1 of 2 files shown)


Thodoris1999:
Thank you for the contribution Phone! Changes make sense to me.
@twill777 Can you also take a look?

Issue #1351, opened 2026-01-20
Illegal instruction due to march=native


simonschmeisser:
I just ran into issues due to defaulting to `march=native` here:

https://github.com/ompl/ompl/blob/8f60adae80c8138880650df2ee527f7b11a94004/CMakeLists.txt#L167

Of course our build farm uses a newer CPU than the box used for manual release tests so we got a crash there (AVX512 is not available)

I vote for changing the default to conservative settings

PR #1422, opened 2026-05-08
Default VAMP_PORTABLE_BUILD to ON when building Debian packages

MarqRazz:
## Default `VAMP_PORTABLE_BUILD` to ON when building Debian packages                                                                                                                                                                                                          
                                                                                                                                                                                                                                                                              
  ### Why                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                
  When OMPL is consumed via apt (`ros-rolling-ompl`, `ros-jazzy-ompl`, distro debs, anyone redistributing a binary deb), VAMP's SIMD intrinsics need to be compiled for an ISA baseline that *every* user's CPU supports — not the build host's CPU. The current default (`OFF`)
   compiles VAMP for whatever ISA extensions the buildfarm/CI host happens to advertise, which today often includes AVX-512.                                                                                                                                                    
                                                                                                                                                                                                                                                                                
  The failure mode is silent and ugly: a deb built on an AVX-512 host installs fine, configures fine, even loads fine — and then `SIGILL`s deep in a motion-planning hot path the first time VAMP issues a vector instruction. Most consumer Intel CPUs still don't have AVX-512
   (it was disabled in Alder Lake / Raptor Lake), and AMD only added it in Zen 4 (2022), so the affected user population is large. The `option()` description in `CMakeLists.txt` already says *"for distribution (package maintainers)"* — i.e., this knob exists exactly for
  this case. It just isn't getting turned on automatically anywhere.                                                                                                                                                                                                            
                  
  ### What this PR does

  A single change to `CMakeLists.txt`: if `DEB_BUILD_OPTIONS` is set in the environment, default `VAMP_PORTABLE_BUILD` to `ON`. Otherwise leave it `OFF` as today.                                                                                                              
   
  ```cmake                                                                                                                                                                                                                                                                      
  if(DEFINED ENV{DEB_BUILD_OPTIONS})
      set(_VAMP_PORTABLE_DEFAULT ON)                                                                                                                                                                                                                                            
  else()
      set(_VAMP_PORTABLE_DEFAULT OFF)                                                                                                                                                                                                                                           
  endif()                                                                                                                                                                                                                                                                       
  option(VAMP_PORTABLE_BUILD "Build VAMP with portable SIMD settings for distribution (package maintainers)" ${_VAMP_PORTABLE_DEFAULT})
                                                                                                                                                                                                                                                                                
  DEB_BUILD_OPTIONS is exported by dpkg-buildpackage / debhelper for every Debian build — bloom-generated ROS buildfarm jobs, apt source rebuilds, downstream distro packaging — and is unset for direct cmake .. and colcon build invocations. Net effect:                     
                                                                                                                                                                                                                                                                                
  ┌─────────────────────────────────────────────────────────┬────────────────────────────────────────┐                                                                                                                                                                          
  │                       Invocation                        │      VAMP_PORTABLE_BUILD default       │
  ├─────────────────────────────────────────────────────────┼────────────────────────────────────────┤
  │ Developer source build (cmake .. / colcon build / IDE)  │ OFF — unchanged, native SIMD preserved │
  ├─────────────────────────────────────────────────────────┼────────────────────────────────────────┤
  │ ROS buildfarm binarydeb job                             │ ON — portable, safe to redistribute    │                                                                                                                                                                          
  ├─────────────────────────────────────────────────────────┼────────────────────────────────────────┤                                                                                                                                                                          
  │ Downstream distro maintainer rebuilding from source deb │ ON                                     │                                                                                                                                                                          
  ├─────────────────────────────────────────────────────────┼────────────────────────────────────────┤                                                                                                                                                                          
  │ Anyone passing -DVAMP_PORTABLE_BUILD=... explicitly     │ their choice (override always wins)    │
  └─────────────────────────────────────────────────────────┴────────────────────────────────────────┘

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ add_feature_info(OMPL_BUILD_VAMP "${OMPL_BUILD_VAMP}" "Whether to build VAMP sub
 # is exported by dpkg-buildpackage / debhelper, so this catches the ROS buildfarm
 # and any apt-builds-from-source automatically). Direct `cmake ..` developer
 # builds keep the native-ISA default.
-if(DEFINED ENV{DEB_BUILD_OPTIONS})
+if(DEFINED ENV{DEB_BUILD_OPTIONS} OR SKBUILD)
     set(_VAMP_PORTABLE_DEFAULT ON)
 else()
     set(_VAMP_PORTABLE_DEFAULT OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -8,7 +8,16 @@ option(VAMP_BUILD_PYTHON_BINDINGS "Build VAMP Python bindings (requires OMPL_BUI
 add_feature_info(VAMP_BUILD_PYTHON_BINDINGS "${VAMP_BUILD_PYTHON_BINDINGS}" "Whether to build VAMP Python bindings.")
 option(OMPL_BUILD_VAMP "Build VAMP submodule" ON)
 add_feature_info(OMPL_BUILD_VAMP "${OMPL_BUILD_VAMP}" "Whether to build VAMP submodule.")
-option(VAMP_PORTABLE_BUILD "Build VAMP with portable SIMD settings for distribution (package maintainers)" OFF)
+# Default to portable SIMD when invoked from a Debian/dpkg build (DEB_BUILD_OPTIONS
+# is exported by dpkg-buildpackage / debhelper, so this catches the ROS buildfarm
+# and any apt-builds-from-source automatically). Direct `cmake ..` developer
+# builds keep the native-ISA default.
+if(DEFINED ENV{DEB_BUILD_OPTIONS})
+    set(_VAMP_PORTABLE_DEFAULT ON)
+else()
+    set(_VAMP_PORTABLE_DEFAULT OFF)
+endif()
+option(VAMP_PORTABLE_BUILD "Build VAMP with portable SIMD settings for distribution (package maintainers)" ${_VAMP_PORTABLE_DEFAULT})
 add_feature_info(VAMP_PORTABLE_BUILD "${VAMP_PORTABLE_BUILD}" "Whether to build VAMP with portable SIMD settings for distribution (package maintainers).")
 
 option(OMPL_BUILD_PYTHON_BINDINGS "Build OMPL Python bindings" ON)


Thodoris1999:
Hi Marq, thank you for maintaining the ROS package.

VAMP support on OMPL is intended for high-performance planning and VAMP itself upstream defaults to compiling with `-march=native`. For local installation users, switching to portable builds by default can slow down motion validation in theory up to 4x compared to if those extensions are available. So we offer the `VAMP_PORTABLE_BUILD` flag to overwrite to portable builds.

Can you switch to `VAMP_PORTABLE_BUILD=ON` from your side on the build configuration?

Related: #1351

MarqRazz:
@mamoll suggested I recommend any updates or add to OMPL's build system so I'm just seeing if you guys are open to this. It seems like better default in general and helps guard against the `SIGILL` case but if you don't want it I can try to move it into the release repo.

## Downsides for non-ROS users                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                
  **TL;DR:** For most non-ROS users, none — their build paths don't set `DEB_BUILD_OPTIONS`, so they keep the existing native default. The guard is genuinely scoped to the case it claims.                                                                                     
   
  ### Who is affected                                                                                                                                                                                                                                                           
                  
  | User | Sets `DEB_BUILD_OPTIONS`? | Net change |
  | --- | --- | --- |
  | Developer running `cmake ..` / `make` from source | No | None — keeps native build |                                                                                                                                                                                        
  | Researcher using upstream OMPL from source | No | None |
  | `pip install ompl` (scikit-build wheels) | No | None |                                                                                                                                                                                                                      
  | `add_subdirectory(ompl)` from a parent CMake project | Only if the parent is itself a Debian build | Tracks the parent (correct) |
  | conda-forge / Homebrew / vcpkg packagers | No | None — they still have to opt into portable manually if they want safe binaries (separate problem this PR doesn't solve) |                                                                                                  
  | Debian / Ubuntu's own (non-ROS) `ompl` maintainers using `dpkg-buildpackage` | Yes | Portable by default — **upside** for their users, who `apt install libompl-dev` onto assorted hardware |                                                                               
  | CI systems (GitHub Actions, GitLab CI, etc.) | Only if running Debian builds | Tracks intent (correct) |                                                                                                                                                                    
                                                                                                                                                                                                                                                                                
  ### Real downsides that do exist                                                                                                                                                                                                                                              
                  
  1. **A non-ROS Debian/Ubuntu packager who specifically wanted native builds** (rare — usually internal tooling that never leaves the build machine) now has to pass `-DVAMP_PORTABLE_BUILD=OFF` explicitly. Tiny friction, but it exists.                                     
                  
  2. **Edge case:** a developer who has `DEB_BUILD_OPTIONS` exported in their shell from a previous session, then runs `cmake ..` for an unrelated source build, gets portable instead of native. Unusual environment hygiene; recoverable with `-DVAMP_PORTABLE_BUILD=OFF` or  
  `unset DEB_BUILD_OPTIONS`.
                                                                                                                                                                                                                                                                                
  3. **PyPI wheel users still face the silent-SIGILL risk** if whoever uploaded the wheel built it on AVX-512 hardware. The current guard doesn't cover the `SKBUILD` code path. This isn't a *new* downside — it predates this PR — but it's a gap in the PR's safety story.   
   
  4. **Ongoing maintenance burden upstream:** ~5 lines of CMake. Easy to revert if it ever causes pain.                                                                                                                                                                         
                  
  The first three are minor and self-recoverable. The fourth is trivially small. None affect the typical non-ROS user (researcher / direct source builder / pip user with a self-built wheel).                                                                                  
                  
  ### Suggested extension: cover wheels too                                                                                                                                                                                                                                     
                  
  If we want symmetric coverage, the same logic applies to scikit-build wheels — they're the largest non-ROS redistribution channel for OMPL:                                                                                                                                   
                  
  ```cmake                                                                                                                                                                                                                                                                      
  if(DEFINED ENV{DEB_BUILD_OPTIONS} OR SKBUILD)
      set(_VAMP_PORTABLE_DEFAULT ON)                                                                                                                                                                                                                                            
  else()
      set(_VAMP_PORTABLE_DEFAULT OFF)                                                                                                                                                                                                                                           
  endif()         
  ```                                                                                                                                                                                                                                                                           
   
  This extends the guard to anyone publishing via `pip` without changing semantics for any other audience. PyPI users get the same protection apt users get. Happy to fold this into the PR if you'd like.

Thodoris1999:
Personally, I think the build system should not concern itself with distribution-specific configuration (e.g. reference `DEB_BUILD_OPTIONS`), since that's different per package manager. Otherwise the build system will eventually be filled with `if (RPM_BUILD_OPTIONS)`, `if (VCPKG_BUILD_OPTIONS)`, `if (HOMEBREW_BUILD_OPTIONS)` etc. Instead, we should provide options to accommodate any behavior, such that the distributions can get the build behavior they desire, which is possible by setting `VAMP_PORTABLE_BUILD`, cc @mamoll let me know if you agree.

I think the better question is whether `VAMP_PORTABLE_BUILD` should be by default on or off. I am currently running tests to see the performance difference, and will report back what I find

mamoll:
Although I see @Thodoris1999 points, I think we should make an exception for ROS & generic deb builds. I suspect the majority of users who depend on a prebuilt binary get ompl as a deb package (although I don't have hard data to back that up). We should make the deb build process as easy as possible. The solution @MarqRazz seems acceptable to me.

mamoll (approved):

CMakeLists.txt
```diff
@@ -8,7 +8,16 @@ option(VAMP_BUILD_PYTHON_BINDINGS "Build VAMP Python bindings (requires OMPL_BUI
 add_feature_info(VAMP_BUILD_PYTHON_BINDINGS "${VAMP_BUILD_PYTHON_BINDINGS}" "Whether to build VAMP Python bindings.")
 option(OMPL_BUILD_VAMP "Build VAMP submodule" ON)
 add_feature_info(OMPL_BUILD_VAMP "${OMPL_BUILD_VAMP}" "Whether to build VAMP submodule.")
-option(VAMP_PORTABLE_BUILD "Build VAMP with portable SIMD settings for distribution (package maintainers)" OFF)
+# Default to portable SIMD when invoked from a Debian/dpkg build (DEB_BUILD_OPTIONS
+# is exported by dpkg-buildpackage / debhelper, so this catches the ROS buildfarm
+# and any apt-builds-from-source automatically). Direct `cmake ..` developer
+# builds keep the native-ISA default.
+if(DEFINED ENV{DEB_BUILD_OPTIONS})
```
mamoll:
```suggestion
if(DEFINED ENV{DEB_BUILD_OPTIONS} OR SKBUILD)
```

mamoll (changes requested):
With this tiny change that @MarqRazz had already suggested I think we should merge this.

mamoll (approved):

PR #1424, opened 2026-05-09
Make OMPL_PYTHON_INSTALL_PREFIX default respect CMAKE_INSTALL_PREFIX

MarqRazz:
`py-bindings/CMakeLists.txt` defaults the install prefix to `Python_SITEARCH`, which is always an absolute system path. CMake's `install()` only prepends `CMAKE_INSTALL_PREFIX` to relative destinations, so the Python bindings always try to land in                       
  `/usr/local/lib/pythonX.Y/dist-packages` regardless of what prefix the user requests. Any non-root build with a custom prefix (colcon workspaces, container builds, `--prefix=$HOME/.local`) errors at install time. The Jazzy devel job [`Jdev__ompl__ubuntu_noble_amd64`    
  #116](https://build.ros2.org/job/Jdev__ompl__ubuntu_noble_amd64/116/consoleFull#console-section-13) hit exactly this:                                                                                                                                                         
                                                                                                                                                                                                                                                                                
  CMake Error at py-bindings/cmake_install.cmake:60 (file):                                                                                                                                                                                                                     
    file cannot create directory: /usr/local/lib/python3.12/dist-packages/ompl.
    Maybe need administrative privileges.                                                                                                                                                                                                                                       
                  
  This change keeps `Python_SITEARCH` as the default for system-wide installs (Debian builds via `DEB_BUILD_OPTIONS`, default `/usr` or `/usr/local` prefix) and switches to a path relative to `CMAKE_INSTALL_PREFIX` for everything else. Existing system-install and
  buildfarm-binarydeb flows are unchanged; explicit `-DOMPL_PYTHON_INSTALL_PREFIX=...` always wins.

### DIFF
--- a/py-bindings/CMakeLists.txt
+++ b/py-bindings/CMakeLists.txt
@@ -7,21 +7,33 @@ nanobind_add_module(_ompl
     ${PY_OMPL_SOURCES}
 )
 
-# Default to Python_SITEARCH only for system-wide installs; for custom
-# prefixes use a path relative to CMAKE_INSTALL_PREFIX so install() works
-# without root and without writing outside the requested prefix.
+# Ask Python's sysconfig where 'purelib' lands under CMAKE_INSTALL_PREFIX, then
+# strip the prefix to get a relative path. Matches ament_cmake_python's
+# algorithm so the result honors distro conventions (dist-packages on Debian,
+# site-packages elsewhere) and works for any prefix.
 if(NOT OMPL_PYTHON_INSTALL_PREFIX AND NOT SKBUILD)
-    if(DEFINED ENV{DEB_BUILD_OPTIONS}
-       OR CMAKE_INSTALL_PREFIX MATCHES "^/usr/?$"
-       OR CMAKE_INSTALL_PREFIX MATCHES "^/usr/local/?$")
-        set(_ompl_default_py_install "${Python_SITEARCH}")
-    else()
-        set(_ompl_default_py_install
-            "${CMAKE_INSTALL_LIBDIR}/python${Python_VERSION_MAJOR}.${Python_VERSION_MINOR}/site-packages")
+    set(_ompl_py_install_dir_code
+"import os, sysconfig
+schemes = sysconfig.get_scheme_names()
+kwargs = {'vars': {'base': '${CMAKE_INSTALL_PREFIX}'}}
+if 'deb_system' in schemes or 'osx_framework_library' in schemes:
+    kwargs['scheme'] = 'posix_prefix'
+elif 'rpm_prefix' in schemes:
+    kwargs['scheme'] = 'rpm_prefix'
+print(os.path.relpath(sysconfig.get_path('purelib', **kwargs),
+                      start='${CMAKE_INSTALL_PREFIX}').replace(os.sep, '/'))")
+    execute_process(
+        COMMAND "${Python_EXECUTABLE}" -c "${_ompl_py_install_dir_code}"
+        OUTPUT_VARIABLE _ompl_default_py_install
+        OUTPUT_STRIP_TRAILING_WHITESPACE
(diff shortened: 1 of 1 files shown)


Thodoris1999:
Similar to #1422, this is just a default. You can change the path or turn off the bindings completely. Again, trying to avoid distribution-specific configuration

mamoll:
This is a tricky one. @MarqRazz how does ROS figure out the right python dir suffix, given an install prefix?

MarqRazz:
the updated approach contains *zero* distro-specific logic in OMPL itself. It defers to Python's own `sysconfig` module, which each distro patches to know its own convention. The only "branching" we do is detecting which sysconfig
  *scheme* to use:

  ```python
  if 'deb_system' in schemes or 'osx_framework_library' in schemes:
      kwargs['scheme'] = 'posix_prefix'   # Debian/macOS-framework
  elif 'rpm_prefix' in schemes:
      kwargs['scheme'] = 'rpm_prefix'     # RHEL/Fedora
  # else: distro's default scheme
```
  That detection is itself distro-agnostic — it asks Python "what schemes do you support?" rather than checking for any particular OS. The whole snippet is copied verbatim from ament_cmake_python-extras.cmake (the _ament_cmake_python_get_python_install_dir macro) — we're matching the algorithm every ament-based ROS package uses.

  On how ROS figures it out given an install prefix: it asks Python directly. The key trick is sysconfig.get_path('purelib', vars={'base': CMAKE_INSTALL_PREFIX}) — sysconfig returns the install path it would use for that base. We then strip the prefix to get a relative
  path that CMake's install() will re-prepend at install time.

  The posix_prefix scheme is the one whose paths are templated on {base}. Debian's default scheme isn't — it hardcodes /usr/local/...dist-packages no matter what you pass in. That's why the explicit scheme='posix_prefix' is necessary on Debian: without it the base
  substitution is silently ignored. (Verified on Ubuntu Noble locally — the default scheme returned /usr/local/local/lib/... for base='/usr/local', which is obviously wrong.)

  On Ubuntu/Debian this lands at lib/python3.X/site-packages for every prefix, including /usr and /usr/local. That's the same path colcon's setup.bash adds to PYTHONPATH and the same path dh_python3 relocates to dist-packages during deb post-processing — so both the
  workspace flow and the buildfarm binarydeb flow keep working without OMPL having to know which one is happening.

  On "just turn off bindings or change the path": that's the existing workaround and it's what users are doing today. It works, but it requires every downstream packager — buildfarm release templates, distro maintainers, anyone consuming OMPL via colcon — to remember to
  pass -DOMPL_PYTHON_INSTALL_PREFIX=... or -DOMPL_BUILD_PYTHON_BINDINGS=OFF whenever they're not building as root with the default prefix. That's the failure mode that landed Jdev__ompl__ubuntu_noble_amd64 # 116 — the buildfarm template didn't remember, and the job failed
  at install time with Maybe need administrative privileges. The fix shifts the default from "wrong unless overridden" to "right unless overridden", with the override still available for anyone who wants the system path explicitly.

Thodoris1999:
Again, I do not think distribution specific logic should be included in the build system if the desired behavior can be achieved through offered flags. Is that an option (we mentioned to make an exception for debian builds in #1422 but this is for ROS only)? @mamoll @MarqRazz 

In the case we do want to make ROS builds work without messing with default flags however:
Switching the default from the active environment's SITEARCH to a path relative to CMAKE_PREFIX_PATH may work for ROS but may break other workflows, it is a bit arbitrary. The logic also seems complicated and may rely on ROS internals that are not standardized, and subject to beak with changes.

Is there a flag we can use to detect that the project is being built as a ROS project? Perhaps from colcon/ament? If that's that case, I think we can use some of the paths from [`ament_cmake_python`](https://docs.ros.org/en/iron/p/ament_cmake_python/) which is public API.

mamoll (approved):

PR #1430, opened 2026-05-26
Fix STRRT* Use-after-free in constructSolution + fix rewireGoalTree bugs

kaizer-nurik:
# Fix use-after-free in `STRRTstar::constructSolution()`

While testing solution refinement in STRRT*, I noticed that about ~5% of the runs crashed with a segmentation fault. After rerunning the failing cases with sanitizers and Valgrind, I found that the crash was caused by a use-after-free in `constructSolution()`.

`pruneGoalTree()` can invalidate the previous `goalMotion`. The old `goalMotion` was then reused as an argument in the recursive call to `constructSolution()`, even though it may already have been freed. This is also logically incorrect: after pruning the goal tree, the planner receives a new solution motion and should continue from that motion, not from the old one. The corresponding start-side connection is `newSolution->connectionPoint`.

The fix changes the recursive call from using the old `goalMotion` to using the newly found solution:

```cpp
constructSolution(newSolution->connectionPoint, newSolution, intermediateSolutionCallback, ptc);
```

I added:
- a stress test with a simple obstacle and path refinement, which triggers the recursive `constructSolution()` call and previously reproduced the segmentation fault;
- a deterministic test with a prebuilt tree that reproduces the same invalid access.

After the fix, all tests pass without segmentation faults.

I also fixed a typo in the local variable name near:

https://github.com/ompl/ompl/blob/4e9bbe060ffd4bd069d474d4a861b4de8d7c0701/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp#L631-L632

The variable should be named `reachedGoal`.

# Fix rewiring in `STRRTstar::rewireGoalTree()`

While reviewing the surrounding STRRT* code, I also found two bugs in `rewireGoalTree()`:

1. When a node is rewired to a better goal-tree parent, the algorithm updates the rewired node's `root`, but does not update the `root` field of its descendants. This can leave descendants pointing to the old goal root, even though their subtree has been moved under a different root.

   This is inconsistent with the logic used when orphaned nodes are rewired in `pruneGoalTree()`, where descendants are updated through `addDescendants()`.
   https://github.com/ompl/ompl/blob/4e9bbe060ffd4bd069d474d4a861b4de8d7c0701/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp#L970

2. The algorithm updates ancestor `numConnections` counters by `1`, even when the rewired subtree contains more than one connection. Since `numConnections` represents the number of start-tree connections in the node and all its descendants, ancestor counters must be adjusted by the rewired node's full `numConnections` value.
https://github.com/ompl/ompl/blob/4e9bbe060ffd4bd069d474d4a861b4de8d7c0701/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp#L1020

   https://github.com/ompl/ompl/blob/4e9bbe060ffd4bd069d474d4a861b4de8d7c0701/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp#L1034

I added deterministic regression tests for both rewiring issues. The tests fail on the old implementation and pass after the fixes.

### DIFF
--- a/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
+++ b/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
@@ -1017,7 +1017,7 @@ bool ompl::geometric::STRRTstar::rewireGoalTree(Motion *addedMotion)
                 Motion *p = otherMotion->parent;
                 while (p != nullptr)
                 {
-                    p->numConnections-=otherMotion->numConnections;
+                    p->numConnections -= otherMotion->numConnections;
                     p = p->parent;
                 }
             }
@@ -1032,7 +1032,7 @@ bool ompl::geometric::STRRTstar::rewireGoalTree(Motion *addedMotion)
             {
                 for (Motion *c : queue.front()->children)
                 {
-                    queue.push(c); 
+                    queue.push(c);
                 }
                 queue.front()->root = addedMotion->root;
                 queue.pop();
@@ -1043,7 +1043,7 @@ bool ompl::geometric::STRRTstar::rewireGoalTree(Motion *addedMotion)
                 Motion *p = otherMotion->parent;
                 while (p != nullptr)
                 {
-                    p->numConnections+=otherMotion->numConnections;
+                    p->numConnections += otherMotion->numConnections;
                     p = p->parent;
                 }
                 if (otherMotion->root->as<ompl::base::CompoundState>()
--- a/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
+++ b/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
@@ -1025,7 +1025,7 @@ bool ompl::geometric::STRRTstar::rewireGoalTree(Motion *addedMotion)
             otherMotion->parent = addedMotion;
             otherMotion->root = addedMotion->root;
             addedMotion->children.push_back(otherMotion);
-            // change root state of descendents
+            // change root state of descendants
             std::queue<Motion *> queue;
             queue.push(otherMotion);
             while (!queue.empty())
--- a/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
+++ b/src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
@@ -628,8 +628,8 @@ void ompl::geometric::STRRTstar::constructSolution(
     }
 
     bestSolution_ = path;
-    auto reachedGaol = path->getState(path->getStateCount() - 1);
-    bestTime_ = reachedGaol->as<ompl::base::CompoundState>()->as<ompl::base::TimeStateSpace::StateType>(1)->position;
+    auto reachedGoal = path->getState(path->getStateCount() - 1);
+    bestTime_ = reachedGoal->as<ompl::base::CompoundState>()->as<ompl::base::TimeStateSpace::StateType>(1)->position;
 
     if (intermediateSolutionCallback)
     {
@@ -647,7 +647,7 @@ void ompl::geometric::STRRTstar::constructSolution(
 
     // loop as long as a new solution is found by rewiring the goal tree
     if (newSolution != nullptr)
-        constructSolution(newSolution->connectionPoint, goalMotion, intermediateSolutionCallback, ptc);
+        constructSolution(newSolution->connectionPoint, newSolution, intermediateSolutionCallback, ptc);
 }
 
 void ompl::geometric::STRRTstar::pruneStartTree()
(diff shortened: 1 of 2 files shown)


mamoll:
This looks right to me, but it'd great if @aorthey or @frangrothe could weigh in.

src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
```diff
@@ -1017,21 +1017,33 @@ bool ompl::geometric::STRRTstar::rewireGoalTree(Motion *addedMotion)
...
-                    p->numConnections--;
+                    p->numConnections-=otherMotion->numConnections;
                     p = p->parent;
                 }
             }
             removeFromParent(otherMotion);
             otherMotion->parent = addedMotion;
             otherMotion->root = addedMotion->root;
             addedMotion->children.push_back(otherMotion);
+            // change root state of descendents
```
vhartman: nit: should be descendants

tests/geometric/spacetime/STRRTstar_test.cpp
```diff
@@ -18,6 +18,16 @@ bool isStateValid(const base::State *state)
...
 }
 
+bool isStateValidHyperball(const ompl::base::State *state)
+{
+    const auto *values =
+        state->as<ompl::base::CompoundState>()->as<ompl::base::RealVectorStateSpace::StateType>(0)->values;
+    double s = 0.0;
+    for (unsigned int i = 0; i < 6; ++i)
+        s += values[i] * values[i];
+    return (s > 0.1) && (s < 24);
```
vhartman: does not matter so much since it is in a test, but where does this 24 come from?

tests/geometric/spacetime/STRRTstar_test.cpp
```diff
@@ -200,4 +506,38 @@ BOOST_AUTO_TEST_CASE(spacetime_planning_3d_diagonal)
...
+// Test case with fixed tree for broken recursive and use-after-free in constructSolution
+BOOST_AUTO_TEST_CASE(spacetime_rewore_goal_tree_updates_descendant_roots)
+{
+    STRRTstarTestAccess planner;
+
+    BOOST_CHECK_NO_THROW(planner.rewireGoalTreeUpdatesDescendantRoots());
+}
+
+
+// // Stress test for pruning and final time optimising with different seeds
```
vhartman, line 512: should this commented test be kept?

vhartman (commented):
Thanks for the fix! As far as I remember we have never encountered this for some reason.

Code changes look good to me, but I didn't have a chance to run this. I would trust you that it compiles/passes the tests.

src/ompl/geometric/planners/rrt/src/STRRTstar.cpp
kaizer-nurik: Thanks! Fixed.

tests/geometric/spacetime/STRRTstar_test.cpp
kaizer-nurik: While constructing the stress test to trigger segfaults, I tried to set vector space bounds to >1, and 24 is relic from those experiments. Removing  && (s < 24) would still trigger segfault (tested just now).

tests/geometric/spacetime/STRRTstar_test.cpp
kaizer-nurik: Yes! This is the stress test. I commented it to validate rewireGoalTree tests and forgot to uncomment it.  I pushed new commit, where I uncommented it + fixed typos in comments.  Thanks for noticing!

aorthey:
LGTM. Thanks @NiklasHargus for checking this locally. @kaizer-nurik Please ensure that clang-format is running through and then we can merge this in.

vhartman:
Yeah looks good to me with the changes. Thanks for the fixes!

aorthey (approved):

kaizer-nurik:
@aorthey I found some strange behaviour with clang-format: If i run as recommended in the[ style guide
](https://github.com/ompl/ompl/blob/8dce6ff6e11bbfcc781ba43d4f73492c386211f9/doc/markdown/styleGuide.md?plain=1#L88)

```
clang-format --i -style=file './tests/geometric/spacetime/STRRTstar_test.cpp'
clang-format --i -style=file './src/ompl/geometric/planners/rrt/src/STRRTstar.cpp'
````

then new auto formatted code won't pass 

```
pre-commit run --show-diff-on-failure --color=always --all-files
```

on a clang-format stage.

So I commited changes made by ```pre-commit run```. Is it correct?

PR #1433, opened 2026-05-29
ci: produce release tarball with vendored submodules

MarqRazz:
# Add release-tarball workflow so downstream packagers get `external/{vamp,nanobind}`

Adds `.github/workflows/release_tarball.yml`. On every published GitHub Release it builds `ompl-<tag>.tar.gz` with `external/vamp` and `external/nanobind` pre-populated, sanity-checks that both `CMakeLists.txt` files are present, and attaches the tarball + `.sha256` to the Release. **No changes to `main`, the build system, or the existing dev/maintainer workflow** — `.gitmodules` is still the source of truth for upstream pins.

## Why

`git archive` does not recurse submodules ([docs](https://git-scm.com/docs/git-archive)), and every release artifact path is rooted in `git archive`: bloom, `dpkg-source`, pip sdist, and GitHub's auto-generated tarball. OMPL silently disables both features when the submodule trees are missing:

```
CMake Warning at CMakeModules/VampConfig.cmake:25 (message):
  VAMP submodule not found. ...
CMake Warning at CMakeLists.txt:97 (message):
  OMPL_BUILD_PYTHON_BINDINGS is ON but Nanobind is not found. ...
```

(reproduced from the `ros-rolling-ompl 1.7.0` buildfarm log). Net effect: **every binary OMPL release we've shipped has had Python bindings and VAMP silently disabled** despite both defaulting to `ON`. PyPI wheels escape this only because `wheels_build.yml` does its own `submodules: recursive` checkout. The maintainers were already aware of the issue — `doc/releaseChecklist.txt` had a manual `cmake --build … -t package_source` step to work around it. This PR removes that manual step.

## Why this approach and not the others

- **Vendor into `main`**: ~120 MB of new files (mostly `vamp/resources/`), loses the submodule pin tracking. No.
- **`FetchContent`**: the ROS buildfarm passes `-DFETCHCONTENT_FULLY_DISCONNECTED=ON` (so does Debian). Would make OMPL un-buildable on those platforms.
- **System `nanobind` via `find_package`**: not in apt yet, doesn't help VAMP.
- **bloom-side hook**: bloom has no first-class hook for this.

The workflow is the smallest change that fixes every consumer at once, doesn't ask the project to take on a structural commitment, and is trivially revertible (delete the file).

## What downstream packagers do

Point their source URL at the new asset:

```
https://github.com/ompl/ompl/releases/download/<tag>/ompl-<tag>.tar.gz
```

For ROS, the `ompl-release` bloom track gets `vcs_type: tar` with that URL templated by `:{version}`. I'll handle that as part of the 2.0.0 bloom release.

Also updates `doc/releaseChecklist.txt`: drops the manual `package_source` step (the workflow does it on `release: published`) and points devs at the workflow as the canonical "what does an OMPL release archive look like" reference.

### DIFF
--- a/.github/workflows/release_tarball.yml
+++ b/.github/workflows/release_tarball.yml
@@ -1,0 +1,98 @@
+name: Release Tarball
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Existing tag to build a release tarball for (e.g. 2.0.0)
+        required: true
+        type: string
+
+jobs:
+  build-tarball:
+    name: Build ompl-${{ github.event.release.tag_name || inputs.tag }}.tar.gz
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - name: Resolve tag
+        id: tag
+        run: |
+          if [ "${{ github.event_name }}" = "release" ]; then
+            echo "name=${{ github.event.release.tag_name }}" >> "$GITHUB_OUTPUT"
+          else
+            echo "name=${{ inputs.tag }}" >> "$GITHUB_OUTPUT"
+          fi
+
+      - name: Check out tag with submodules
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ steps.tag.outputs.name }}
+          submodules: recursive
+          fetch-depth: 1
+
+      - name: Vendor submodule content into the working tree
+        # actions/checkout has already fetched submodule content. Strip the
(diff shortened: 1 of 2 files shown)


mamoll (approved):
This is great.  I have been frustrated by git submodules in creating releases in the past and it's nice that you have found a much better way to do this. I'll give other OMPL maintainers a few days to review it, but I am inclined to merge it.

Thodoris1999:
Pretty neat! I am going to take a closer look this weekend

PR #1435, opened 2026-06-08
Fix null goalState_ dereference in FMT (heuristics + extendedFMT)

Raphclou14:
## Bug description
  
  When using FMT via MoveIt2 (Humble, OMPL 1.7.0), `move_group` crashes with a segfault during motion planning if the goal pose is in a configuration where the IK solver cannot find a valid solution. For example when the target is near the workspace boundary and all IK solutions place a link in collision with the scene.
  
  The crash occurs because `assureGoalIsSampled()` fails to assign `goalState_` (it stays `nullptr`), and two call sites then pass it unconditionally to `motionCostHeuristic()`, which forwards it to
  `StateSpace::distance()`:

  -2  ompl::geometric::FMT::solve(ompl::base::PlannerTerminationCondition const&)
  -1  ompl::base::PathLengthOptimizationObjective::motionCostHeuristic(...)
  -0 ompl_interface::ModelBasedStateSpace::distance(ompl::base::State const*, ompl::base::State const*) const
  Segmentation fault (Address not mapped to object [0x8])

  ## Root cause

  `assureGoalIsSampled()` samples the goal state and calls `si_->getStateValidityChecker()->isValid()` on it. When `isValid()` returns `false` (IK found joint angles but they are in collision), the state is
  discarded and `goalState_` is never assigned — it remains `nullptr`.

  Two call sites then dereference it unconditionally:

  **`expandTreeFromNode()`** - heuristics path:
  ```cpp
  // Before fix - no null guard
  x->setHeuristicCost(opt_->motionCostHeuristic(x->getState(), goalState_));
```
   **`solve()`**  - extendedFMT branch:
 ```cpp
  // Before fix - no null guard
  m->setHeuristicCost(opt_->motionCostHeuristic(m->getState(), goalState_));
```
  ## Fix

  - Initialize goalState_ to nullptr explicitly in the header (in-class initializer) and in the constructor initializer list.
  - Guard both call sites with if (goalState_ != nullptr) 

##  Regression test

  tests/geometric/test_fmt_null_goal.cpp adds three test cases. Each sets up a 2D planning problem where the goal region is entirely invalid (the validity checker rejects every goal sample), forcing goalState_ to remain nullptr. The tests verify the planner returns FAILURE gracefully instead of crashing:

  - fmt_null_goal_heuristics — covers the expandTreeFromNode() path (heuristics=true)
  - fmt_null_goal_extended — covers the solve() extendedFMT path (extended_fmt=true)
  - fmt_null_goal_heuristics_and_extended — both flags active simultaneously


src/ompl/geometric/planners/fmt/src/FMT.cpp
```diff
@@ -50,7 +50,8 @@
 #include <ompl/base/objectives/PathLengthOptimizationObjective.h>
 #include <ompl/geometric/planners/fmt/FMT.h>
 
-ompl::geometric::FMT::FMT(const base::SpaceInformationPtr &si) : base::Planner(si, "FMT")
+ompl::geometric::FMT::FMT(const base::SpaceInformationPtr &si)
+  : base::Planner(si, "FMT"), goalState_(nullptr)
```
mamoll: explicit initialization of goalState is unnecessary because of the brace initialization you added in the header.

mamoll (changes requested):
Thanks for the fix. Minor change requested.

src/ompl/geometric/planners/fmt/src/FMT.cpp
Raphclou14: Good catch, removed in the latest commit.

Raphclou14:
Great. Let me know if you need something else.

mamoll (approved):

mamoll:
Can you make this tiny formatting change to make our CI happy?:
```diff
-ompl::geometric::FMT::FMT(const base::SpaceInformationPtr &si)
-  : base::Planner(si, "FMT")
+ompl::geometric::FMT::FMT(const base::SpaceInformationPtr &si) : base::Planner(si, "FMT")
```
You can also configure your IDE to run clang-format on save or install [precommit](https://[pre-commit](https://pre-commit.com/).com/).

mamoll:
Oh, there is apparently also a merge conflict.

mamoll:
Closed via #1438

PR #1440, opened 2026-06-10
Adds bindings for virtual functions

r4stered:
Several trampoline classes declare NB_OVERRIDE entries for virtuals that are never .def()'d on the binding. When a Python subclass doesn't override one of them, the trampoline's getattr() lookup fails and nanobind raises:

```nanobind::detail::get_trampoline('...::<method>()'): lookup failed!```

instead of falling back to the C++ default. In practice any Python subclass crashes on first use:

- OptimizationObjective — RRT* calls infiniteCost() during setup
- Planner — SimpleSetup::setup() calls Planner::setup()
- Goal — start-state processing calls isStartGoalPairValid()
- Constraint — the default project() calls jacobian() (its .def was commented out; the unqualified &Constraint::jacobian is ambiguous — bound via nb::overload_cast)

Bind every trampolined virtual on its class, per the nanobind docs pattern — getattr then resolves to an nb_method, which the trampoline treats as "no Python override" and routes to the C++ default. ostream&-taking virtuals (print, printProperties, printSettings) are bound as std::string-returning lambdas, mirroring the existing ```__repr__``` bindings. PathLengthOptimizationObjective is fixed transitively through the base-class binding.

### DIFF
--- a/py-bindings/base/Constraint.cpp
+++ b/py-bindings/base/Constraint.cpp
@@ -51,7 +51,13 @@ void ompl::binding::base::init_Constraint(nb::module_ &m)
         .def("getMaxIterations", &ob::Constraint::getMaxIterations)
         .def("setTolerance", &ob::Constraint::setTolerance, nb::arg("tolerance"))
         .def("setMaxIterations", &ob::Constraint::setMaxIterations, nb::arg("iterations"))
-        // .def("jacobian", &ob::Constraint::jacobian, nb::arg("x"), nb::arg("out"))
+        .def("jacobian",
+             nb::overload_cast<const Eigen::Ref<const Eigen::VectorXd> &, Eigen::Ref<Eigen::MatrixXd>>(
+                 &ob::Constraint::jacobian, nb::const_),
+             nb::arg("x"), nb::arg("out"))
+        .def("jacobian",
+             nb::overload_cast<const ob::State *, Eigen::Ref<Eigen::MatrixXd>>(&ob::Constraint::jacobian, nb::const_),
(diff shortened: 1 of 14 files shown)


WeihangGuo:
Thank you @r4stered. The PR looks good to me. Can you add the corresponding unit tests?

r4stered:
> Thank you @r4stered. The PR looks good to me. Can you add the corresponding unit tests?

I have added unit tests in the last commit

mamoll (approved):

mamoll:
@WeihangGuo approving & merging this so that I can release a bug fix release. See if this PR suggests that changes in other Python bindings are also needed (like the big pending PR to complete the Python bindings) or the documentation for the Python bindings.

mamoll:
@r4stered any idea why pytest would fail in the build_wheels workflow? Locally, pytests pass, but they [fail in CI](https://github.com/ompl/ompl/actions/runs/27843795099/job/82408466412) like so:
>       constraint.jacobian(x, out)
  E       TypeError: jacobian(): incompatible function arguments. The following argument types are supported:
  E           1. jacobian(self, x: numpy.ndarray[dtype=float64, shape=(*), writable=False], out: numpy.ndarray[dtype=float64, shape=(*, *)]) -> None
  E           2. jacobian(self, state: ompl._ompl.base.State, out: numpy.ndarray[dtype=float64, shape=(*, *)]) -> None
  E       
  E       Invoked with types: test_constraint.UnitSphere, ndarray, ndarray

r4stered:
@mamoll I believe i fixed the underlying issue in #1445

PR #1445, opened 2026-06-22
Fix row/column order restriction in constraint bindings

r4stered:
Fixes the build wheels issue with #1440 

Constraint.jacobian's out parameter bound to a mutable Eigen::Ref<Eigen::MatrixXd>, which nanobind only accepts in place when the numpy array is column-major — and whether a 1×N jacobian array satisfies that is numpy version dependent, so the test passed under the CI numpy but failed under the older numpy cibuildwheel pins for the wheel build. The fix binds the output via nb::DRef (a dynamic-stride Eigen::Ref), which accepts any array layout on any numpy version.


mamoll:
If the cibuildwheel version is the problem, wouldn't it be easier to bump the cibuildwheel version in our pipeline? I don't think there's a strong reason to stick to version pypa/cibuildwheel@v2.23.2. Any version that can build portable binaries for python version >=3.10 is fine.

r4stered:
> If the cibuildwheel version is the problem, wouldn't it be easier to bump the cibuildwheel version in our pipeline? I don't think there's a strong reason to stick to version pypa/cibuildwheel@v2.23.2. Any version that can build portable binaries for python version >=3.10 is fine.

The version of cibuildwheel isnt the root issue, its the fact that the api without this MR restricts you to column major outputs, so something like this 

constraint.jacobian(x, np.zeros((1,3)))

would give you a TypeError. 

whatever version of numpy is grabbed when testing the wheel would determine if the API as is would work or not, this change makes it more robust so we dont have to pin the numpy version that happens to work.

mamoll (approved):

mamoll:
Thanks for the fix and explanation!

Issue #1429, opened 2026-05-13
Request for new release to support ROS 2 Lyrical (Ubuntu 26.04)


mjcarroll:
Hello,

I am working on unblocking the ROS 2 Lyrical (Ubuntu 26.04 Resolute) build farm. Currently, OMPL 1.7.0 is failing to build because it explicitly requests the Boost `system` and `filesystem` components, which are problematic/header-only in Boost 1.90.

I've verified that the current `main` branch builds successfully on Resolute because these dependencies were removed in recent refactors.

Would it be possible to cut a new release (e.g., 1.7.1 or 1.8.0) that includes these build system modernizations? This would allow us to update the ROS 2 Lyrical distribution to a working version.

I previously opened #1428 to address this, but since `main` is already healthy, a new release is the preferred path forward.

Thanks!

Thodoris1999:
Hi Michael,

Does 2.0.0 work for you [https://github.com/ompl/ompl/releases/tag/2.0.0](https://github.com/ompl/ompl/releases/tag/2.0.0)?

mjcarroll:
Yeah, 2.0.0 looks like it should work without a problem.  Can someone make a release of that into Lyrical and Rolling?

Thodoris1999:
> Yeah, 2.0.0 looks like it should work without a problem. Can someone make a release of that into Lyrical and Rolling?

Sure, I am not familiar with the procedure but I am happy to help if someone can walk me through it cc @mamoll

mini-1235:
Any updates on this? We would love to get the 2.0.0 released for Lyrical and Rolling to unblock the nav2 Lyrical release

mamoll:
@mjcarroll I released OMPL 2.0.1 today. @MarqRazz was (also?) planning to create a ROS release, so coordinate with him if you want to get out ASAP.

MarqRazz:
I have triggered the released of 2.0.1 for [Rolling](https://github.com/ros/rosdistro/pull/52189#pullrequestreview-4538179404) and [Lyrical](https://github.com/ros/rosdistro/pull/52190#pullrequestreview-4538185150) today and updated [ompl-release](https://github.com/ros2-gbp/ompl-release/commit/c116564debff6eb25df8e6154aa1b76e89dfbff2) to have Rolling/Lyrical distro's use the release tarball for vendored submodules.

Sorry it took a while to get all of the new 2.0 features ready for ros-distro.

MarqRazz:
[...]

PR #1446, opened 2026-06-23
Vendor vamp deps

MarqRazz:
I tried to release OMPL to ROS 2 Rolling/Lyrical with VAMP enabled but discovered that it has several transient dependencies that are not currently vendored with our existing `release_tarball.yml` workflow. Because it is blocking several packages (#1429) I disabled VAMP for the `2.0.1` release to unblock them. This PR should fix that issue and when we are ready to release the next version of OMPL I can update Rolling/Lyrical with VAMP enabled.

VAMP fetches its CPM-managed deps (`nigh, pdqsort, SIMDxorshift`, and `nanobind` when Python bindings are enabled) from the network at configure time, and `FetchInitCPM.cmake` raw-downloads `CPM.cmake` itself.
None of those resolve in an offline build environment (FETCHCONTENT_FULLY_DISCONNECTED=ON) most notably, downstream packagers like the ROS buildfarm cannot build VAMP from it.

This change extends `release_tarball.yml` to:
- Discover every `CPMAddPackage("gh:...")` call by parsing `external/vamp/cmake/*.cmake` and clone each into `external/vamp/external/<name>/` (with recursive submodules, since e.g. `nanobind` has an `ext/robin_map submodule`).
- Parse `CPM.cmake's URL + SHA256` from `FetchInitCPM.cmake` and vendor it to `external/vamp/cmake/CPM.cmake`.
- Patch each call site with a vendor-first conditional that uses the vendored copy when present.
- Run a defensive grep that fails the workflow if any unguarded fetch pattern remains (so a future VAMP submodule bump that adopts a new CPM call shape can't silently break offline builds).

## Verified
- Offline `cmake -DFETCHCONTENT_FULLY_DISCONNECTED=ON` configure completes against the produced tarball inside `unshare -rn`.
- `demo_VAMPPlanning` (the target that previously failed with "pdqsort.h: No such file or directory" on the [ROS buildfarm](https://build.ros2.org/job/Rbin_unv8_uRv8__ompl__ubuntu_resolute_arm64__binary/56/console)) compiles successfully from the tarball with no network access.
- I tested the [workflow locally on my fork](https://github.com/MarqRazz/ompl/actions/runs/28049714405) and verified that all transient deps are included.

### DIFF
--- a/.github/workflows/release_tarball.yml
+++ b/.github/workflows/release_tarball.yml
@@ -209,37 +209,28 @@ jobs:
           set -euo pipefail
           TAG="${{ steps.tag.outputs.name }}"
           BASE="ompl-${TAG}"
+          LST=$(mktemp)
+          tar -tzf "${BASE}.tar.gz" > "$LST"
+
           # Vendored submodule CMakeLists must be present.
-          tar -tzf "${BASE}.tar.gz" | grep -E \
-            "^${BASE}/external/(vamp|nanobind)/CMakeLists.txt$" \
+          grep -qE "^${BASE}/external/(vamp|nanobind)/CMakeLists.txt$" "$LST" \
             || { echo "::error::vendored submodule CMakeLists not in tarball"; exit 1; }
+
           # VAMP's CPM transitive deps must be vendored for offline builds.
-          # Don't hardcode names — discover from the tarball itself.
-          vendored=$(tar -tzf "${BASE}.tar.gz" \
-            | grep -oE "^${BASE}/external/vamp/external/[^/]+/$" | sort -u)
+          vendored=$(grep -oE "^${BASE}/external/vamp/external/[^/]+/$" "$LST" | sort -u)
           if [ -z "$vendored" ]; then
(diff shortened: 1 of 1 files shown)


mamoll:
@zkingston @claytonwramsey this seems like the kind of thing that should really be fixed in the upstream VAMP repo. How attached are you to the CPM stuff? That stuff just doesn't work when building binary packages for package managers. I have heard the same thing from the Debian maintainer for libompl-dev.

claytonwramsey:
I think we are not especially attached to it -- it was a choice of convenience more than anything else. Vendoring the external VAMP deps is probably the best solution; @zkingston knows more about the build setup than I do so I will wait for a second opinion from him.

zkingston:
Yes, not particularly attached to CPM, was a convenience. Vendoring the core deps should be fine and I believe they are relatively small.

MarqRazz:
I'm open to any path forward from here. We could merge this as is and then remove/update it to the new process when its ready to release or we could hold off on these changes and I could wait till VAMP is ready.

mamoll (approved):
thanks for doing this!

PR #1447, opened 2026-07-02
Add controlMotionCost for control planners and paths

ZhuoyunZhong:
**Overview**

This PR aims to improve support for optimization objectives for control planners and control paths. 

The main change is the introduction of a new virtual method in _base::OptimizationObjective_:
```
virtual Cost controlMotionCost(const State *s1, const control::Control *c, unsigned int steps, const State *s2) const;
```

**Motivation**

For geometric planners, an edge cost depending only on two states is often sufficient. For control planners, however, the cost of an edge may depend on both the state and the control used to propagate from that state. An example would be the energy consumption (cost) of a car depends on both the terrain (state) it drives on and the applied acceleration and steering (control).

Currently, _OptimizationObjective_ only has _motionCost(s1, s2)_ and _controlCost(c, steps)_ separately. The optimal kinodynamic planner, such as **SST**, can only compute them separately and add them together. Having _controlMotionCost_ allows users to override it when a kinodynamic edge cost depends jointly on the states and controls.

**Changes**

This PR makes the following changes:

- Adds _OptimizationObjective::controlMotionCost(...)_. The default implementation returns `combineCosts(motionCost(s1, s2), controlCost(c, steps));`.
- Adds Python binding support for overriding _controlMotionCost(...)_.
- Updates the two control planners, _control::SST_ and _control::HySST_, to use _controlMotionCost(...)_.
- Implements _control::PathControl::cost(...)_, which was left unimplemented before. The implementation mimics that of  _geometry::PathGeometry::cost(...)_, but with the porposed _controlMotionCost(...)_.

Finally, I also noticed that the current implementation of 

_ProblemDefinition::addSolutionPath(const PathPtr &path, ...)_

creates a planner solution and adds it directly, without setting the cost of the solution. As a result, solution paths added through this overload may not be ordered according to the provided optimization objective, but by the path length. This is especially problematic for user code that adds multiple solution paths and expects the best one to be selected according to the optimization objective.

- Adds a proper cost setup to the solution in _ProblemDefinition::addSolutionPath(const PathPtr &path, ...)_. Given that _control::PathControl::cost(...)_ is now implemented, this should work for both _PathGeometry_ and _PathControl_
```
if (optimizationObjective_)
{
    auto solCost = path->cost(optimizationObjective_);
    sol.setOptimized(optimizationObjective_, solCost, optimizationObjective_->isSatisfied(solCost));
}
```

### DIFF
--- a/src/ompl/base/src/ProblemDefinition.cpp
+++ b/src/ompl/base/src/ProblemDefinition.cpp
@@ -426,6 +426,11 @@ void ompl::base::ProblemDefinition::addSolutionPath(const PathPtr &path, bool ap
     PlannerSolution sol(path);
     if (approximate)
         sol.setApproximate(difference);
+    if (optimizationObjective_)
+    {
+        auto solCost = path->cost(optimizationObjective_);
+        sol.setOptimized(optimizationObjective_, solCost, optimizationObjective_->isSatisfied(solCost));
+    }
     sol.setPlannerName(plannerName);
     addSolutionPath(sol);
 }
--- a/py-bindings/base/OptimizationObjective.cpp
+++ b/py-bindings/base/OptimizationObjective.cpp
@@ -14,7 +14,7 @@ void ompl::binding::base::init_OptimizationObjective(nb::module_ &m)
 {
     struct PyOptimizationObjective : ob::OptimizationObjective
     {
-        NB_TRAMPOLINE(ob::OptimizationObjective, 19);
+        NB_TRAMPOLINE(ob::OptimizationObjective, 20);
         // Optional override
         bool isSatisfied(ob::Cost c) const override
         {
@@ -61,6 +61,13 @@ void ompl::binding::base::init_OptimizationObjective(nb::module_ &m)
         ob::Cost controlCost(const oc::Control *c, unsigned int steps) const override
         {
             NB_OVERRIDE(controlCost, c, steps);
+        }
+
+        // Optional override
+        ob::Cost controlMotionCost(const ob::State *s1, const oc::Control *c, unsigned int steps,
+                                   const ob::State *s2) const override
+        {
+            NB_OVERRIDE(controlMotionCost, s1, c, steps, s2);
         }
 
         // Optional override
--- a/src/ompl/base/OptimizationObjective.h
+++ b/src/ompl/base/OptimizationObjective.h
@@ -120,6 +120,13 @@ namespace ompl
              * steps. The default implementation uses the identityCost. */
             virtual Cost controlCost(const control::Control *c, unsigned int steps) const;
 
+            /** \brief Get the cost of the motion that starts at state \e s1, applies control \e c for \e steps
+             * propagation steps, and arrives at state \e s2.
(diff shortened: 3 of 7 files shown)


Thodoris1999:
I think this is really nice, the only comment I have is to fix the pre-commit format checks.

@mamoll quick summary:
- The new overload allows cost definition for motions that depend jointly on both the control and states
- Adds an implementation for `PathControl::cost`, necessary for selecting best paths in optimizing kinodynamic planning
- No API breaks
- HySST and SST are the only optimizing control planners as far as I know and the PR adds support for `controlMotionCost` to both of them

ZhuoyunZhong:
Sorry, I didn't know about the pre-commit format check before. I have amended the format fix.

Thodoris1999 (approved):

src/ompl/base/src/ProblemDefinition.cpp
```diff
@@ -426,6 +426,11 @@ void ompl::base::ProblemDefinition::addSolutionPath(const PathPtr &path, bool ap
     PlannerSolution sol(path);
     if (approximate)
         sol.setApproximate(difference);
+    if (optimizationObjective_)
```
mamoll, line 429: Why is this necessary? It looks like existing optimizing planners already do this.

mamoll (commented):
I have one small clarification question, but looks good!

src/ompl/base/src/ProblemDefinition.cpp
ZhuoyunZhong: Good question. The explanation is actually a bit long.

Most existing optimizing planners like **RRT*** already handle this themselves by constructing a `PlannerSolution`, setting the optimized cost, setting the solution property, and then calling:

`ProblemDefinition::addSolutionPath(const PlannerSolution &sol)`.

But I noticed that this convenient overload

`ProblemDefinition::addSolutionPath(const PathPtr &path, ...)`

currently creates a `PlannerSolution` internally, but does not set the optimization objective or objective cost. Since PlannerSolution falls back to path length when no optimized cost is set, any planner that uses this convenience overload will have its solutions ordered by path length rather than by the provided `OptimizationObjective`.

I ran into this issue while looking at rewriting **AORRTC** to use `OptimizationObjective`. At the moment, AORRTC uses geometric path length directly as the cost and calls this `PathPtr` overload when adding multiple solution paths. If AORRTC (or a future optimal planner) is updated to support custom `OptimizationObjective` but continues to use this overload, the stored solutions would not be sorted according to the objective.

So yeah, it is true that planners like **RRT*** already construct and add `PlannerSolution` themselves. (Maybe they knew this convenient overload was not storing the cost of the path and had to use the other overload?). My motivation is to make the simpler `addSolutionPath(path, ...)` overload behave correctly when a problem definition has an optimization objective. This makes the overload correct, safer, and more useful for future optimizing planners, including a possible future version of **AORRTC** that supports custom objectives.

mamoll: thanks for the detailed explanation. this makes sense to me now

PR #1448, opened 2026-07-02
Add control planner SSTstar (SST*)

ZhuoyunZhong:
**Summary**

This PR adds a control planner implementation of **SST**.

The implementation is adapted from the existing OMPL **SST** planner, but adds the radius-shrinking schedule described in the original SST/SST* paper [1] to make the planner **asymptotically optimal**, rather than only **asymptotically near-optimal** like SST. 

[1] [Asymptotically Optimal Sampling-based Kinodynamic Planning](https://arxiv.org/pdf/1407.2896)

**Core Implementation Details**

The core difference is that SST uses fixed selection and pruning radii, while SST* progressively shrinks these radii during planning.

In this implementation, SST* starts with the same sparse tree / witness-set structure as SST, but planning proceeds in batches. After each batch, the selection radius and pruning radius are multiplied by a shrink factor: 
```
selectionRadius_ = selectionRadius_ * shrinkFactor_;
pruningRadius_ = pruningRadius_ * shrinkFactor_;
```
The number of iterations in later batches is increased according to the SST* schedule from the paper:
```
const double mult = (1.0 + std::log(j)) * std::pow(1.0 / shrinkFactor_, (d + l + 1.0) * j);
innerIteration = static_cast<unsigned long>(std::ceil(mult * static_cast<double>(initialN_)));
```
This allows the planner to become less sparse over time, which is the key change needed for asymptotic optimality.

**Changes**

- Added **control::SSTstar** 
  - Use the existing **SST** tree expansion, witness selection, pruning, and solution tracking logic.
  - Add **SST***-specific parameters `initial_iteration` and `shrink_factor`.
  - Shrink `selection_radius` and `pruning_radius` over planning batches. Increase the batch iteration count according to the SST* schedule.

- Added related python binding files for **control::SSTstar**.

**Notes**

This PR is intentionally split into two commits to make the implementation easier to review.

The first commit copies the existing **control::SST** implementation and renames it to control::SSTstar without changing anything. The second commit contains the actual **SST*** changes. Reviewing the second commit separately should make the behavioral differences between **SST** and **SST*** easier to inspect.

Also, I have a seperated pull request #1447 proposing a new cost function specifically for control planners. If that pull request gets passed, I will change this implementation to use the new cost function as well.


mamoll (commented):
would it be possible for SSTstar to derive from SST and override the solve method? It looks like the two planners are otherwise the same?

ZhuoyunZhong:
> would it be possible for SSTstar to derive from SST and override the solve method? It looks like the two planners are otherwise the same?

Good question. Actually, that may be better. I will think of it and try to do it in the next few days.

ZhuoyunZhong:
I have updated the code based on @mamoll 's recommendation.

Summary:
- Refactored SSTstar to inherit from SST, reducing duplicated code.
- Added the missing Python bindings for SST (`py-bindings/control/planners/sst/SST.cpp‎`).
- Extended the Python control planner test script (`tests/pytests/test_control_planners.py`) to cover RRT, SST, and SSTstar.

PR #1449, opened 2026-07-02
Add control planner AORRT

ZhuoyunZhong:
**Summary**

This PR adds a control planner implementation of **AORRT**.

AORRT is an asymptotically optimal kinodynamic planner based on planning in an augmented **state-cost space**. The implementation follows the AO-x idea from Hauser and Zhou’s state-cost-space formulation, where optimal kinodynamic planning is transformed into a sequence of feasible planning problems under progressively tighter cost bounds. [1]

This implementation uses a **single-tree** version of AORRT and also supports objectives with a **terminal cost**, following the asymptotic optimality proof in [2].

[1] [Asymptotically Optimal Planning by Feasible Kinodynamic Planning in State-Cost Space](https://arxiv.org/pdf/1505.04098)
[2] [Terminal Matters: Kinodynamic Planning with a Terminal Cost and Learned Uncertainty in Belief State-Cost Space](https://arxiv.org/pdf/2605.09046)

## Core Implementation Details

Before an initial solution is found, AORRT behaves similarly to a vanilla control RRT: it samples states, samples controls, forward-propagates the system dynamics, and grows a feasible kinodynamic search tree.

After a solution is found, the planner switches to state-cost-space search. It samples both a state and a cost coordinate under the current best solution cost, uses an augmented distance metric in state-cost space, and prunes motions whose accumulated running cost exceeds the current cost bound. Each new solution tightens the bound and guides the next search toward lower-cost trajectories.

Terminal cost is handled when evaluating goal-reaching motions. The planner stores the accumulated running cost along the trajectory and evaluates the full solution cost as: `totalCost = runningCost + terminalCost(goalState)`. AORRT is proven to remain asymptotically optimal with this cost formulation (trajectory running cost + terminal cost).

**Changes**

- Added **control::AORRT**.
  - Grows a single kinodynamic search tree.
  - Starts with regular state-space nearest-neighbor search before the first solution.
  - Switches to augmented state-cost-space nearest-neighbor search after a cost bound is available.
  - Prunes motions above the current best cost bound.
  - Supports terminal-cost objectives if provided by **OptimizationObjective::terminalCost()**.

- Added Python bindings for **control::AORRT**.

**Notes**

I have a separate pull request #1447 proposing a new cost function specifically for control planners. If that pull request gets passed, I will change this implementation to use the new cost function as well.


claytonwramsey:
Does it make more sense to instead make a general implementation of the AO-x meta-algorithm by adding support for cost-augmented configuration spaces?

ZhuoyunZhong:
> Does it make more sense to instead make a general implementation of the AO-x meta-algorithm by adding support for cost-augmented configuration spaces?

That would be ideal, but I think it is difficult to implement as a generic wrapper around existing planners. AO-x is not just adding a cost dimension to the state space. It also requires cost-bounded sampling, cumulative-cost bookkeeping in motions, and pruning nodes above the current cost bound. These operations need to be integrated into the planner’s expansion logic itself, especially for control planners where the realized cost is only known after propagation.

So in practice we would still need to rewrite much of RRT’s (or EST's) internal logic. For that reason, this PR implements AORRT as a dedicated planner.

ZhuoyunZhong:
Just following up on this PR.

Are there any remaining concerns with the current AORRT implementation?

PR #1450, opened 2026-07-27
fix check for valid pitch in Vana and Vana-Owen state space

mamoll:
Previously pitch was only checked after first turn. This fixes it so that pitch is also checked after second turn.


mamoll:
This fixes one bug, but there's possibly another bug. Converting to draft while I am investigating...

PR #1456, opened 2026-08-13
Add build export dependencies for Boost libraries

SteveMacenski:
This is needed for downstream ROS binary builds to properly have boost as a dependency from OMPL. 

This will fix Nav2's Lyrical binary release. If you could re-bloom that would be amazing. I assume this causes issues for other OMPL users, though I see moveit seems to work because they straight up just `<depend>` on boost for `moveit_core`. We don't use `boost` anywhere in Nav2 except for OMPL so that's probably why we see it.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -31,6 +31,10 @@
   <exec_depend>libboost-serialization</exec_depend>
   <exec_depend>libflann</exec_depend>
 
+  <build_export_depend>libboost-program-options-dev</build_export_depend>
+  <build_export_depend>libboost-serialization-dev</build_export_depend>
+  <build_export_depend>libboost-system-dev</build_export_depend>
+  
   <export>
     <build_type>cmake</build_type>
   </export>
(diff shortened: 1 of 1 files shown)


Thodoris1999:
Thank you for pointing this out! Just a small pre-commit format issue on the newline whitespace.

CC @mamoll @MarqRazz for the re-bloom. Let me know if you want a new tag/github release

SteveMacenski:
That's what I get for doing this on the github UX :laughing: 

Resolved.

Thodoris1999 (approved):

anttix:
Fixes most of #1457 Eigen3 dependency should also be added.

MarqRazz:
> Let me know if you want a new tag/github release

Yes please, a new tag/release will allow me to enable VAMP for the binaries too.

Thodoris1999:
> > Let me know if you want a new tag/github release
> 
> Yes please, a new tag/release will allow me to enable VAMP for the binaries too.

Does this release look good? This is the first time using your new release tarball workflow
https://github.com/ompl/ompl/releases/tag/2.0.2

PR #1458, opened 2026-08-13
Export Eigen dependency required by omplConfig.cmake

### LINKED ISSUE #1457: ROS package does not export required Boost development dependency to downstream consumers
## Summary

The ROS binary package for OMPL 2.0.1 does not propagate the Boost development dependency required by its exported CMake configuration. This causes isolated downstream ROS builds to fail during CMake configuration.

This currently blocks the ROS 2 Lyrical binary build of Nav2's `nav2_smac_planner` package.

## Failure

Nav2 build-farm job:

https://build.ros2.org/job/Lbin_uR64__nav2_smac_planner__ubuntu_resolute_amd64__binary/5/consoleFull

The downstream package calls `find_package(ompl)`. OMPL's installed `omplConfig.cmake` then calls:

```cmake
find_dependency(Boost REQUIRED COMPONENTS serialization)
```

CMake fails because the isolated downstream environment does not contain either:

```text
BoostConfig.cmake
boost-config.cmake
```

## Cause

OMPL's `package.xml` declares the relevant development package only as a build dependency:

```xml
<build_depend>libboost-serialization-dev</build_depend>
```

That installs Boost while OMPL itself is built, but does not propagate it to packages consuming OMPL's exported CMake configuration.

The generated Lyrical Debian package therefore depends on the Boost serialization runtime library, but not the development package that provides the CMake metadata required by `find_dependency(Boost ...)`.

Full source builds or broad development images can mask this because a Boost development package is often already installed transitively. The isolated Nav2 binary job exposes the missing exported dependency.

## Expected result

Installing the ROS OMPL binary package should provide the complete dependency closure needed for a downstream package to successfully run `find_package(ompl)`.

The development dependencies referenced by `omplConfig.cmake` should be declared as exported dependencies, likely using `build_export_depend` (including `libboost-serialization-dev`; the other exported `find_dependency` calls may warrant the same audit).

## Related history

- #755 added the dependency lookup to `omplConfig.cmake`.
- #1312 / #1313 narrowed the ROS Boost dependencies, but the component development packages were retained only as build dependencies.

anttix:
## Summary

Export Eigen3 to downstream ROS packages because OMPL's installed CMake configuration unconditionally calls:

```cmake
find_dependency(Eigen3 REQUIRED)
```

The dependency is currently declared only as `build_depend`, so it is available while compiling OMPL but not propagated to packages that call `find_package(ompl)`.

This is a follow-up to #1456. That PR added the Boost development exports and resolves the immediate Nav2 SMAC build failure, but it did not export Eigen3.

Fixes #1457.

## Validation

Validated from clean `ros:lyrical-ros-core` containers using the Lyrical testing repository:

- Tested a minimal consumer whose only CMake dependency is OMPL. With Boost development metadata present but no Eigen development package, `find_package(ompl)` fails at `find_dependency(Eigen3)`.
- Added the Eigen export, then configured, compiled, linked, and ran the minimal `ompl::ompl` consumer successfully.
- Ran `rosdep install --dependency-types build_export` against the patched manifest and confirmed it installs `libeigen3-dev`.
- Ran `bloom-generate rosdebian` for Ubuntu Resolute / ROS 2 Lyrical and confirmed `libeigen3-dev` appears in the generated binary run dependency closure.

The installed Lyrical export was audited for other active external lookups. FLANN, Triangle, and SPOT are disabled in that build, while Threads is provided by CMake/the toolchain.

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -31,6 +31,7 @@
   <exec_depend>libboost-serialization</exec_depend>
   <exec_depend>libflann</exec_depend>
 
+  <build_export_depend>eigen</build_export_depend>
   <build_export_depend>libboost-program-options-dev</build_export_depend>
   <build_export_depend>libboost-serialization-dev</build_export_depend>
   <build_export_depend>libboost-system-dev</build_export_depend>


Thodoris1999:
Minor: could you move that line next to the rest of the newly added `build_export_depend` declaration?
Otherwise looks good to go, thanks

anttix:
Addressed: moved the Eigen declaration alongside the other `build_export_depend` entries. Thanks!

Thodoris1999 (approved):

PR #1459, opened 2026-08-13
Automatically sort package.xml entries

Timple:
Random interested party here, but I noticed your [comment](https://github.com/ompl/ompl/pull/1458#issuecomment-5285654235). 
And since you already have pre-commit configured...


Thodoris1999 (approved):
hah, the things that can get randomly handed to you right when you need them 😄

PR #1461, opened 2026-08-16
Fix Python binding builds with system nanobind and MSVC

wolfv:
This fixes several Python binding portability issues found while packaging OMPL 2.0.2 for conda-forge:

- fall back to an installed `nanobind` CMake package when the git submodule is unavailable (for example, in GitHub-generated source archives)
- search the standard nanobind location below Python's `site-packages`
- use MSVC's `/wd4100` instead of the GCC/Clang-only `-Wno-unused-parameter`
- include `<string>` explicitly before using `std::to_string`
- update the missing-nanobind warning to mention both supported installation methods

The vendored submodule remains preferred when present, so existing source checkout and wheel workflows are unchanged.

Validation:
- configured without submodules against conda-forge's system `nanobind` package
- built the complete `_ompl` extension successfully on Linux
- the MSVC fixes were identified from conda-forge's Windows CI

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -94,7 +94,7 @@ if(OMPL_BUILD_PYTHON_BINDINGS)
     if(OMPL_HAVE_NANOBIND)
         add_subdirectory(py-bindings)
     else()
-        message(WARNING "OMPL_BUILD_PYTHON_BINDINGS is ON but Nanobind is not found. Use git submodule update --init --recursive to initialize the Nanobind submodule.")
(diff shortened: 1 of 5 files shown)


Thodoris1999:
Thank you for the MSVC support! This is very helpful

Would you mind adding the (previously removed) Windows runner here to make sure CI passes on Windows?

Thodoris1999:
Feel free to leave it for future work if making cibuildwheel work is too much, but 
there's also the PyPI wheel creation pipeline here https://github.com/ompl/ompl/blob/main/.github/workflows/wheels_build.yml

PR #1462, opened 2026-08-16
Add Pixi package build and CI

wolfv:
This adds a native Pixi package definition using the `pixi-build-cmake` backend and validates the resulting conda package across all supported runner platforms.

The `pixi.toml` defines:
- OMPL package metadata
- CMake build configuration
- build, host, and runtime dependencies
- Linux x86_64/aarch64, macOS x86_64/arm64, and Windows platforms
- tests against the installed package for both Python imports/API behavior and the exported `ompl::ompl` CMake target

The new workflow:
1. builds an actual conda package with `pixi build`
2. installs that local package into the locked Pixi workspace
3. runs the 89 Python tests
4. builds the CMake export consumer test
5. uploads the package artifact

The package intentionally disables demos, upstream build-tree tests, and optional VAMP integration. CI tests the installed package rather than only the source build tree.

Local validation on Linux x86_64:
- `pixi build --path pixi.toml` succeeded
- all 89 Python tests passed against the packaged installation
- `tests/cmake_export` configured and built against the packaged installation

The Windows matrix entry depends on the portability fixes in #1461. The PR itself is kept as an independent diff against `main` so it can be reviewed separately.

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -1,0 +1,56 @@
+[workspace]
+channels = ["https://prefix.dev/conda-forge"]
+platforms = ["linux-64", "linux-aarch64", "osx-64", "osx-arm64", "win-64"]
+preview = ["pixi-build"]
+
+[dependencies]
+ompl = { path = "." }
+cmake = ">=3.18"
+cxx-compiler = "*"
+pytest = "*"
+
+[tasks.test-python]
+cmd = "python -m pytest tests/pytests --ignore=tests/pytests/deprecated"
+
+[tasks.test-cmake-export]
+cmd = "cmake -S tests/cmake_export -B build/cmake-export && cmake --build build/cmake-export"
+
+[tasks.test]
+depends-on = ["test-python", "test-cmake-export"]
+
+[package]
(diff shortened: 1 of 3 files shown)


pixi.toml
```diff
@@ -0,0 +1,56 @@
...
+  "-DOMPL_BUILD_TESTS=OFF",
+  "-DOMPL_BUILD_PYTHON_BINDINGS=ON",
+  "-DOMPL_BUILD_VAMP=OFF",
+]
+
+[package.host-dependencies]
+libboost-devel = ">=1.68"
+eigen = ">=3.3"
+flann = ">=1.9.2"
+python = "3.12.*"
```
Thodoris1999, line 46: Is there anything preventing python 3.10, 3.11, 3.13, 3.14 support? The CI builds wheels for these versions

Thodoris1999 (commented):
Overall looks good to me. `releaseChecklist.txt` would need to be updated at some point to include bumping the version in the .toml
@mamoll just checking in with you before adding maintenance of another ompl package.

PR #1463, opened 2026-08-24
Fix StateSpace real-value copies in Python

### LINKED ISSUE #1454: Segfault in StateSpace.copyFromReals/copyToReals (Python bindings), reproduces on a plain RealVectorStateSpace
## Summary

`ompl.base.StateSpace.copyFromReals` (and `copyToReals`) segfault when
called from Python — no Python exception, the process just dies
(SIGSEGV / exit 139). This reproduces on a **plain, unconstrained
`RealVectorStateSpace`**, so it isn't specific to constrained planning
even though that's where we first hit it.

## Environment

- ompl: 2.0.1 (also reproduces on 2.0.0) via `pip install ompl`
- Python: 3.13
- OS: reproduced on macOS arm64; independently reproduced in a Linux
  arm64 (Docker) environment as well
- Install: prebuilt wheel, no source build

## Minimal reproducer

```python
from ompl import base as ob

space = ob.RealVectorStateSpace(3)
bounds = ob.RealVectorBounds(3)
bounds.setLow(-2); bounds.setHigh(2)
space.setBounds(bounds)

s = space.allocState()
space.copyFromReals(s, [1.0, 2.0, 3.0])  # segfaults here, no exception raised
print(s[0], s[1], s[2])  # never reached
```

Exits with code 139 (SIGSEGV) instead of raising or completing.

## Expected behavior

Either the state's values get set from the list without error, or a
Python-level exception is raised if this is a usage mistake — not a
silent process crash.

## Additional context

We ran into this while trying to set/read state values on a
`ConstrainedStateSpace`/`ProjectedStateSpace` state from Python. Two
paths we tried both failed:

- `state[i] = v` — raises `TypeError: ... does not support item
  assignment` (there's no bound `__setitem__` on
  `ConstrainedStateSpaceStateType`/`WrapperStateType`). This looks like
  it's by design rather than a bug.
- `StateSpace.copyFromReals`/`copyToReals` — segfaults as shown above,
  even with no constraint involved at all.

For anyone else who hits this: `ConstrainedStateSpace::StateType` does
bind a working alternative — `.copy(list_of_floats)` to write and
`__getitem__` to read (see
`py-bindings/base/spaces/constraint/ConstrainedStateSpace.cpp`), the
same pattern `demos/ConstrainedPlanningCommon.py`'s
`setStartAndGoalStates` uses (`sstart.copy(start)`). It isn't mentioned
on the [constrained planning docs
page](https://ompl.kavrakilab.org/constrainedPlanning.html) or the
[Python bindings page](https://ompl.kavrakilab.org/python.html), so it
took some digging through the binding source and demo code to find —
might be worth a doc note for future users, separate from the segfault
above.

tandede:
## Problem

The Python bindings expose `StateSpace.copyToReals` and `copyFromReals` by binding their C++ signatures directly, but the two APIs do not behave correctly through nanobind.

`copyFromReals` indexes the state space's value-location cache without first computing it. In a fresh Python state space, the cache can therefore be empty while the input contains values, which causes an out-of-bounds access in release builds. `copyToReals` writes into a converted temporary `std::vector<double>`, so the Python list supplied by the caller remains unchanged.

The inherited `WrapperStateSpace` bindings repeat the same direct bindings and bypass any fix made only on the base Python API.

## Approach

Keep the public Python API unchanged, but route both operations through small binding helpers:

- compute the value locations before either copy operation;
- let C++ populate a local vector for `copyToReals`, then replace the contents of the original Python list so callers receive the values;
- accept `list[float]` explicitly in the generated Python signature;
- remove the duplicate `WrapperStateSpace` bindings so wrappers inherit the same corrected implementation and still dispatch through the virtual C++ methods;
- replace the debug-only input-size assertion in `copyFromReals` with a descriptive exception, preventing release builds from indexing past the available locations.

This keeps location discovery in the state-space implementation and avoids duplicating its traversal logic in the bindings.

## Result

A newly constructed `RealVectorStateSpace` can now round-trip values without requiring a preceding `setup()` call. The caller-provided output list is updated in place, wrapper spaces use the same behavior, and an incorrectly sized input produces a Python `RuntimeError` instead of a release-build crash.

Regression coverage exercises both a direct real-vector state space and a wrapper state space, including the mismatched-length error path.

## Validation

- 2 focused real-copy regression tests passed.
- All 91 maintained Python tests passed.
- All 25 native CTest tests passed.
- All applicable pre-commit checks passed.
- `git diff --check` passed.

Fixes #1454.


fallenmi (approved):
This review was prepared with OpenAI Codex assistance under the account owner's authorization.

I reviewed exact head `73d24e11706b406009e0e2fd9b7705a006ca6548` and its clean merge with current `main` (`48552087421d86c0def3ace22e8733dc7fc75601`). An independent Python 3.14 oracle exercised fresh direct, wrapper, compound, pre-setup, prefilled-output, tuple-input, and mismatched-length paths five times on exact base, head, and current merge. The base crashes on fresh direct/compound/tuple copies, leaves the caller's output list unchanged after setup, rejects the wrapper state, and silently partially mutates a state for a short input. Head and current merge pass every valid case, update caller lists in place, and reject the short input before mutation with the documented exception.

The upstream `test_state_spaces.py` suite passes 15/15 on base and 17/17 on both head and current merge, including the two new regression tests. The native `test_state_spaces` target also passes on all three trees. `git diff --check` is clean, and all five published checks are successful.

No blocking findings.

Thodoris1999:
Addressed by #1466

PR #1467, opened 2026-09-07
Only install the shared library in Python directory on Wheel builds

traversaro:
I may be wrong or missing some detail, but I would guess that packages built for package managers that install both C++ and Python libraries (like `apt` or `conda`) should not install the `ompl` shared library in the Python install directory, that should instead be done when buildings wheels, otherwise the package end up containing a duplicated shared library.

However, I did not tested the ROS .deb package generation myself.

### DIFF
--- a/py-bindings/CMakeLists.txt
+++ b/py-bindings/CMakeLists.txt
@@ -98,7 +98,9 @@ add_custom_command(
 # install module
 install(TARGETS _ompl LIBRARY DESTINATION ${OMPL_PYTHON_INSTALL_DIR})
 # install ompl shared lib
-install(TARGETS ompl DESTINATION ${OMPL_PYTHON_INSTALL_DIR})
+if(SKBUILD)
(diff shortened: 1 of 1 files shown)


mamoll (approved):
Looks correct to me, but I will let @Thodoris1999 have the final word.

Thodoris1999:
> I may be wrong or missing some detail, but I would guess that packages built for package managers that install both C++ and Python libraries (like `apt` or `conda`) should not install the `ompl` shared library in the Python install directory, that should instead be done when buildings wheels, otherwise the package end up containing a duplicated shared library.
> 
> However, I did not tested the ROS .deb package generation myself.

This intentional because typically `OMPL_PYTHON_INSTALL_DIR` and `CMAKE_INSTALL_DIR` are different paths and are expected to work independently. As you can see the python tests fail without this installation.

Unless we drop support for `OMPL_PYTHON_INSTALL_DIR` and only support the skbuild python wheel, I think this line is needed.

PR #1470, opened 2026-09-16
Dubins airplane paths with asymmetric climb rate constraints

Jaeyoung-Lim:
**Problem Description**
The Dubins airplane assumes that the flight path angle for climbing and descending is identical.

However, for real fixed-wing vehicles this is not the case, as the maximum climb rate depends on the specific excess power, and the minimum climb rate (max descend rate) depends on specific drag of the aircraft. Typically, the smaller absolute value is chosen to keep the Dubins airplane paths feasible.

**Solution**
This PR sepearates the min and max pitch constraints of the OwenStateSpace such that seperate values can be sued for climbing and descending.

We could show the added feasibility of using higher climb rate constraints resulted in faster solve times for low altitude terrain navigation tasks.
<div align="center">
image
image
</div>

**Additional Context**
More detail can be found in this paper: https://arxiv.org/abs/2607.15863
```
@article{lim2026minimum,
  title={Minimum Time Dubins Airplane Paths with Asymmetric Climb Rates},
  author={Lim, Jaeyoung and Loianno, Giuseppe},
  journal={arXiv preprint arXiv:2607.15863},
  year={2026}
}
```


demos/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ if (OMPL_BUILD_DEMOS)
     add_ompl_demo(demo_StateSampling StateSampling.cpp)
     add_ompl_demo(demo_GeometricCarPlanning GeometricCarPlanning.cpp)
     add_ompl_demo(demo_DubinsAirplane DubinsAirplane.cpp)
+    add_ompl_demo(demo_AsymDubinsAirplane AsymDubinsAirplane.cpp)
```
mamoll, line 29: you forgot to add this file

src/ompl/base/spaces/src/OwenStateSpace.cpp
```diff
@@ -54,13 +54,14 @@ namespace
     constexpr std::uintmax_t MAX_ITER = 32;
 }  // namespace
 
-OwenStateSpace::OwenStateSpace(double turningRadius, double maxPitch)
+OwenStateSpace::OwenStateSpace(double turningRadius, double maxPitch, double minPitch)
```
mamoll, line 57: this can be confusing. it's easy to call the constructor with min/max reversed and not realize why this doesn't work. Can you refactor to follow the VanaOwenStateSpace pattern?:
```cpp
VanaOwenStateSpace::VanaOwenStateSpace(double turningRadius, double maxPitch)
  : VanaOwenStateSpace(turningRadius, {-maxPitch, maxPitch})
{
}
VanaOwenStateSpace::VanaOwenStateSpace(double turningRadius, std::pair<double, double> pitchRange)
  : rho_(turningRadius), minPitch_(pitchRange.first), maxPitch_(pitchRange.second), dubinsSpace_(turningRadius)
{
...
}
```

mamoll (changes requested):

PR #1472, opened 2026-09-19
feat: Add support for solving kinodynamic planning problems using geometric planners via differential flatness

claytonwramsey:
Hi!

My colleague @thaipduong and I recently published a paper on solving kinodynamic problems using differential flatness.
The original planner that we implemented was based on [VAMP](https://github.com/KavrakiLab/vamp), but we're porting it over to OMPL so that more people can take advantage of it.

This PR holds the bare-minimum implementation of differentially flat planning.
I have more PRs that I want to bring, with support for the following things:
- Atlas/chart support so we can do flat planning on spaces like SE(3)
- Python binding support
- Batched edge validation

On top of that, I've found a few bugs/holes in OMPL while implementing this.
Specifically, the support for asymmetric geometric spaces, where the motion from `q1 -> q2` is distinct from the motion from `q2 -> q1`.
- Lots of planners incorrectly report that they support non-symmetric spaces, even when they don't actually support asymmetric motions.
- PRM requires bidirectionality, but we can easily work around this by adding a `DirectedPRM` planner to support asymmetric motions

You may notice that this PR has a lot of tests on it. If this is too much to review or maintain, I can cull the test set to something easier to keep abreast of.
Also, I generated reference data for the tests in `generate_flat_steering_reference.cpp`, which refers to the currently-private flat planner implementation.
I can strip that out too if needed, but have left it in for now for sake of completeness.

I'll upload some stacked PRs based off of this one when they're ready. 

## LLM Coding Statement

The code in this PR was coded with the assistance of LLMs, but I have manually reviewed all the code in this PR and checked it for correctness.
I stand by it and consider it my own work.
The body text of my submission was all written by hand.


src/ompl/base/objectives/src/FlatEffortObjective.cpp
```diff
@@ -0,0 +1,74 @@
+/*********************************************************************
+ * Software License Agreement (BSD License)
+ *
+ *  Copyright (c) 2017, Rice University
```
mamoll, line 4: update year in all files

src/ompl/base/spaces/src/FlatMotion.cpp
```diff
@@ -0,0 +1,547 @@
...
+#include <algorithm>
+#include <cmath>
+#include <limits>
+#include <vector>
+
+/** Helper functions for manipulating polynomials. */
+namespace
+{
+    /** \brief Evaluate the polynomial whose ascending-power coefficients are \e c at \e t. */
+    double evaluatePolynomial(const Eigen::VectorXd &c, double t)
```
mamoll: It might be cleaner to refactor the code in this anonymous namespace by defining a Polynomial class like so:
```cpp
class Polynomial
{
public:
   Polynomial(const Eigen::VectorXd &c);
   evaluate(double t);
   Polynomial differentiate();
   std::vector<double> roots();
   ...
};
```
I guess you also do: `class Polynomial : private Eigen::VectorXd {...}`.

src/ompl/base/spaces/FlatTrajectory.h
```diff
@@ -0,0 +1,133 @@
...
+
+#include <vector>
+
+namespace ompl::base
+{
+    /// @cond IGNORE
+    OMPL_CLASS_FORWARD(FlatTrajectory);
+    /// @endcond
+
+    /** \brief A run of polynomial curves joined end to end, parameterized by time.
```
mamoll: this seems clearer:
```suggestion
    /** \brief A sequence of polynomial curves joined end to end, parameterized by time.
```

mamoll (changes requested):

src/ompl/base/spaces/src/FlatStateSpace.cpp
```diff
@@ -0,0 +1,350 @@
...
+
+    void FlatStateSpace::checkPlanner(const Planner *planner) const
+    {
+        if (planner == nullptr)
+            return;
+
+        const std::string &name = planner->getName();
+        if (REVERSING_PLANNERS.count(name) > 0u)
+            OMPL_WARN("%s doesn't check every edge in the direction its path runs through it, and edges in "
+                      "%s run forward in time. The paths it returns fail PathGeometric::check().",
```
mamoll:
```suggestion
                      "%s run forward in time. The paths it returns may fail PathGeometric::check().",
```

claytonwramsey:
I'm planning to add support for flat charts and atlases as well. Does it make sense to move all the flat-related source into a dedicated `base/spaces/flat` subdirectory?

PR #1474, opened 2026-09-19
fix: geometric planners now correctly report their support for asymmetric edge validation

claytonwramsey:
Turns out most of the planners incorrectly report whether they can handle state spaces with asymmetric edge validation! I've gone ahead and fixed all the ones that are easy changes.


src/ompl/geometric/planners/rrt/src/ATRRT.cpp
```diff
@@ -355,12 +357,13 @@ ompl::base::PlannerStatus ompl::geometric::ATRRT::solve(const base::PlannerTermi
...
 
-        std::vector<Motion *> mpath = computeDijkstraLowestCostPath(startMotion, solution);
+        std::vector<Motion *> mpath = computeDijkstraLowestCostPath(startMotion_, solution);
 
         // set the solution path
+        // computeDijkstraLowestCostPath already returns it in start to goal order
         auto path(std::make_shared<PathGeometric>(si_));
-        for (int i = mpath.size() - 1; i >= 0; --i)
-            path->append(mpath[i]->state);
+        for (auto *m : mpath)
```
mamoll, line 365: The original code reverses the path. This seems wrong.

mamoll (changes requested):
That's a lot of fixes! Thanks.

I have one comment. Rest looks good.

src/ompl/geometric/planners/rrt/src/ATRRT.cpp
claytonwramsey: I think this was a defect in the original implementation. `computeDijkstraLowestCostPath` already has this snippet:

```cpp
    std::vector<Motion *> path;
    for (Motion *m = b; m != nullptr; m = parentMap[m])
        path.push_back(m);

    std::reverse(path.begin(), path.end());
```

so the reversing segment above likely is a mistake.

